PR #2, opened 2017-06-28
LIVY-357. Web UI: Updated staticResourceServlet to return 404s and not throw WARNs

ajbozarth:
PR moved from old repo https://github.com/cloudera/livy/pull/339

[LIVY-357](https://issues.cloudera.org/browse/LIVY-357)

Since https://github.com/cloudera/livy/pull/319 added the UI whenever a static resource is accessed a `MimeException` is thrown as a `WARN`. This was caused because a `InputStream` must be wrapped in a `BufferedInputStream` for a full implementation.

I updated `staticResourceServlet` with the wrapper as well as adding a `404 File not found` return when a file does not exist (currently an empty file is returned in such a case)

Tested manually

### DIFF
--- a/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
+++ b/server/src/main/scala/com/cloudera/livy/server/LivyServer.scala
@@ -18,6 +18,7 @@
 
 package com.cloudera.livy.server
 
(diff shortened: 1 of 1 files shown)


jerryshao:
LGTM, merging to master.

PR #3, opened 2017-06-28
LIVY-352. Web UI: Added pagination by updating to use JQuery DataTables

ajbozarth:
PR moved from old repo https://github.com/cloudera/livy/pull/326

[LIVY-352](https://issues.cloudera.org/browse/LIVY-352)

Followup to https://github.com/cloudera/livy/pull/319 adding pagination and search by switching to use jQuery DataTables like the Spark Web UI.

none
both
sessions
batch

PR #5, opened 2017-06-28
LIVY-365. Include statement code in statement object

ajbozarth:
PR moved from old repo https://github.com/cloudera/livy/pull/339

[LIVY-365](https://issues.cloudera.org/browse/LIVY-365)

Based on a user list request and my work on the Livy Web UI including the code that was run in the statement return is useful. Currently there is no way for a user to check what code was run through the Livy REST API. This will be helpful when the Session Page in the Web UI is added in [LIVY-343](https://issues.cloudera.org/browse/LIVY-343) which will include a table of statements. Currently such a table could only display output without context.

I choose to insert `code` between `id` and `state` because it does not change value during execution like `state` and `output`. Since it can also makes sense to include between `state` and `output` given the potential length of its value, I can make such a change if requested.

Tested manually and passes current Livy UTs and ITs

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -914,6 +914,8 @@ A statement represents the result of an execution statement.
 +========+======================+=====================+
 | id     | The statement id     | integer             |
 +--------+----------------------+---------------------+
+| code   | The execution code   | string              |
++--------+----------------------+---------------------+
 | state  | The execution state  | statement state     |
 +--------+----------------------+---------------------+
 | output | The execution output | statement output    |
--- a/repl/src/main/scala/com/cloudera/livy/repl/Session.scala
+++ b/repl/src/main/scala/com/cloudera/livy/repl/Session.scala
@@ -103,7 +103,7 @@ class Session(
 
   def execute(code: String): Int = {
     val statementId = newStatementId.getAndIncrement()
-    val statement = new Statement(statementId, StatementState.Waiting, null)
+    val statement = new Statement(statementId, code, StatementState.Waiting, null)
     _statements.synchronized { _statements(statementId) = statement }
 
     Future {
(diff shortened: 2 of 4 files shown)


jerryshao:
Hi @ajbozarth , let's wait for the Travis is ready. Currently we still have some infra issues during moving to Apache.

Besides lets keep the PR title convention: LIVY-xxx. xxxx

ajbozarth:
@jerryshao I'll change the names back (used the Spark way by habit) and I figured reviews wouldn't go through until testing is set up, I've been staying in the loop on the move progress (Luciano sits next to me).

jerryshao:
I think the code is fine to me. Just requires Travis to run again to make sure every thing is fine. 

Currently Travis on incubator-livy is not ready, I already created a ticket about this (https://issues.apache.org/jira/browse/INFRA-14468). Can we please wait for some additional days? Sorry about the long delay.

ajbozarth:
Thanks for the link and the delay is nbd, it's actually given me time to focus on an import side project of mine.

jerryshao:
Actually the title should be (LIVY-365. xxx) :).

ajbozarth:
Ack, only one i messed up too >.<

jerryshao:
(unrelated to here) We should also consider ACLs when exposing this either on the UI or other places, it's not good to let others without permission to see the codes.

Another thing is that we should carefully modify REST APIs, this will potentially break usr's assumption. It might be OK for now since we're still in 0.4 version, but we still should think carefully about the necessity to change the API.

ajbozarth:
Thanks for the feedback @jerryshao I agree that we should be careful of modifying API and as you said, if this wasn't 0.4 I would have turned down the idea. As for ACLs this API call already checks for them, so currently the UI fails to display the statements table when ACLs are turned on (this is in the next PR I will be opening).

ajbozarth:
As for the necessity issue, this was requested in the user list plus, as I mentioned in the description, in the new Web UI (in a PR not yet opened that's waiting on #3 to merge) the statement table can be confusing to understand with only a result and no input code for context. With the assumption that as Livy moves forward more users will rely on the UI to read through old statements (as seen in Spark) an addition like this in v0.X is net positive in my eyes.

jerryshao:
Understood, thanks! Merging to master.

PR #10, opened 2017-06-30
[LIVY-364] livy-server writes one log file when set LIVY_MAX_LOG_FILE…

davidxdh:
[https://issues.cloudera.org/browse/LIVY-364](https://issues.cloudera.org/browse/LIVY-364)
When set LIVY_MAX_LOG_FILES to 0, livy-server renames old log file to livy-root-server.out.0 while startup. So there has one log file which is inconsistent with LIVY_MAX_LOG_FILES value 0.

### DIFF
--- a/bin/livy-server
+++ b/bin/livy-server
@@ -49,6 +49,11 @@ pid="$LIVY_PID_DIR/livy-$LIVY_IDENT_STRING-server.pid"
 livy_rotate_log() {
   log=$1
   num=$LIVY_MAX_LOG_FILES
+
+  if [ $LIVY_MAX_LOG_FILES -lt 1 ]; then
+    num=5
+  fi
+
   if [ -f "$log" ]; then # rotate logs
 	while [ $num -gt 1 ]; do
 	  prev=`expr $num - 1`


davidxdh:
@ajbozarth @jerryshao please review this PR. Thanks.

jerryshao:
So with your change, what's the behavior when LIVY_MAX_LOG_FILES sets to 0.

davidxdh:
Do not keep history log file when LIVY_MAX_LOG_FILES sets to 0. Only keep the log file of currently running livy server. @jerryshao

jerryshao:
I would consider LIVY_MAX_LOG_FILES setting to "0" as a invalid number, rather then to achieve the behavior for LIVY_MAX_LOG_FILES to "0".

What's your opinion @zjffdu ?

davidxdh:
Spark is how to deal with this problem?@jerryshao

jerryshao:
The behavior of Spark is the same as Livy:

```
spark_rotate_log ()
{
    log=$1;
    num=5;
    if [ -n "$2" ]; then
        num=$2
    fi
    if [ -f "$log" ]; then # rotate logs
        while [ $num -gt 1 ]; do
            prev=`expr $num - 1`
            [ -f "$log.$prev" ] && mv "$log.$prev" "$log.$num"
            num=$prev
        done
        mv "$log" "$log.$num";
    fi
}
```

I think this is really not so much necessary to fix.

ajbozarth:
I know I said LGTM, but I'm leaning to agree with @jerryshao on this. I think keeping consistent with Spark is a good idea and it's only one extra log file to worry about when set to 1

davidxdh:
We can change our minds. When the LIVY_MAX_LOG_FILES value is illegal or 0, the default value is used. @jerryshao @ajbozarth

jerryshao:
Ok, that's fine for me.

davidxdh:
@jerryshao @ajbozarth I have already resubmitted the PR, please review it.

bin/livy-server
```diff
@@ -49,6 +49,12 @@ pid="$LIVY_PID_DIR/livy-$LIVY_IDENT_STRING-server.pid"
 livy_rotate_log() {
   log=$1
   num=$LIVY_MAX_LOG_FILES
+
+  expr $LIVY_MAX_LOG_FILES + 0 &>/dev/null
```
jerryshao: Is it used to check whether `$LIVY_MAX_LOG_FILES` is integer or not?

davidxdh: You are right, do not need to check whether  LIVY_MAX_LOG_FILES is numbers. I have submitted pr, please check again.

davidxdh:
@jerryshao You are right, do not need to check whether LIVY_MAX_LOG_FILES is numbers. I have submitted pr. Please review it again.

jerryshao:
Thanks, LGTM.

PR #15, opened 2017-07-07
LIVY-348. Improve the ACLs in Livy

jerryshao:
This PR propose to improve current Livy's access control mechanism with fine-grained control:

1. If `livy.server.access-control.enabled` is disabled, which means ACLs is disabled, any user could send any request to Livy.
2. If `livy.server.access-control.enabled` is enabled, then this ACL mechanism divides users into 3 groups:
    1. view accessible users: users could get session, statement data, but cannot POST any queries.
    2. modify accessible users: users could submit new statements, kill sessions.
    3. super users: this is the same as previous, super user could impersonate any user.
   
    In the meanwhile, modify accessible users automatically have the view accessibility, and super user has all the permissions.

Also add new configuration `livy.server.access-control.allowed-users`, this is the same as previous `livy.server.access-control.users`, when ACLs is enabled only users in the allowed list could issue REST queries to Livy server, other users will get 403.

Please review and comment.

### DIFF
--- a/server/src/main/scala/org/apache/livy/LivyConf.scala
+++ b/server/src/main/scala/org/apache/livy/LivyConf.scala
@@ -73,7 +73,12 @@ object LivyConf {
   val SUPERUSERS = Entry("livy.superusers", null)
 
   val ACCESS_CONTROL_ENABLED = Entry("livy.server.access-control.enabled", false)
-  val ACCESS_CONTROL_USERS = Entry("livy.server.access-control.users", null)
+  // Allowed users to access Livy, by default any user is allowed to access Livy. If user want to
+  // limit the users who could access Livy, user should configure all the users with comma
+  // separated.
+  val ACCESS_CONTROL_ALLOWED_USERS = Entry("livy.server.access-control.allowed-users", "*")
+  val ACCESS_CONTROL_MODIFY_USERS = Entry("livy.server.access-control.modify-users", null)
+  val ACCESS_CONTROL_VIEW_USERS = Entry("livy.server.access-control.view-users", null)
 
   val SSL_KEYSTORE = Entry("livy.keystore", null)
   val SSL_KEYSTORE_PASSWORD = Entry("livy.keystore.password", null)
@@ -183,7 +188,6 @@ object LivyConf {
     ENABLE_HIVE_CONTEXT.key -> DepConf("livy.repl.enableHiveContext", "0.4"),
     CSRF_PROTECTION.key -> DepConf("livy.server.csrf_protection.enabled", "0.4"),
     ACCESS_CONTROL_ENABLED.key -> DepConf("livy.server.access_control.enabled", "0.4"),
-    ACCESS_CONTROL_USERS.key -> DepConf("livy.server.access_control.users", "0.4"),
     AUTH_KERBEROS_NAME_RULES.key -> DepConf("livy.server.auth.kerberos.name_rules", "0.4"),
     LAUNCH_KERBEROS_REFRESH_INTERVAL.key ->
       DepConf("livy.server.launch.kerberos.refresh_interval", "0.4"),
@@ -196,7 +200,7 @@ object LivyConf {
 
   private val deprecatedConfigs: Map[String, DeprecatedConf] = {
     val configs: Seq[DepConf] = Seq(
-      // There are no deprecated configs without alternatives currently.
+      DepConf("livy.server.access_control.users", "0.4")
     )
 
     Map(configs.map { cfg => (cfg.key -> cfg) }: _*)
@@ -211,9 +215,6 @@ object LivyConf {
 class LivyConf(loadDefaults: Boolean) extends ClientConf[LivyConf](null) {
 
   import LivyConf._
-
-  private lazy val _superusers = configToSeq(SUPERUSERS)
-  private lazy val _allowedUsers = configToSeq(ACCESS_CONTROL_USERS).toSet
 
   lazy val hadoopConf = new Configuration()
   lazy val localFsWhitelist = configToSeq(LOCAL_FS_WHITELIST).map { path =>
@@ -257,12 +258,6 @@ class LivyConf(loadDefaults: Boolean) extends ClientConf[LivyConf](null) {
     sparkHome().map { _ + File.separator + "bin" + File.separator + "spark-submit" }.get
   }
 
-  /** Return the list of superusers. */
-  def superusers(): Seq[String] = _superusers
-
-  /** Return the set of users allowed to use Livy via SPNEGO. */
-  def allowedUsers(): Set[String] = _allowedUsers
-
   private val configDir: Option[File] = {
     sys.env.get("LIVY_CONF_DIR")
       .orElse(sys.env.get("LIVY_HOME").map(path => s"$path${File.separator}conf"))
@@ -282,7 +277,7 @@ class LivyConf(loadDefaults: Boolean) extends ClientConf[LivyConf](null) {
     }
   }
 
-  private def configToSeq(entry: LivyConf.Entry): Seq[String] = {
+  def configToSeq(entry: LivyConf.Entry): Seq[String] = {
     Option(get(entry)).map(_.split("[, ]+").toSeq).getOrElse(Nil)
   }
 
--- a/server/src/main/scala/org/apache/livy/LivyConf.scala
+++ b/server/src/main/scala/org/apache/livy/LivyConf.scala
@@ -74,7 +74,7 @@ object LivyConf {
 
   val ACCESS_CONTROL_ENABLED = Entry("livy.server.access-control.enabled", false)
   // Allowed users to access Livy, by default any user is allowed to access Livy. If user want to
-  // limit the users who could access Livy, user should configure all the users with comma
+  // limit who could access Livy, user should list all the permitted users with comma
   // separated.
   val ACCESS_CONTROL_ALLOWED_USERS = Entry("livy.server.access-control.allowed-users", "*")
   val ACCESS_CONTROL_MODIFY_USERS = Entry("livy.server.access-control.modify-users", null)
--- a/server/src/main/scala/org/apache/livy/server/AccessFilter.scala
+++ b/server/src/main/scala/org/apache/livy/server/AccessFilter.scala
@@ -20,7 +20,7 @@ package org.apache.livy.server
 import javax.servlet._
 import javax.servlet.http.{HttpServletRequest, HttpServletResponse}
 
-class AccessFilter(accessManager: AccessManager) extends Filter {
+private[livy] class AccessFilter(accessManager: AccessManager) extends Filter {
 
   override def init(filterConfig: FilterConfig): Unit = {}
 
--- a/server/src/main/scala/org/apache/livy/server/AccessFilter.scala
+++ b/server/src/main/scala/org/apache/livy/server/AccessFilter.scala
@@ -20,9 +20,7 @@ package org.apache.livy.server
 import javax.servlet._
 import javax.servlet.http.{HttpServletRequest, HttpServletResponse}
 
-import org.apache.livy.LivyConf
-
-class AccessFilter(livyConf: LivyConf) extends Filter {
+class AccessFilter(accessManager: AccessManager) extends Filter {
 
   override def init(filterConfig: FilterConfig): Unit = {}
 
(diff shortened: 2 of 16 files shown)


jerryshao:
@zjffdu @ajbozarth would you please help to review this PR? Thanks!

ajbozarth:
I'll do my best to take a look at this later today or tomorrow

server/src/main/scala/org/apache/livy/server/LivyServer.scala
```diff
@@ -247,15 +250,10 @@ class LivyServer extends Logging {
       server.context.addFilter(csrfHolder, "/*", EnumSet.allOf(classOf[DispatcherType]))
     }
 
-    if (livyConf.getBoolean(ACCESS_CONTROL_ENABLED)) {
-      if (livyConf.get(AUTH_TYPE) != null) {
```
ajbozarth: Previously this was used to check for Kerberos. With this change ACLs now work independent of Kerberos, how do these changes interact with Kerberos then?

server/src/main/scala/org/apache/livy/server/AccessManager.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  private val viewAcls = (superUsers ++ modifyUsers ++ viewUsers).toSet
+  private val modifyAcls = (superUsers ++ modifyUsers).toSet
+  private val superAcls = superUsers.toSet
+
+  if (aclsOn) {
+    // Livy will load AccessFilter if acls is on. In this case if configured users (view users,
+    // modify users, super users) are not in the allowed user list, then AccessFilter check will
+    // be failed, so all these configured users should be included in the allowed users.
+    val notAllowedSuperUsers = superUsers.filter(!isUserAllowed(_))
```
ajbozarth: With this change to use view and modify lists is there a reason to keep an allowed list at all? It seems to me that all these checks are a bit much.

server/src/main/scala/org/apache/livy/server/AccessFilter.scala
```diff
@@ -31,11 +29,11 @@ class AccessFilter(livyConf: LivyConf) extends Filter {
                         chain: FilterChain): Unit = {
     val httpRequest = request.asInstanceOf[HttpServletRequest]
     val remoteUser = httpRequest.getRemoteUser
-    if (livyConf.allowedUsers.contains(remoteUser)) {
+    if (accessManager.isUserAllowed(remoteUser)) {
       chain.doFilter(request, response)
     } else {
       val httpServletResponse = response.asInstanceOf[HttpServletResponse]
-      httpServletResponse.sendError(HttpServletResponse.SC_UNAUTHORIZED,
+      httpServletResponse.sendError(HttpServletResponse.SC_FORBIDDEN,
```
jerryshao: Usually we use 403 forbidden rather than 401 unauthorized for user who doesn't have acces permission. In Spark we also use 403 rather than 401. 

401 is mostly used in spnego auth, where it returns 401 for authentication failure.

server/src/main/scala/org/apache/livy/LivyConf.scala
```diff
@@ -73,7 +73,12 @@ object LivyConf {
   val SUPERUSERS = Entry("livy.superusers", null)
```
jerryshao: which one do you prefer to?

server/src/main/scala/org/apache/livy/server/AccessManager.scala
jerryshao: "allowed users" is mainly used in `AccessFilter`, so you mean that with view and modify list `AccessFilter` is not necessary?

Currently in my implementation, by default any user is allowed to access Livy. But user could set "livy.server.access-control.allowed-users" to only allow a list could access Livy. If this list exists, then super/modify/view users should be in this list, otherwise it will break the semantics, that's why I put lots of check code here.

What do you think, do you think if `AccessFilter` is required or not?

server/src/main/scala/org/apache/livy/LivyConf.scala
ajbozarth: Given the value is only used for ACLs I think it's naming format should match the rest.

jerryshao: super user is not used in ACLs, generally it is used in user impersonation.

server/src/main/scala/org/apache/livy/server/AccessManager.scala
ajbozarth: I feel with the addition of view/modify users, allowed users seems redundant. Allowed users is the list of users that can access the REST API and view/modify users are the list that can view/modify the REST API, these seem to hold the same function. I think the allowed list could be replaced with a concatenation of the view, modify, and super lists.

server/src/main/scala/org/apache/livy/LivyConf.scala
ajbozarth: Oh ok, then we should leave it as is then

server/src/main/scala/org/apache/livy/server/AccessManager.scala
jerryshao: For example if user "A" is an allowed user, but has no view/modify access, then user "A" could still create his own and view this own session, but has no permission to view and modify others. If user "B" is not an allowed user, then user "B" cannot issue a REST query.  This is the purpose of allowed user, what do you think?

ajbozarth: Ok, so to clarify the lists:
Allowed: Can view/modify their own sessions but can't view/modify other users sessions
View: Can view/modify their own sessions and can view but not modify other users sessions
Modify: Can view/modify their own sessions and can view/modify other users sessions
Super: Can impersonate any user when impersonation is turned on plus has modify access

Based on these descriptions I think the naming was what was throwing me off. Taking another look at how you've constructed the view, modify, and super lists I think the allowed list should be constructed the same way: `private val allowedAcls = (superUsers ++ modifyUsers ++ viewUsers ++ allowedUsers).toSet` then uses `allowedAcls` instead of `allowedUsers`, this would remove the need to have the checks and bring it inline with how view and modify work.

server/src/main/scala/org/apache/livy/server/LivyServer.scala
ajbozarth: After a better understanding of your code I can clarify this question: Without this check you can't enable ACLs without enabling Kerberos. How will ACL authentication work without Kerberos? How will the user authenticate their user in their REST API calls?

server/src/main/scala/org/apache/livy/server/AccessManager.scala
jerryshao: Actually `allowed users = super + modify + view + other allowed`, Allowed users is a sum of all the users have additional permissions + just allowed users (cannot view or modify other's session).

ajbozarth: Ok so we're on the same page now on what each list means then. So on the code review side: I think rather than having the checks, you should just include the super, modify, and view user lists in the allowed users list in the code like you wrote in your comment.

jerryshao: I still don't understand you comment, can you describe more?

>you should just include the super, modify, and view user lists in the allowed users list in the code like you wrote in your comment.

ajbozarth: I'll make some comments in-line for clarity

server/src/main/scala/org/apache/livy/server/AccessManager.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  private val superUsers = conf.configToSeq(LivyConf.SUPERUSERS)
+  private val modifyUsers = conf.configToSeq(LivyConf.ACCESS_CONTROL_MODIFY_USERS)
+  private val viewUsers = conf.configToSeq(LivyConf.ACCESS_CONTROL_VIEW_USERS)
+  private val allowedUsers = conf.configToSeq(LivyConf.ACCESS_CONTROL_ALLOWED_USERS).toSet
+
+  private val viewAcls = (superUsers ++ modifyUsers ++ viewUsers).toSet
+  private val modifyAcls = (superUsers ++ modifyUsers).toSet
+  private val superAcls = superUsers.toSet
+
```
ajbozarth: I think you should add the following line here:
`private val allowAcls = (superUsers ++ modifyUsers ++ viewUsers ++ allowedUsers).toSet`

server/src/main/scala/org/apache/livy/server/AccessManager.scala
```diff
@@ -0,0 +1,124 @@
...
+    }
+  }
+
+  /**
+   * Check whether the given user has the permission to access REST APIs.
+   */
+  def isUserAllowed(user: String): Boolean = {
+    debug(s"user=$user aclsOn=$aclsOn, allowedUsers=${allowedUsers.mkString(", ")}")
+    if (!aclsOn || user == null || allowedUsers.contains(user) ||
+      allowedUsers.contains(WILDCARD_ACL)) {
```
ajbozarth: And replace `allowedUsers` in this function with `allowedAcls`

server/src/main/scala/org/apache/livy/server/AccessManager.scala
```diff
@@ -0,0 +1,124 @@
...
+  private val superUsers = conf.configToSeq(LivyConf.SUPERUSERS)
+  private val modifyUsers = conf.configToSeq(LivyConf.ACCESS_CONTROL_MODIFY_USERS)
+  private val viewUsers = conf.configToSeq(LivyConf.ACCESS_CONTROL_VIEW_USERS)
+  private val allowedUsers = conf.configToSeq(LivyConf.ACCESS_CONTROL_ALLOWED_USERS).toSet
+
+  private val viewAcls = (superUsers ++ modifyUsers ++ viewUsers).toSet
+  private val modifyAcls = (superUsers ++ modifyUsers).toSet
+  private val superAcls = superUsers.toSet
+
+  if (aclsOn) {
```
ajbozarth: This check could then be removed

server/src/main/scala/org/apache/livy/server/LivyServer.scala
jerryshao: Basically ACL and spnego are two things, for ACL we only require user name in http header,  which doesn't bind to a specific auth method, we can use LDAP/basic/spengo, once the user name exists in header, then ACLs can be worked.

server/src/main/scala/org/apache/livy/server/AccessManager.scala
jerryshao: I think our understanding of allowed users are different, in your description, allowed users are just one list separate from super/view/modify users. But in my code, allowed user **includes** super/view/modify users plus other users, so for example:

```
view user: view1, view2
modify user: modify1, modify2
super user: super1, super2,

Then allowed user should be: view1, view2, modify1, modify2, super1, super2, other1, other...

If view3 is in view list but not in allowed user list, then an exception will be thrown.

Comment (ajbozarth) [inline server/src/main/scala/org/apache/livy/server/AccessManager.scala]: So I'm suggesting both a change to what the allowedUsers config includes as well as how your code handles it. The super/view/modify configs separately list those users as you showed above, but for example in the code the `viewAcls` val will include the super, modify and view config lists. Whereas the allowed users config is expected to manually include the super/modify/view users. What I'm suggesting is that the allowed users config be like the other 3: using your example, the config would only list `other1, other2...` then in the code, just like the others, you would add the super/modify/view users to it. By including those lists in the code you would be able to remove the checks.

Comment (jerryshao) [inline server/src/main/scala/org/apache/livy/server/AccessManager.scala]: I see, I understand your comments now, let me think about it.

Comment (ajbozarth) [inline server/src/main/scala/org/apache/livy/server/AccessManager.scala]: Sorry I had trouble getting my idea across. My main goals for the change are to remove the checks in your code and bring consistency to how the super, modify, view and allowed users lists are config'd and used

**Comment by ajbozarth** [comment]
Thanks for the updates, the main code LGTM, but I haven't done a detailed reading through the test code yet, I'll try to make time to review it by the end of the week

File: server/src/test/scala/org/apache/livy/server/SessionServletSpec.scala [hunk 8]
```diff
@@ -116,25 +119,86 @@ class SessionServletSpec
...
           assert(res.owner === "alice")
-          assert(res.logs === Nil)
+          assert(res.logs === IndexedSeq("log"))
         }
         delete(res.id, aliceHeaders, SC_OK)
       }
     }
 
-    it("should prevent non-owners from modifying sessions") {
+    it("should allow non-owners from modifying sessions") {
```
ajbozarth: grammar nit: `from` should be `to`

ajbozarth:
Looked through the tests and they LGTM, but I'd still prefer a second opinion on them. @zjffdu mind doing a quick review?

jerryshao:
@ajbozarth do you have any further comment?

ajbozarth:
It LGTM, I'm just not confident about my review of the test code since I haven't spent much time in it yet, thus why I pinged @zjffdu for a second opinion

jerryshao:
Thanks @ajbozarth , @zjffdu  I'm going to merge this if you don't have further comment.

PR #21, opened 2017-07-25
[LIVY-376] [Docs] Update README with new naming and links and move Docs into separate files

ajbozarth:
[LIVY-376](https://issues.cloudera.org/browse/LIVY-376)

 * The README has been updated to match our new Apache naming and links.
 * The API Documentation has been split into separate files
 * Examples have been moved to the website https://github.com/apache/incubator-livy-website/pull/5

This PR will be followed up with a way to build out the Docs for the website


ajbozarth:
@jerryshao @lresende mind taking a look? (1/2)

jerryshao:
I would suggest to maintain API and examples in Livy source code, like what Spark did.

1. Docs are related to the specific version of Livy, when checking the APIs we should point to the right version of Livy.
2. Sometimes when changing API, it usually requires to change docs, so it would be better to put docs into Livy code.

ajbozarth:
Ok, tomorrow I'll update the PRs to move the two API md files back and find a way to still show them on the website. Should the examples md file go with them?

jerryshao:
Maybe examples are fine to stay in livy-website repo.

ajbozarth:
So the more I think about it, Spark is a unique case where the documentation is kept in the main repo then built and copied into the website repo. For most smaller projects I've seen where there's only a few files, like Livy, the docs are storied in the website. So I'm going to sleep on it and might open up the question to the dev list tomorrow.

jerryshao:
My concern is that if we keep document in another repo, we may forget to update docs in another repo when we change some API related things in code.

Sure we can bring this out in mail list to hear about others. But thinking is putting things not frequently changed in livy-website repo, whereas API reference, configuration docs in main repo.

ajbozarth:
While we're discussing where to keep the Docs on the mailing list I made a minor update to the order and wording of the REAME. If we decide to keep the Docs here I'll move them back over in this PR and setup a way to display them on the website in a separate PR

ajbozarth:
@jerryshao I moved the docs back into this PR so we can focus on reviewing these changes then I can follow up with a PR that actually lets us build the Docs like in Spark and deploy them to the website.

ajbozarth:
Created [LIVY-384](https://issues.cloudera.org/browse/LIVY-384) for followup mentioned above

docs/programmatic-api.md
```diff
@@ -0,0 +1,101 @@
...
+    </snapshots>
+  </repository>
+</repositories>
+```
+
+And add the Livy client dependency:
+
+```xml
+<dependency>
+  <groupId>com.cloudera.livy</groupId>
```
jerryshao: I think we should also fix it later on.

docs/programmatic-api.md
```diff
@@ -0,0 +1,101 @@
...
+To run Spark jobs within your applications, extend ``org.apache.livy.Job`` and implement
+the functionality you need. Here's an example job that calculates an approximate value for Pi:
+
+```java
+import java.util.*;
+
+import org.apache.spark.api.java.*;
+import org.apache.spark.api.java.function.*;
+
+import com.cloudera.livy.*;
```
jerryshao: Looks like here we still use old package name.

docs/programmatic-api.md
ajbozarth: Wow I didn't catch this when I was updating it, updating this would mean we'd have to have a apache release first though correct?

docs/programmatic-api.md
```diff
@@ -0,0 +1,101 @@
...
+Spark without having to maintain a local Spark context. Here shows how to use the Java API.
+
+Add the Cloudera repository to your application's POM:
+
+```xml
+<repositories>
+  <repository>
+    <id>cloudera.repo</id>
+    <url>https://repository.cloudera.com/artifactory/cloudera-repos</url>
+    <name>Cloudera Repositories</name>
```
ajbozarth: @jerryshao what should we do about this section? I'm not sure what the Apache repo url is and if you can get snapshots from it

docs/programmatic-api.md
jerryshao: Yes, that's my concern for here and above `repository` related docs. My thinking is that we can remove the above repository related docs and changes here to use apache one.

In the next release we will push the artifacts to maven repo, so adding above repository is not necessary. Also for the `SNAPSHOT` artifacts, we could add a note in the docs to ask user to install artifacts locally.

docs/programmatic-api.md
jerryshao: We could add notes to let user install snapshot artifacts locally by `mvn install`.

ajbozarth:
@jerryshao addressed your comments, what do you think about my "Note" ?

jerryshao:
LGTM, we could update the docs later when we have a first Apache release. I'm thinking of kicking off a first Apache release when all the migration things are done.

ajbozarth:
Yeah we should probably get it out soon, I'll try to speed up my web ui work so it can all get out in the release (I believe my only key remaining work after my current pr is auth support and selenium tests)

jerryshao:
Thanks! Will merge it after travis test finished.

jerryshao:
@ajbozarth can you please retrigger the travis?

jerryshao:
I'm going to merge this, since it is an unrelated failure.

PR #24, opened 2017-07-25
LIVY-380 Livy should overwrite spark.sql.catalogImplementation if Hive classes are not present on classpath

Victsm:
In SparkContextInitializer.spark2CreateContext, livy checks for whether catalog implementation is "hive" or "in-memory" to properly initialize the SparkSession.
If the config is set to "hive" (this is the case if livy.repl.enable-hive-context is set to "true"), livy will further check if hive classes are present or not.
It only invokes the enableHiveSupport method of SparkSession.Builder if hive classes are found.
If not, it proceeds to initializing a SparkSession without Hive support enabled.
However, in Spark's code base, when creating SparkSession, it checks for spark.sql.catalogImplementation to determine which session state to initialize.
If spark.sql.catalogImplementation is set to "hive", it will initialize a HiveSessionState.
However, since Livy has already checked that hive classes are not present, it should not go ahead invoking SparkSession.builder.getOrCreate without overwriting spark.sql.catalogImplementation to "in-memory".

### DIFF
--- a/repl/src/main/scala/org/apache/livy/repl/SparkContextInitializer.scala
+++ b/repl/src/main/scala/org/apache/livy/repl/SparkContextInitializer.scala
@@ -93,6 +93,8 @@ trait SparkContextInitializer extends Logging {
         spark = builder.getClass.getMethod("getOrCreate").invoke(builder)
         info("Created Spark session (with Hive support).")
       } else {
+        builder.getClass.getMethod("config", classOf[String], classOf[String])
+          .invoke(builder, "spark.sql.catalogImplementation", "in-memory")
         spark = builder.getClass.getMethod("getOrCreate").invoke(builder)
         info("Created Spark session.")
       }


jerryshao:
@Victsm what will be happened if we continue to enable `SparkSession` but without hive class presented?

Victsm:
@jerryshao 

Spark side will throw CNF exceptions when initializing `HiveExternalCatalog`.

repl/src/main/scala/org/apache/livy/repl/SparkContextInitializer.scala
```diff
@@ -93,6 +93,8 @@ trait SparkContextInitializer extends Logging {
         spark = builder.getClass.getMethod("getOrCreate").invoke(builder)
         info("Created Spark session (with Hive support).")
       } else {
+        builder.getClass.getMethod("config", classOf[String], classOf[String])
+          .invoke(builder, "spark.sql.catalogImplementation", "in-memory")
```
jerryshao: @Victsm why don't you override this in `SparkConf`, instead change it in `builder.config`?

Victsm: This config needs to be changed before invoking `builder.getOrCreate()`.
Otherwise, CNF exception will be thrown at line 98:
```
Caused by: java.lang.IllegalArgumentException: Error while instantiating 'org.apache.spark.sql.hive.HiveSessionState':
	at org.apache.spark.sql.SparkSession$.org$apache$spark$sql$SparkSession$$reflect(SparkSession.scala:981)
	at org.apache.spark.sql.SparkSession.sessionState$lzycompute(SparkSession.scala:110)
	at org.apache.spark.sql.SparkSession.sessionState(SparkSession.scala:109)
	at org.apache.spark.sql.SparkSession$Builder$$anonfun$getOrCreate$5.apply(SparkSession.scala:878)
	at org.apache.spark.sql.SparkSession$Builder$$anonfun$getOrCreate$5.apply(SparkSession.scala:878)
	at scala.collection.mutable.HashMap$$anonfun$foreach$1.apply(HashMap.scala:98)
	at scala.collection.mutable.HashMap$$anonfun$foreach$1.apply(HashMap.scala:98)
	at scala.collection.mutable.HashTable$class.foreachEntry(HashTable.scala:226)
	at scala.collection.mutable.HashMap.foreachEntry(HashMap.scala:39)
	at scala.collection.mutable.HashMap.foreach(HashMap.scala:98)
	at org.apache.spark.sql.SparkSession$Builder.getOrCreate(SparkSession.scala:878)
```

jerryshao: I'm still not quite sure about your comment. My thought is that instead of calling `builder.config`, why not using `conf.set("spark.sql.catalogImplementation", "in-memory")`? And this will also be invoked before `builder.getOrCreate`.

Victsm: When invoking `builder.config(conf)` at line 81, SparkSession will take all key/value pairs from the passed in conf object and add them to an internal HashMap.
https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/SparkSession.scala#L835
It does not keep a reference to the `conf` object.
So updating this config value from `conf` does not solve this issue, and it has to be changed using `builder.config`

jerryshao: Oh, I see, thanks for explanation.

jerryshao:
LGTM, merging to master.

PR #14, opened 2017-07-07
LIVY-359. Cache livy logs as config driven

praveenkanamarlapudi:
**Changes Made:**
1. Added logs cacheing as config driven.
2. Fixed adding driver process logs livy logs for local and client mode in interactive mode.

PR #25, opened 2017-07-26
[LIVY-344] [Web UI] Create Session Log Page

ajbozarth:
[LIVY-344](https://issues.cloudera.org/browse/LIVY-344)

Created a Session Log page for viewing the session log.
Updated All Sessions Page and Session Page with links to the Log Page.
Log Page exists for both Interactive Sessions and Batches.

Tested Manually, screenshots below

screen shot 2017-07-26 at 1 30 35 pm
screen shot 2017-07-26 at 1 30 40 pm
screen shot 2017-07-26 at 1 30 44 pm

### DIFF
--- a/server/src/main/resources/org/apache/livy/server/ui/static/js/session-log.js
+++ b/server/src/main/resources/org/apache/livy/server/ui/static/js/session-log.js
@@ -26,7 +26,7 @@ function getLogPath(type, id) {
 }
 
 function parseLog(logLines) {
-  // TODO: Separate out stdout, stderr, and YARN Diagnostics into different viewers
+  // TODO: [LIVY-387]: Separate out stdout, stderr, and YARN Diagnostics into different viewers
   return preWrap(logLines.join("\n"));
 }
 
--- a/server/src/main/resources/org/apache/livy/server/ui/static/js/session-log.js
+++ b/server/src/main/resources/org/apache/livy/server/ui/static/js/session-log.js
@@ -25,14 +25,19 @@ function getLogPath(type, id) {
   }
 }
 
+function parseLog(logLines) {
+  // TODO: Separate out stdout, stderr, and YARN Diagnostics into different viewers
+  return preWrap(logLines.join("\n"));
+}
+
 $(document).ready(function () {
   var pathArr = getPathArray();
   var type = pathArr.shift();
   var id = pathArr.shift();
 
-  $.getJSON(location.origin + getLogPath(type, id), {from: 0, size: 100}, function(response) {
+  $.getJSON(location.origin + getLogPath(type, id), {size: -1}, function(response) {
     if (response) {
-      $("#session-log").append(preWrap(response.log.join("\n")));
+      $("#session-log").append(parseLog(response.log));
     }
   });
 });
--- a/server/src/main/resources/org/apache/livy/server/ui/static/js/session-log.js
+++ b/server/src/main/resources/org/apache/livy/server/ui/static/js/session-log.js
@@ -15,23 +15,24 @@
  * limitations under the License.
  */
 
-body {
-  padding: 15px 0;
+function getLogPath(type, id) {
+  if (type == "session") {
+    return "/sessions/" + id + "/log";
+  } else if (type == "batch") {
+    return "/batches/" + id + "/log";
+  } else {
+    return "";
+  }
 }
 
-pre {
-  font-size: 12px;
-  line-height: 18px;
-  padding: 6px;
-  margin: 0;
-  word-break: break-word;
-  white-space: pre-wrap;
-}
+$(document).ready(function () {
+  var pathArr = getPathArray();
+  var type = pathArr.shift();
+  var id = pathArr.shift();
 
-td .progress {
-  margin: 0;
-}
-
-#session-summary {
-  margin: 20px 0;
-}
+  $.getJSON(location.origin + getLogPath(type, id), {from: 0, size: 100}, function(response) {
+    if (response) {
+      $("#session-log").append(preWrap(response.log.join("\n")));
+    }
+  });
+});
--- a/server/src/main/scala/org/apache/livy/server/SessionServlet.scala
+++ b/server/src/main/scala/org/apache/livy/server/SessionServlet.scala
@@ -230,8 +230,11 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
   private def serializeLogs(session: S, fromOpt: Option[Int], sizeOpt: Option[Int]) = {
     val lines = session.logLines()
 
-    val size = sizeOpt.getOrElse(100)
+    var size = sizeOpt.getOrElse(100)
     var from = fromOpt.getOrElse(-1)
+    if (size < 0) {
+      size = lines.length
+    }
     if (from < 0) {
       from = math.max(0, lines.length - size)
     }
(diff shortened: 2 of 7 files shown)


ajbozarth:
**Feedback Question:**
Currently I'm just displaying the log as a text dump, but should I address the following:
* Parse the log for `stdout`, `stderr`, and `YARN Diagnostics` so as to separate them
* Restrict the log size and implement a form of scrolling (like Spark's Log Page)

ajbozarth:
@jerryshao Here's the next UI PR

server/src/main/resources/org/apache/livy/server/ui/static/js/session-log.js
```diff
@@ -0,0 +1,38 @@
...
+    return "";
+  }
+}
+
+$(document).ready(function () {
+  var pathArr = getPathArray();
+  var type = pathArr.shift();
+  var id = pathArr.shift();
+
+  $.getJSON(location.origin + getLogPath(type, id), {from: 0, size: 100}, function(response) {
```
jerryshao: Based on the definition of `from` and `size`, here with your settings, it will always only show first 100 lines of log. I would instead use the default one. Please see the implementation of `serializeLogs` in `SessionServlet.scala`.

server/src/main/scala/org/apache/livy/server/ui/UIServlet.scala
```diff
@@ -63,9 +67,15 @@ class UIServlet extends ScalatraServlet {
   private def getNavBar(page: Page): Seq[Node] = {
     val tabs: Seq[Node] = page match {
       case _: AllSessionsPage => <li class="active"><a href="#">Sessions</a></li>
-      case pageInfo: SessionPage => {
+      case sessionPage: SessionPage => {
+        <li><a href="/ui">Sessions</a></li> ++
+          <li class="active"><a href="#">{sessionPage.name}</a></li>
+      }
+      case logPage: LogPage => {
+        val sessionLink = if (logPage.sessionType == "Session") "/ui/session/" + logPage.id else "#"
```
jerryshao: does it mean that log will only be shown for interactive session?

jerryshao:
> Parse the log for stdout, stderr, and YARN Diagnostics so as to separate them

This would be good if it is not difficult to achieve.

>Restrict the log size and implement a form of scrolling (like Spark's Log Page)

Here based on you implementation, we will only show 100 lines of logs at most, do you think it is necessary to add a form scrolling?

server/src/main/scala/org/apache/livy/server/ui/UIServlet.scala
ajbozarth: This is the link used in the header. You'll notice when on the Log Page the Header will have the tabs: `Sessions | Session # | Log` clicking `Session #` will go to the session summary page. In the case of a Batch log the Header will look like `Sessions | Batch # | Log` and click `Batch #` will do nothing (`href="#"`) since there's no Session Pages for Batches. I chose this option over linking back to the All Sessions Page, which was my other idea.

server/src/main/resources/org/apache/livy/server/ui/static/js/session-log.js
ajbozarth: I put 100 here as a placeholder until we address whether we want scrolling, as for why 100, it's the default value in the function you mentioned.

jerryshao: My concern is actually `from: 0`, by default it is -1 which has different meanings.

ajbozarth:
I'll separate out stderr/stdout/YARN tomorrow.

And by scrolling I meant a form of infinite scrolling (which I implemented on Sparks Log Page), though after your comment about 100 lines and my suggestion on splitting up the log I've realized that if stderr or stdout was too long YARN Diag won't even load. So I will probably implement a way to load the whole log no matter length and have scrolling within each of the "sub-logs"

server/src/main/resources/org/apache/livy/server/ui/static/js/session-log.js
ajbozarth: Ah, that was so it always displays the first 100 lines, whereas using -1 it will display the last 100 lines. Either way I'll be addressing this based on my comment below since we'll need to read the whole log.

ajbozarth:
Also based on everything I'm realizing in this conversation I may toy will how the logs are stored and accessed in order to better facilitate this page. (Without changing current API functionality of course)

ajbozarth:
Pushed an update addressing the log length, after a little work I found that separating out the different logs is simple, but from a UI perspective splitting them up doesn't do much without optimizing the UI for multiple logs, therefore I'm splitting off a followup JIRA to deal with splitting up the logs so we can merge more of the UI now and move forward with a 0.4 release faster.

ajbozarth:
opened up followup JIRA mentioned above: https://issues.cloudera.org/browse/LIVY-387

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -230,8 +230,11 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
   private def serializeLogs(session: S, fromOpt: Option[Int], sizeOpt: Option[Int]) = {
     val lines = session.logLines()
 
-    val size = sizeOpt.getOrElse(100)
+    var size = sizeOpt.getOrElse(100)
     var from = fromOpt.getOrElse(-1)
+    if (size < 0) {
+      size = lines.length
```
jerryshao: Please be aware even we set `size` to `-1`, still we cannot get the whole logs, by default it will only cache the latest 200 lines `(livy.cache-log.size)`.

jerryshao:
LGTM based on my limited knowledge of JS :).

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
ajbozarth: Oh, that code was merged after I wrote this (this has been sitting on my computer for a while waiting for the previous PRs to merge).

I feel like 200 is a really small default size for the logs, is there a reason that number was chosen @praveenkanamarlapudi ?

As for `size=-1`, since it's used in js the conf value wouldn't be available so it will work well for what ever `livy.cache-log.size` is set to.

jerryshao: Yeah, it's fine for JS. I just want to mention relying on this to show the whole log is not a good idea.

praveenkanamarlapudi: @ajbozarth we use livy a lot. We have seen memory issues due to logs(as it stored in memory). So, we set the default log size to 200 to resolve the memory issues.

ajbozarth: So I read through #14 to see exactly how `livy.cache-log.size` is used and it's a line limit on each of the stdout and stderr sections of the log, not the log length itself. So the default max log size is actually `(livy.cache-log.size * 2) + yarnDiagnostics.size + 3` (3 for each log header) and `yarnDiagnostics` has no size limit that I can find. So I'm ok with that default size I think.

ajbozarth: thanks @praveenkanamarlapudi

ajbozarth:
I pushed a small update to the TODO comment, shouldn't affect travis results

jerryshao:
Thanks, will merge after travis finished.

ajbozarth:
Travis failure are unrelated and passed before the update to the comment, is it ok to merge or should I re-trigger travis?

jerryshao:
Yes, please try again, thank you!

jerryshao:
Yes, please try again, thank you!

PR #27, opened 2017-08-01
[LIVY-387] [Web UI] Update Log page to split apart different logs

ajbozarth:
[LIVY-387](https://issues.cloudera.org/browse/LIVY-387)

Added javascript string parsing that finds the stdout, stderr, and YARN Diagnostics sections of the log and displays them in separate scrollable `<pre>` tags.

The `<pre>` tags are set to `max-height: 80%`, this makes scrolling down to each log easy so there is no need for an anchor link at the top of the page, also 80% allows a good sized viewing window. Each log is also set to scroll to the bottom on page load.

Manually Tested

Screenshots
screen shot 2017-08-01 at 2 58 50 pm
[more will be added tomorrow when I can use my desktop monitor for different sized screenshots, I'm currently working from my laptop]

### DIFF
--- a/server/src/main/resources/org/apache/livy/server/ui/static/js/session-log.js
+++ b/server/src/main/resources/org/apache/livy/server/ui/static/js/session-log.js
@@ -25,6 +25,13 @@ function getLogPath(type, id) {
   }
 }
 
+function logHeader(name) {
+  return "<h4><div data-toggle='tooltip' data-placement='right' "
+    + "title='Only the most recent log lines are displayed "
+    + "(max log lines is set by the livy.cache-log.size config)'>"
+    + name + "</div></h4>";
+}
+
 function parseLog(logLines) {
   // display non-standard log formats
   if (!logLines[0].startsWith("stdout")) {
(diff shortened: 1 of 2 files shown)


ajbozarth:
@jerryshao I had a moment to look at my followup to the log ui and came up with this, it's not as feature rich and complex as I was thinking when I split it off, but its simple and does the job well.

ajbozarth:
Added tooltip explaining `livy.cache-log.size`
screen shot 2017-08-01 at 3 49 58 pm

jerryshao:
>[more will be added tomorrow when I can use my desktop monitor for different sized screenshots, I'm currently working from my laptop]

Is this already done or you will push more commits to this PR?

ajbozarth:
It's done if those screenshots look ok to you, no more commits

server/src/main/resources/org/apache/livy/server/ui/static/js/session-log.js
```diff
@@ -25,9 +25,43 @@ function getLogPath(type, id) {
...
+
+  for (var i = 1; i < logLines.length; i++) {
+    if (stderrIndex == 0 && logLines[i].startsWith("\nstderr")) {
+      stdoutLines = logLines.slice(1, i);
+      stderrLines = logLines.slice(i, logLines.length);
+      stderrIndex = i;
+    } else if (logLines[i].startsWith("\nYARN Diagnostics")) {
+      stderrLines = logLines.slice(stderrIndex + 1, i);
+      yarnDiagLines = logLines.slice(i + 1, logLines.length);
+    }
```
jerryshao: If we already get stdout, stderr and yarnDiag I think we can jump out of loop, no need to continue the loop. Currently it will always loop from begin to end. I think we could improve this logics to be more efficient.

ajbozarth: nice catch, I meant to do that but must've missed it

ajbozarth: pushed an update

jerryshao:
LGTM, would you please retrigger the test, thanks!

jerryshao:
Failed with unrelated issue (will fix that flaky test).

Will merge to master, thanks for your work.

PR #28, opened 2017-08-02
[LIVY-194][REPL] Add shared language support for Livy interactive session

jerryshao:
This is a ongoing work. Putting it here to leverage travis to test.

In the current Livy when we create a new session we need to specify session kind, which represents the interpreter kind bound with this session. User could choose either `spark` or `pyspark` or `sparkr`, Livy internally will create a related interpreter with current session. Also in Livy each session represents a complete Spark application / SparkContext.

This brings a limitation that user can only choose one language along with its own Spark application. Furthermore, data generated by `pyspark` interpreter cannot be shared with `spark` interpreter. So to improve the usability of Livy we propose to add multiple interpreters support in one session.


core/src/main/scala/org/apache/livy/sessions/Kind.scala
```diff
@@ -30,21 +30,21 @@ case class PySpark() extends Kind {
   override def toString: String = "pyspark"
 }
 
-case class PySpark3() extends Kind {
```
jerryshao: I think it would be better to control the python exec with PYTHONPATH env variable, not interpreter kind, so that we could support python2, python3 or pypy. That's what Spark did.

Furthermore, in our shared language support, we cannot start both python2 and 3 interpreters, to add this `PySpark3` kind seems not meaningful, we can use `PYTHONPATH` to control the python exec we wanted.

integration-test/pom.xml
```diff
@@ -286,7 +286,7 @@
         <configuration>
           <environmentVariables>
             <LIVY_HOME>${execution.root}</LIVY_HOME>
-            <LIVY_TEST>false</LIVY_TEST>
+            <LIVY_TEST>true</LIVY_TEST>
```
jerryshao: Previously in our integration test we ignore PySpark or SparkR tests, because we don't have such libraries in mini cluster tests, so leaving `LIVY_TEST` false is fine. But now with my change, pyspark and SparkR interpreters will always be enabled, which requires such libraries, so like what we did in UT, we should set to `true` to ignore some library importing logics.

repl/scala-2.10/src/main/scala/org/apache/livy/repl/SparkInterpreter.scala
```diff
@@ -102,33 +102,28 @@ class SparkInterpreter(conf: SparkConf)
...
     }
   }
 
   override def close(): Unit = synchronized {
-    if (sparkContext != null) {
-      sparkContext.stop()
-      sparkContext = null
-    }
-
     if (sparkIMain != null) {
```
zjffdu: Where is SparkContext closed ?

zjffdu:
@jerryshao Could you add document about how to use shared interpreter via rest-api ?

repl/scala-2.10/src/main/scala/org/apache/livy/repl/SparkInterpreter.scala
jerryshao: It is closed in JobContextImpl.

jerryshao:
I will do the doc things in the following PR.

repl/src/main/scala/org/apache/livy/repl/Session.scala
```diff
@@ -77,40 +76,56 @@ class Session(
...
     val future = Future {
       changeState(SessionState.Starting())
-      val sc = interpreter.start()
-      _sc = Option(sc)
+
+      // Start SparkInterpreter first to make sure SparkContext is created within Scala REPL's
+      // classloader, otherwise the class created by REPL cannot be found by SparkContext.
+      interpGroup.get(Spark()).foreach(_.start(jobContext))
+      interpGroup.get(PySpark()).foreach(_.start(jobContext))
+      interpGroup.get(SparkR()).foreach(_.start(jobContext))
```
zjffdu: Does that means even scala-repl, python process and R process would always get started at the begining ? It would better to start them lazily, which means only start python process when user submit python code.

jerryshao: Yes, it will be started at beginning, my thinking is that the overhead is not so big. Let me think about how to handle this.

zjffdu: Overhead is one concern, another concern is that the starting of one interpreter may affect other interpreters. e.g. session may fail to start if R is not installed, but user may don't need RInterpreter.

jerryshao: @zjffdu , currently in the repl code, if Python interpreter is created first (with SparkContext intialized), then creating Spark interpreter will have classloader issue. Please see the comments above, I'm not sure how to handle it now.

jerryshao: SparkContext should be created with this wrapper in SparkInterpreter, I'm not sure how Zeppelin handle this issue.

```scala

    restoreContextClassLoader {
      // Call sparkIMain.setContextClassLoader() to make sure SparkContext and repl are using the
      // same ClassLoader. Otherwise if someone defined a new class in interactive shell,
      // SparkContext cannot see them and will result in job stage failure.
      val setContextClassLoaderMethod = sparkIMain.getClass().getMethod("setContextClassLoader")
      setContextClassLoaderMethod.setAccessible(true)
      setContextClassLoaderMethod.invoke(sparkIMain)

      // With usejavacp=true, the Scala interpreter looks for jars under System Classpath. But it
      // doesn't look for jars added to MutableURLClassLoader. Thus extra jars are not visible to
      // the interpreter. SparkContext can use them via JVM ClassLoaders but users cannot import
      // them using Scala import statement.
      //
      // For instance: If we import a package using SparkConf:
      // "spark.jars.packages": "com.databricks:spark-csv_2.10:1.4.0"
      // then "import com.databricks.spark.csv._" in the interpreter, it will throw an error.
      //
      // Adding them to the interpreter manually to fix this issue.
      var classLoader = Thread.currentThread().getContextClassLoader
      while (classLoader != null) {
        if (classLoader.getClass.getCanonicalName == "org.apache.spark.util.MutableURLClassLoader")
        {
          val extraJarPath = classLoader.asInstanceOf[URLClassLoader].getURLs()
            // Check if the file exists. Otherwise an exception will be thrown.
            .filter { u => u.getProtocol == "file" && new File(u.getPath).isFile }
            // Livy rsc and repl are also in the extra jars list. Filter them out.
            .filterNot { u => Paths.get(u.toURI).getFileName.toString.startsWith("livy-") }
            // Some bad spark packages depend on the wrong version of scala-reflect. Blacklist it.
            .filterNot { u =>
              Paths.get(u.toURI).getFileName.toString.contains("org.scala-lang_scala-reflect")
            }

          extraJarPath.foreach { p => debug(s"Adding $p to Scala interpreter's class path...") }
          sparkIMain.addUrlsToClassPath(extraJarPath: _*)
          classLoader = null
        } else {
          classLoader = classLoader.getParent
        }
      }
      postStart(jobContext)
    }
```

jerryshao:
@zjffdu , can you please help to review again. Based on your comments, now `PythonInterpreter` and `SparkRInterpreter` are started lazily until the related statement submitted. I think the core logics should not be so difficult to review (getting rid of all UI changes).

Also CC @ajbozarth to help to review.

Thanks a lot.

client-common/src/main/java/org/apache/livy/client/common/HttpMessages.java
```diff
@@ -82,13 +82,15 @@ private SessionInfo() {
   public static class SerializedJob implements ClientMessage {
 
     public final byte[] job;
+    public final String jobType;
 
-    public SerializedJob(byte[] job) {
+    public SerializedJob(byte[] job, String jobType) {
```
jerryshao: Because we now support different interpreters in one session, so we should have a way to figure out which type of job it is, is it a Spark job or a python job that should go to related interpreters.

core/src/main/scala/org/apache/livy/msgs.scala
```diff
@@ -28,7 +28,7 @@ case class Msg[T <: Content](msg_type: MsgType, content: T)
 
 sealed trait Content
 
-case class ExecuteRequest(code: String) extends Content {
+case class ExecuteRequest(code: String, kind: Option[String]) extends Content {
```
jerryshao: Also for the statement submission, we need to a field to differentiate what kind of code submitted. Here using `Option` means if it is `None` then we will choose the default kind which is specified in session creation (like Spark, PySpark, SparkR). If we specify `Shared` in session creation, here we must include kind in statement submission, otherwise it will throw exception.

repl/scala-2.11/src/main/scala/org/apache/livy/repl/SparkInterpreter.scala
```diff
@@ -89,16 +90,17 @@ class SparkInterpreter(conf: SparkConf)
         }
       }
 
-      createSparkContext(conf)
+      entries = new SparkEntries(conf)
```
jerryshao: The SparkInterpreter is responsible to create a `SparkEntries` which is the entries for Spark application. Python interpreter and R interpreter will require this `SparkEntries` to initialize python and R context.

repl/scala-2.10/src/main/scala/org/apache/livy/repl/SparkInterpreter.scala
```diff
@@ -26,20 +26,21 @@ import scala.tools.nsc.interpreter.JPrintWriter
...
 /**
  * This represents a Spark interpreter. It is not thread safe.
  */
-class SparkInterpreter(conf: SparkConf)
-  extends AbstractSparkInterpreter with SparkContextInitializer {
+class SparkInterpreter(conf: SparkConf) extends AbstractSparkInterpreter {
 
   private var sparkIMain: SparkIMain = _
-  protected var sparkContext: SparkContext = _
+  private var entries: SparkEntries = _
```
zjffdu: Is it possible to put `SparkEntries` into `AbstractSparkInterpreter` ? Because I see some duplicates in `SparkInterpreter.scala` of version 2.10 and 2.11

repl/src/main/scala/org/apache/livy/repl/PythonInterpreter.scala
```diff
@@ -37,21 +37,18 @@ import py4j.reflection.PythonProxyHandler
...
 object PythonInterpreter extends Logging {
 
-  def apply(conf: SparkConf, kind: Kind): Interpreter = {
-    val pythonExec = kind match {
-        case PySpark() => sys.env.getOrElse("PYSPARK_PYTHON", "python")
-        case PySpark3() => sys.env.getOrElse("PYSPARK3_PYTHON", "python3")
-        case _ => throw new IllegalArgumentException(s"Unknown kind: $kind")
-    }
+  def apply(conf: SparkConf, sparkEntries: SparkEntries): Interpreter = {
+    val pythonExec = sys.env.get("PYSPARK_PYTHON")
```
zjffdu: We may introduce regression issue here as we discard env `PYSPARK3_PYTHON` when the kind is `pyspark3`. Either we keep the compatibility or add document for that .

zjffdu:
@jerryshao I have tested the shared context in zeppelin, it works well. I just left a few minor comments. And here's 2 things left that could be done in separate tickets.
1. Add document for shared context
2. Add or enable system test of PythonInterpreter and SparkRInterpreter

zjffdu:
@jerryshao I just found it doesn't work for spark 1.6, could you check that ?

```
17/08/31 07:23:50 INFO driver.RSCDriver: SparkContext not yet up, queueing job request.
Exception in thread "pool-6-thread-1" java.lang.NoClassDefFoundError: org/apache/spark/repl/SparkIMain
    at org.apache.livy.repl.SparkInterpreter.start(SparkInterpreter.scala:49)
    at org.apache.livy.repl.Session$$anonfun$1.apply(Session.scala:121)
    at org.apache.livy.repl.Session$$anonfun$1.apply(Session.scala:115)
    at scala.concurrent.impl.Future$PromiseCompletingRunnable.liftedTree1$1(Future.scala:24)
    at scala.concurrent.impl.Future$PromiseCompletingRunnable.run(Future.scala:24)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
    at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.ClassNotFoundException: org.apache.spark.repl.SparkIMain
    at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
    ... 8 more
```

repl/scala-2.10/src/main/scala/org/apache/livy/repl/SparkInterpreter.scala
jerryshao: OK, I will change it.

repl/src/main/scala/org/apache/livy/repl/PythonInterpreter.scala
jerryshao: I'd rather adding some docs, making pyspark3 "kind" is not worked now and should be removed.

zjffdu:
@jerryshao Please ignore my last comments, the issue is due to my env.

jerryshao:
@zjffdu do you any further comment?

zjffdu:
LGTM

jerryshao:
Thanks @zjffdu , I will merge this and work on the following items.

PR #29, opened 2017-08-02
[Minor] Clean some old package name left in code and configuration file

jerryshao:
Looks like we still left some old package name uncleaned, so search and modify them again.


jerryshao:
Merging to master.

PR #30, opened 2017-08-03
[LIVY-390][BUILD] Only using JDK8 to build with Spark 2.2

jerryshao:
Because Spark 2.2 removes JDK7 support but Livy still supports JDK7, so updating travis file to use JDK8 to build against Spark 2.2, whereas still using JDK7 to build with Spark 2.2-.


jerryshao:
Merging to master.

PR #31, opened 2017-08-04
[LIVY-389] [Minor] Pom updates

ajbozarth:
Updated pom.xml to match other Apache and Apache Incubator projects

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,12 +20,6 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
-  <parent>
-    <groupId>org.apache</groupId>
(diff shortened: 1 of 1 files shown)


ajbozarth:
Had to remove parent section for now since it breaks the build, we should add it and address dependency issue in a future release.

jerryshao:
LGTM, I was thinking to remove cdh related repos in pom file, since it is not used any more. I think we will push artifacts to maven or apache repo for the Apache release (still I'm not sure how to push artifacts to maven or apache repo 😄 ). What do you think?

ajbozarth:
I left them in (just below the apache ones) because I noticed other incubator projects (incubator-streams for example) that did. I assume we can remove them eventually but I don't see a reason to without at least one release out. As for the release process I believe there was an email on the dev list recently with links about that.

jerryshao:
I see, thanks!

jerryshao:
Merging to master branch-0.4

PR #36, opened 2017-08-16
[LIVY-336] Livy should not spawn one thread per job to track the job on Yarn

meisam:
* Moved YARN related code from `SparkYarnApp` to a new class `YarnInterface`
* Removed "yarnAppMinotorThreads" that are created for each Spark application
* Created one thread that polls YARN to get `ApplicationId`s and
* Created a pool of threads that gets `ApplicationAttempts` and `ContainerReports` from YARN
  * because there is no API to get  `ApplicationAttempts` and `ContainerReports` in bulk
  * the thread pool is actually used by Scala `Future`s
* Updated test cases in `SparkYarnAppSpec` to match the new design

Task-Url: https://issues.cloudera.org/browse/LIVY-336

### DIFF
--- a/server/src/main/scala/org/apache/livy/server/LivyServer.scala
+++ b/server/src/main/scala/org/apache/livy/server/LivyServer.scala
@@ -40,6 +40,7 @@ import org.apache.livy.server.recovery.{SessionStore, StateStore}
 import org.apache.livy.server.ui.UIServlet
 import org.apache.livy.sessions.{BatchSessionManager, InteractiveSessionManager}
 import org.apache.livy.sessions.SessionManager.SESSION_RECOVERY_MODE_OFF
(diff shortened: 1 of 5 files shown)


mgaido91:
@meisam this PR looks old and inactive. May you please close it? Thanks.

vanzin:
IIRC there was a huge discussion on the mailing list about this... this would be a good feature to have if it's brought up to date.

meisam:
Sure! I'll update it.
Is the IIRC channel public? If yes, can you point me to it?

vanzin:
IIRC = if I remember correctly.

Azuaron:
I would love for this to actually get in. We are having problems right now with Livy thread starvation if more than 20 jobs are running at the same time.

runzhiwang:
Hi @meisam , Now we need this part. If there are no updates, I will continue your work in the new PR and keep your previous commits.

PR #37, opened 2017-08-23
[LIVY-393] [Minor] Update httpclient to 4.5.3

ajbozarth:
Currently the httpclient version in the root pom is 4.5.2 and in the client-http module it is 4.5.1

Now both use a common version and it's been updated to the latest version 4.5.3


jerryshao:
Looks good to me, merge to master and branch 0.4.

PR #38, opened 2017-08-30
[LIVY-141][LIVY-175][Docs] Update javadocs and scaladocs and include in Docs build

ajbozarth:
[LIVY-141](https://issues.apache.org/jira/browse/LIVY-141) [LIVY-175](https://issues.apache.org/jira/browse/LIVY-175)

Adding javadocs to Livy:
- Add ability to build Livy javadocs
- Update current javadoc comments to address build errors and warnings
- Add more javadoc comments to fully describe API
- Include public API javadocs and scaladocs in Livy Documentation

Noted Remaining Issues:
- Since only the public javadocs are build with the Livy Docs not all javadoc warnings in other modules have been addressed.
- There are still some warnings in the Scala API scaladocs build due to the docs not linking to an external lib for referenced scala lib classes, this does not break the build in any way.
- Scaladocs is not supported for Livy as a whole. With the update to Scala 2.11 many scaladoc warnings were upgraded to error and Livy fundamentally can't fix them.


ajbozarth:
Current Status:
- I've added the necessary plugins and configs to the pom so javadocs can be build with the `man javadoc:aggregate` command.
- scaladocs can already be built with the `man package scaladoc` command. This does not aggregate the docs (no existing way to aggregate scaladocs in a maven project), but the only public API docs are in the scala-api module.
- I addressed all the javadoc build errors and a majority of the warnings

Still TODO:
- Need to flesh out the documentation, especially in the rsc module
- Need to add javadocs/scaladocs to Docs build

Discussion:
- I ignored the example and test modules from the docs since they're not for the API
- I also ignored the client-common module, every class in the module is annotated `@Private` but javadocs doesn't have a way to ignore classes based on an interface so I just left out the module. The only other `@Private` class is in the api module, but it has a comment telling the user which class to use instead so I figured it was ok to leave in.
- ** *Input Needed* ** Many API java classes are not marked public and thus default to protected and are ignored by javadocs. Are they protected for a reason or should I make them public to include their documentation?

ajbozarth:
@jerryshao I'll most likely be picking this back up once I open a PR for Scala 2.11, could I get some feedback on my previous comments to use as a jumping point? I'd like to get this done within the next work week, I've let it sit *WIP* too long.

ajbozarth:
@jerryshao I'm about to start writing the script that will build then copy the API Docs into the Docs build and was wondering if you had a personal presence for language. For scripting my go-to is usually bash, but I could also use ruby like Spark or try python if you think using one of those would be better than bash. (I plan on writing the script tomorrow)

jerryshao:
Sorry for late response. I don't have a personal preference on language, you can choose your own.

As to expose Java/Scala APIs, actually what we need to expose is only about Job APIs for Java and Scala. I don't think it is necessary to expose other modules like rsc. This exposes unnecessary details to the users, and makes us hard to refactor.

jerryshao:
Currently Because we don't have a good scope definition for all the modules/classes, so I'm afraid building a javadoc will unnecessarily expose the internals.

ajbozarth:
If we'd rather not publish the java/scaladocs to the website then this is good to go as is (without the script). But I think having javadocs for at least the Job API is worthwhile. How about I allow users to build all the docs but the script (used for the website docs) will only build the Job API portion?

ajbozarth:
So I've made some updates as well as added a jekyll script to include the api docs

- Based on @jerryshao comments about the rsc module I created a `public-docs` build profile. I then moved the javadoc package restrictions there as well as added more so only the Java API docs are built with the profile. This also allows developers to build all the javadocs themselves if desired.
- I wrote a script that runs with a jekyll build that includes the Java API and Scala API docs with the Livy Docs build. This uses the above profile to only include the public API docs and not all of the javadocs and scaladocs

ajbozarth:
FYI I'm doing some refactoring after merging the Scala 2.11 update. Apparently 2.11.3 changed a bunch of WARNs into ERRORs for scaladoc

jerryshao:
Hi @ajbozarth , I think what I was saying is that we don't need to build java doc for internal modules like rsc and others, because user will not touch it. For now I think only Java/Scala Job API related docs should be enough.

What do you think?

ajbozarth:
Yes, sorry if I wasn't clear. I have updated the pom so anyone can build any of the javadocs or scaladocs, _BUT_ only the Java/Scala Job API portions are included in the Livy Documentation build.

jerryshao:
Yes, that's what I mean 😄 .

ajbozarth:
So the 2.11 update has brought to light new frustrating limitations of scaladocs with maven. I plan on spending some time tomorrow trying to find a solution, if I can't I'll just update the build script to bypass the root build to get the Scala API docs and I'll remove references to Livy supporting scaladocs for the project.

ajbozarth:
I'm pushed my final update as well as updated the PR description to include noted issues with javadocs and scaladocs for future reference.

In the end Livy can't support scaladocs in every module after the update to 2.11 due to our fractured versioning in our code (supporting both 2.10 and 2.11) causing errors in the scaladoc build (previously in 2.10 they were warnings not errors). As such I have updated the Docs build script to build the Java and Scala API Docs directly inside their respective modules to bypass project-wide issues. If devs wish to build scala/javadocs themselves they can deal with imperfect/broken private docs.

ajbozarth:
@jerryshao could you take a final look at this? I'd like to include it in RC1 when i cut it next Monday

ajbozarth:
@jerryshao not to pester, but I was holding off rc1 for your ok this, if I don't hear from you by your next EOD then I'll just cut rc1 without this.

jerryshao:
LGTM, sorry for the late response. Please go ahead.

PR #39, opened 2017-08-30
[LIVY-396] Livy does not map YARN app states correctly

meisam:
Make sure that the invalid combinations of `state` and `finalStatus` do
not occur, and if they do, put the application in the `FAILED` state.

Task-Url: https://issues.apache.org/jira/browse/LIVY-396

PR #44, opened 2017-09-06
[WIP]: [Livy-336]: Livy should not spawn one thread per job to track the job on Yarn (Use YARN REST API)

meisam:
Currently, Livy spawns one thread to for each Spark application to track the status of the application on YARN. The threads are called `yarnAppMonitorThread-xxxx` and make multiple calls to YARN using YARN's Java API to get the status of running applications or to kill applications. Spawning so many threads is a bottle neck and limits the number of Spark applications that can be submitted in a short span of time.

This pull request addresses the above issue in 2 ways:
1. It creates only one thread to poll YARN and get the status of all Spark applications.
1. It replaces the Java API with YARN REST API to eliminate multiple calls to YARN.

## How was this patch tested?

- The existing unit tests and integration test are updated to reflect the changes in the code.
- Every iteration of this PR is tested internally at PayPal at dev and production development. Some iterations have been running in productions for months.

## What is not in the scope of this pull request?
Based on the discussion and feedback on this pull request, and based on the feedback
- Calls to `yarnClient.kill` wont be replaced with REST API to kill YARN apps because:
  - The current kill REST API is experimental.
- Maven dependencies to YARN won't be removed because:
  - There is minimal benefit in removing all YARN dependencies. Livy still depends on Hadoop libraries.
  - If we remove YARN dependencies, we should find a different way to get the Name Node URL. One way to do that is to add new parameters in `livy.con, but that adds complexity and possibly confusion.

Task-url: https://issues.apache.org/jira/browse/LIVY-336

### DIFF
--- a/server/src/main/scala/org/apache/livy/LivyConf.scala
+++ b/server/src/main/scala/org/apache/livy/LivyConf.scala
@@ -122,6 +122,21 @@ object LivyConf {
   // Livy will cache the max no of logs specified. 0 means don't cache the logs.
   val SPARK_LOGS_SIZE = Entry("livy.cache-log.size", 200)
 
+  // The resource manager address for YARN's REST API.
+  val YARN_REST_SCHEMA = Entry("livy.server.yarn.rest.schema", "http")
+
+  // The resource manager address for YARN's REST API.
+  val YARN_REST_HOST = Entry("livy.server.yarn.rest.host", "localhost")
+
+  // The resource manager port for YARN's REST API.
+  val YARN_REST_PORT = Entry("livy.server.yarn.rest.port", 0)
+
+  // The resource manager port for YARN's REST API.
+  val YARN_REST_PATH = Entry("livy.server.yarn.rest.path", "/ws/v1/cluster/apps")
+
+  // The limit to the number of application reports to get on each REST call from YARN RM.
+  val YARN_REST_REPORTS_LIMIT = Entry("livy.server.yarn.resource-manager.report-limit", 100000)
+
   // If Livy can't find the yarn app within this time, consider it lost.
   val YARN_APP_LOOKUP_TIMEOUT = Entry("livy.server.yarn.app-lookup-timeout", "60s")
 
--- a/server/src/main/scala/org/apache/livy/utils/ApplicationReport.scala
+++ b/server/src/main/scala/org/apache/livy/utils/ApplicationReport.scala
@@ -1,0 +1,45 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.livy.utils
+
+/**
+  * A case class to keep all Application data that we get form YARN.
+  */
+case class ApplicationReport(
+                              id: String,
+                              name: String,
+                              user: String,
+                              applicationTags: Option[String],
+                              applicationType: String,
+//                              clusterId: Long,
+//                              clusterUsagePercentage: Float,
+                              diagnostics: String,
+                              elapsedTime: Long,
+                              finalStatus: String,
+                              finishedTime: Long,
+                              amContainerLogs: Option[String],
+//                              logAggregationStatus: String,
+//                              priority: Int,
+//                              progress: Float,
+                              queue: String,
+//                              runningContainers: Int,
+                              startedTime: Long,
+                              state: String,
+                              trackingUI: Option[String],
+                              trackingUrl: Option[String]
+                            )
--- a/server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
+++ b/server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
@@ -148,27 +148,30 @@ class SparkYarnApp private[utils] (
 
   // Exposed for unit test.
   private[utils] def mapYarnState(
-      appId: ApplicationId,
-      yarnAppState: YarnApplicationState,
-      finalAppStatus: FinalApplicationStatus): SparkApp.State.Value = {
-    yarnAppState match {
-      case (YarnApplicationState.NEW |
-            YarnApplicationState.NEW_SAVING |
-            YarnApplicationState.SUBMITTED |
-            YarnApplicationState.ACCEPTED) => SparkApp.State.STARTING
-      case YarnApplicationState.RUNNING => SparkApp.State.RUNNING
-      case YarnApplicationState.FINISHED =>
-        finalAppStatus match {
-          case FinalApplicationStatus.SUCCEEDED => SparkApp.State.FINISHED
-          case FinalApplicationStatus.FAILED => SparkApp.State.FAILED
-          case FinalApplicationStatus.KILLED => SparkApp.State.KILLED
-          case s =>
-            error(s"Unknown YARN final status $appId $s")
-            SparkApp.State.FAILED
-        }
-      case YarnApplicationState.FAILED => SparkApp.State.FAILED
-      case YarnApplicationState.KILLED => SparkApp.State.KILLED
+      appId: String,
+      yarnAppState: String,
+      finalAppStatus: String): SparkApp.State.Value = {
+    (yarnAppState, finalAppStatus) match {
+      case ("NEW", "UNDEFINED") |
+           ("NEW_SAVING", "UNDEFINED") |
+           ("SUBMITTED", "UNDEFINED") |
+           ("ACCEPTED", "UNDEFINED") =>
+        SparkApp.State.STARTING
+      case ("RUNNING", "UNDEFINED") =>
+        SparkApp.State.RUNNING
+      case ("FINISHED", "SUCCEEDED") =>
+        SparkApp.State.FINISHED
+      case ("FAILED", "FAILED") =>
+        SparkApp.State.FAILED
+      case ("KILLED", "KILLED") =>
+        SparkApp.State.KILLED
+      case (state, finalStatus) => // any other combination is invalid
+        error(s"Unknown YARN state $state for app $appId with final status $finalStatus.")
+        SparkApp.State.FAILED
     }
   }
 
+  def needsUpdating(appInfo: AppInfo): Boolean = {
+    (isRunning || appInfo.driverLogUrl == None || appInfo.sparkUiUrl == None)
+  }
 }
--- a/server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
+++ b/server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
@@ -16,20 +16,15 @@
  */
 package org.apache.livy.utils
 
-import java.util.concurrent.TimeoutException
-
-import scala.annotation.tailrec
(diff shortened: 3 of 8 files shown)


vanzin:
>  we can drop all maven dependencies to YARN

Not sure that brings a lot of gains; unless something changed, Livy uses a bunch of HDFS APIs already, so removing YARN would probably not remove a whole lot of dependencies.

meisam:
> Not sure that brings a lot of gains; unless something changed, Livy uses a bunch of HDFS APIs already, so removing YARN would probably not remove a whole lot of dependencies.

On top of that, if we remove all dependencies to YARN, we have to put the YARN cofings, such as resouce manager REST URL, in `livy.conf`. If we keep the dependencies, we can read the configs from yarn-site.xml and build the REST URL.

jerryshao:
What's the benefit of using REST API?

meisam:
The main benefit is that with one call we can get all the info we need for the app, that is appId, state, finalStatus, diagnostics, trackingUrl, SparkUiUrl, etc.
With Yarn's Java API, we can get appId with one call, but to get trakingURL, SparkUiUrl, etc., we have to make extra calls. First we need to get the attemptId, then we need to get the containerId, and then we can get the remaining information.

There is one other potential benefit, which is we can drop dependencies to YARN. But, as Marcelo described, the gains from dropping YARN dependencies are minimal.

I am alright with either YARN's Java API or REST API. I just added this PR because the proposal on the mailing list recommended REST.

server/src/main/scala/org/apache/livy/utils/ApplicationReport.scala
```diff
@@ -0,0 +1,45 @@
...
+//                              logAggregationStatus: String,
+//                              priority: Int,
+//                              progress: Float,
+                              queue: String,
+//                              runningContainers: Int,
+                              startedTime: Long,
+                              state: String,
+                              trackingUI: Option[String],
+                              trackingUrl: Option[String]
+                            )
```
ajbozarth: Two issues with this whole declaration:
1. Lots of commented out params, why?
2. I'm pretty sure this is bad style, try running style check and looking at other similar declarations

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -130,182 +128,50 @@ class SparkYarnApp private[utils] (
...
+        SparkApp.State.RUNNING
+      case ("FINISHED", "SUCCEEDED") =>
+        SparkApp.State.FINISHED
+      case ("FAILED", "FAILED") =>
+        SparkApp.State.FAILED
+      case ("KILLED", "KILLED") =>
+        SparkApp.State.KILLED
+      case (state, finalStatus) => // any other combination is invalid
+        error(s"Unknown YARN state $state for app $appId with final status $finalStatus.")
+        SparkApp.State.FAILED
```
ajbozarth: Are the above changes just #39 pulled over so this change works?

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -111,17 +106,20 @@ object SparkYarnApp extends Logging {
...
+    val listener: Option[SparkAppListener],
     livyConf: LivyConf,
-    yarnClient: => YarnClient = SparkYarnApp.yarnClient) // For unit test.
+    yarnInterface: YarnInterface)
   extends SparkApp
   with Logging {
-  import SparkYarnApp._
 
-  private val appIdPromise: Promise[ApplicationId] = Promise()
+  import scala.concurrent.ExecutionContext.Implicits.global
```
ajbozarth: What is the purpose of this import?

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -111,17 +106,20 @@ object SparkYarnApp extends Logging {
...
+
+  val appId: Future[ApplicationId] = Future {
+    appIdOption.fold(yarnInterface.getAppIdFromTag(appTag, process))(ConverterUtils.toApplicationId)
```
ajbozarth: This seems messy but personally I'm not sure how exactly to clean it up

server/src/main/scala/org/apache/livy/utils/YarnInterface.scala
```diff
@@ -0,0 +1,440 @@
...
+
+import org.apache.livy.{LivyConf, Logging, Utils}
+
+/**
+  * An interface to handle all interactions with Yarn.
+  */
+class YarnInterface(livyConf: LivyConf, yarnClient: YarnClient, httpClient: HttpClient)
+  extends Logging {
+
+  import scala.concurrent.ExecutionContext.Implicits.global
```
ajbozarth: again, purpose?

server/src/main/scala/org/apache/livy/utils/YarnInterface.scala
```diff
@@ -0,0 +1,440 @@
...
+//    println(s)
+//  }
+//
+//  def warn(s: String) = {
+//    println(s)
+//  }
+//
+//  def error(s: String) = {
+//    println(s)
+//  }
```
ajbozarth: these debugs should be removed

server/src/main/scala/org/apache/livy/utils/YarnInterface.scala
```diff
@@ -0,0 +1,440 @@
...
+    val port = livyConf.getInt(LivyConf.YARN_REST_PORT)
+    val path = livyConf.get(LivyConf.YARN_REST_PATH)
+
+    val address = InetAddress.getByName(host)
+    val target = new HttpHost(address, host, port, scheme)
+    val applicationTypes = "Spark"
+    val limit = livyConf.getInt(LivyConf.YARN_REST_REPORTS_LIMIT)
+    val query = s"applicationTypes=$applicationTypes&limit=$limit"
+
+    // parameters URI(scheme, user, host, port, path, query, fragment)
```
ajbozarth: purpose? remove?

ajbozarth (commented):
Gave feedback where I found it, but I'm no YARN expect so my review of the new `YarnInterface` class wasn't detailed. I also haven't had a chance to read through the `Spec` changes.

server/src/main/scala/org/apache/livy/LivyConf.scala
```diff
@@ -122,6 +122,21 @@ object LivyConf {
...
   val SPARK_LOGS_SIZE = Entry("livy.cache-log.size", 200)
 
+  // The resource manager address for YARN's REST API.
+  val YARN_REST_SCHEMA = Entry("livy.server.yarn.rest.schema", "http")
+
+  // The resource manager address for YARN's REST API.
+  val YARN_REST_HOST = Entry("livy.server.yarn.rest.host", "localhost")
+
+  // The resource manager port for YARN's REST API.
+  val YARN_REST_PORT = Entry("livy.server.yarn.rest.port", 0)
```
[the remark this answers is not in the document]
meisam: I'll change it the the default port 8088

server/src/main/scala/org/apache/livy/LivyConf.scala
```diff
@@ -122,6 +122,21 @@ object LivyConf {
...
+
+  // The resource manager port for YARN's REST API.
+  val YARN_REST_PATH = Entry("livy.server.yarn.rest.path", "/ws/v1/cluster/apps")
```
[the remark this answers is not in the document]
meisam: It is just the default path. See https://hadoop.apache.org/docs/stable/hadoop-yarn/hadoop-yarn-site/ResourceManagerRest.html#Cluster_Information_API

server/src/main/scala/org/apache/livy/utils/ApplicationReport.scala
meisam: This is one thing that I wanted to discuss with the Livy community. Scala cannot have case classes with more than 22 fields. I put all the fields that we can get from YARN REST for reference, but I commented out ones that are not as useful for Livy. I'd like to get feedback from the Livy users on which fields we should keep and which one we should leave out.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
meisam: It is needed as an implicit parameter to scala Future.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
meisam: Yes, except for one thing. #39 uses Java objects and enums. With YARN REST API, these are string values.

server/src/main/scala/org/apache/livy/utils/ApplicationReport.scala
ajbozarth: That makes sense, @jerryshao @zjffdu you guys probably know this side of Livy better, what fields do you think are best to include?

server/src/main/scala/org/apache/livy/utils/YarnInterface.scala
meisam: Just to clarify null parameters on the next line. The `URI` class has too many constructors.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
ajbozarth: It is? I've never noticed it in other classes that use scala Future

meisam:
Thanks @ajbozarth. This was helpful feedback.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
ajbozarth: I'll trust you on that and say this is ok then

server/src/main/scala/org/apache/livy/utils/ApplicationReport.scala
vanzin: @meisam I thought that only applied to case classes in 2.10; so you could work around it by not having a case class or by using 2.11 (which is what Livy uses, no?).

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
ajbozarth: Ok so this is a separate but related/dependent change then?

server/src/main/scala/org/apache/livy/utils/ApplicationReport.scala
ajbozarth: Livy itself is still built with scala 2.10, there's an open JIRA for the change that I tried to get traction on a while back with no response.

server/src/main/scala/org/apache/livy/utils/YarnInterface.scala
ajbozarth: I think it's ok without, the URI class is a pretty standard class

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
meisam: That is right. This is a separate but related change.

jerryshao:
I don't have a strong preference on changing to REST APIs, except one advantage you mentioned above (get all infos in one request). I'm not sure usually how many application will we launch in one Livy server, if it is only tens or hundreds of application, I think the communication overhead should not be big. 

Also avoiding yarn dependency is not a big problem, anyway we still need to depend on hadoop jars.

One difference I can think of is to visit security Hadoop, now because we change to REST API, so we should provide spnego principal/keytab, rather than Livy principal/keytab.

meisam:
> I don't have a strong preference on changing to REST APIs, except one advantage you mentioned above (get all infos in one request).

The main gain from this PR is LIVY-336: Livy should not spawn one thread per job to track the job on Yarn.

> One difference I can think of is to visit security Hadoop, now because we change to REST API, so we should provide spnego principal/keytab, rather than Livy principal/keytab.

[YARN's Cluster Applications API](https://hadoop.apache.org/docs/stable/hadoop-yarn/hadoop-yarn-site/ResourceManagerRest.html#Cluster_Applications_API) does not need authentication with one exception: deleting a YARN app needs authentication. But the delete API is experimental and I am not sure if it is a good idea to use it.

Since dropping the YARN dependencies has minimal benefits, it makes sense to kill YARN apps with java

PR #47, opened 2017-09-15
[LIVY-398][DOC] Update rest API to reflect the changes of shared context session

jerryshao:
LIVY-194 changes the semantics of session in Livy and add supports of shared context for Livy session, so here change the docs in REST API to reflect the changes of LIVY-194.

### DIFF
--- a/docs/rest-api.md
+++ b/docs/rest-api.md
@@ -153,8 +153,8 @@ Creates a new interactive Scala, Python, or R shell in the cluster.
   </tr>
 </table>
 
-<a name="footnote1">1</a>: Starting with version 0.5.0-incubating this field is not required. To be
-compatible with previous versions users could still specify this with spark, pyspark or sparkr,
+<a id="footnote1">1</a>: Starting with version 0.5.0-incubating this field is not required. To be
+compatible with previous versions users can still specify this with spark, pyspark or sparkr,
 implying that the submitted code snippet is the corresponding kind.
 
 #### Response Body
@@ -277,7 +277,7 @@ Runs a statement in a session.
   </tr>
 </table>
 
-<a name="footnote2">2</a>: If session kind is not specified or the submitted code is not the kind
+<a id="footnote2">2</a>: If session kind is not specified or the submitted code is not the kind
 specified in session creation, this field should be filled with correct kind.
 Otherwise Livy will use kind specified in session creation as the default code kind.
 
@@ -632,7 +632,7 @@ A session represents an interactive shell.
 </table>
 
 Starting with version 0.5.0-incubating, each session can support all three Scala, Python and R
-interpreters. The ``kind`` field in session creation is no longer required, instead users must
+interpreters. The ``kind`` field in session creation is no longer required, instead users should
 specify code kind (spark, pyspark or sparkr) during statement submission.
 
 To be compatible with previous versions, users can still specify ``kind`` in session creation,
--- a/docs/rest-api.md
+++ b/docs/rest-api.md
@@ -78,7 +78,7 @@ Creates a new interactive Scala, Python, or R shell in the cluster.
   <tr><th>Name</th><th>Description</th><th>Type</th></tr>
   <tr>
     <td>kind</td>
-    <td>The session kind<sup>[1](#footnote1)</sup></td>
+    <td>The session kind<sup><a href="#footnote1">[1]</a></sup></td>
     <td><a href="#session-kind">session kind</a></td>
   </tr>
   <tr>
@@ -272,7 +272,7 @@ Runs a statement in a session.
   </tr>
   <tr>
     <td>kind</td>
-    <td>The kind of code to execute<sup>[2](#footnote2)</sup></td>
+    <td>The kind of code to execute<sup><a href="footnote2">[2]</a></sup></td>
     <td><a href="#session-kind">code kind</a></td>
   </tr>
 </table>
--- a/docs/rest-api.md
+++ b/docs/rest-api.md
@@ -78,7 +78,11 @@ Creates a new interactive Scala, Python, or R shell in the cluster.
   <tr><th>Name</th><th>Description</th><th>Type</th></tr>
   <tr>
     <td>kind</td>
-    <td>The session kind (required)</td>
+    <td>
+      The session kind. Starting from 0.5, this field is not required, to be compatible
+      with old protocol, user could still specify this with spark, pyspark or sparkr, which
+      implies that the submitted code snippet is one of these kinds.
+    </td>
     <td><a href="#session-kind">session kind</a></td>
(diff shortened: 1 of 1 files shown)


jerryshao:
@ajbozarth would you please help to review when you have time, thanks!

maziyarpanahi:
Hi @jerryshao 
Just a quick note:

> Starting from Livy 0.5.0, each session could support **both** Scala, Python and R interpreters

I think it's best to replace "**both**" with "**all three**".

Thanks and cheers.

docs/rest-api.md
```diff
@@ -78,7 +78,11 @@ Creates a new interactive Scala, Python, or R shell in the cluster.
   <tr><th>Name</th><th>Description</th><th>Type</th></tr>
   <tr>
     <td>kind</td>
-    <td>The session kind (required)</td>
+    <td>
+      The session kind. Starting from 0.5, this field is not required, to be compatible
```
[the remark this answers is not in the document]
ajbozarth: Also it may be cleaner to move everything after `The session kind.` into a footnote below the table.

docs/rest-api.md
```diff
@@ -266,6 +270,16 @@ Runs a statement in a session.
     <td>The code to execute</td>
     <td>string</td>
   </tr>
+  <tr>
+    <td>kind</td>
+    <td>
+      The kind of code to execute (optional). If session kind is not specified or the submitted
```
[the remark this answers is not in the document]
ajbozarth: Similar to above everything after the first sentence might be better as a footnote.

docs/rest-api.md
```diff
@@ -610,18 +624,25 @@ A session represents an interactive shell.
...
-    <td>Interactive Python 3 Spark session</td>
+    <td>Interactive Python Spark session</td>
   </tr>
   <tr>
     <td>sparkr</td>
     <td>Interactive R Spark session</td>
   </tr>
 </table>
 
+Starting from Livy 0.5.0, each session could support both Scala, Python and R interpreters. The
```
ajbozarth: - `0.5.0-incubating`
- `could` -> `can`
- `both` -> all three`

docs/rest-api.md
```diff
@@ -610,18 +624,25 @@ A session represents an interactive shell.
...
+``kind`` field in session creation is not required anymore, instead user needs to specify code kind
```
ajbozarth: Grammar:
- `not required anymore` -> `no longer required`
- `user needs to` -> `users must`

docs/rest-api.md
```diff
@@ -610,18 +624,25 @@ A session represents an interactive shell.
...
+(spark, pyspark or sparkr) during statement submission.
+
+To be compatible with old protocol, user could still specify ``kind`` field in session creation
```
ajbozarth: Grammar:
- `old protocol` -> `previous versions`
- `could` -> `can`
- ~~`field`~~

docs/rest-api.md
```diff
@@ -610,18 +624,25 @@ A session represents an interactive shell.
...
+as a hint, while ignoring the ``kind`` field in statement submission, Livy will use this session
```
ajbozarth: - ~~`as a hint`~~
- end sentence after submission
- `Livy will then use`

docs/rest-api.md
```diff
@@ -610,18 +624,25 @@ A session represents an interactive shell.
...
+``kind`` as the default kind for all the submitted statements.
```
ajbozarth: ~~`the`~~

docs/rest-api.md
```diff
@@ -610,18 +624,25 @@ A session represents an interactive shell.
...
+
+If user want to submit code other than the default ``kind`` specified in session creation, user
```
ajbozarth: Grammar: plural `users` sounds better

docs/rest-api.md
```diff
@@ -610,18 +624,25 @@ A session represents an interactive shell.
...
+needs to specify code kind (spark, pyspark or sparkr) during statement submission.
+
```
ajbozarth: maybe include a note that `pyspark3` was deprecated and combined into `pyspark`?

jerryshao:
Sorry about my bad grammar, I will update it soon.

docs/rest-api.md
```diff
@@ -157,6 +153,10 @@ Creates a new interactive Scala, Python, or R shell in the cluster.
   </tr>
 </table>
 
+<a name="footnote1">1</a>: Starting with version 0.5.0-incubating this field is not required. To be
+compatible with previous versions users could still specify this with spark, pyspark or sparkr,
```
ajbozarth: `could` -> `can`

docs/rest-api.md
```diff
@@ -632,21 +631,25 @@ A session represents an interactive shell.
   </tr>
 </table>
 
-Starting from Livy 0.5.0, each session could support both Scala, Python and R interpreters. The
-``kind`` field in session creation is not required anymore, instead user needs to specify code kind
-(spark, pyspark or sparkr) during statement submission.
+Starting with version 0.5.0-incubating, each session can support all three Scala, Python and R
+interpreters. The ``kind`` field in session creation is no longer required, instead users must
```
ajbozarth: `must` -> `should`
On a second read, `should` makes more sense since it can still work the original way.

docs/rest-api.md
```diff
@@ -78,7 +78,7 @@ Creates a new interactive Scala, Python, or R shell in the cluster.
   <tr><th>Name</th><th>Description</th><th>Type</th></tr>
   <tr>
     <td>kind</td>
-    <td>The session kind (required)</td>
+    <td>The session kind<sup>[1](#footnote1)</sup></td>
```
ajbozarth: This syntax won't work with how we generate the Docs, this will need to be an href like other links in this doc

docs/rest-api.md
```diff
@@ -153,6 +153,10 @@ Creates a new interactive Scala, Python, or R shell in the cluster.
   </tr>
 </table>
 
+<a name="footnote1">1</a>: Starting with version 0.5.0-incubating this field is not required. To be
```
ajbozarth: to work with the href like I mentioned above this should be `id` instead of `name`

jerryshao:
@ajbozarth would you please review again to see if there's html issues. Thanks!

ajbozarth:
None that I can see from reading the code, did you build the Docs using the Docs README to check if they render fine with the Jekyll build?

ajbozarth:
If you can confirm the Docs build works with a screen shot then this LGTM

jerryshao:
OK, I will try to build the Docs to check it (though I haven't done it before).

jerryshao:
Thanks for your review, merging to master branch.

PR #48, opened 2017-09-19
[LIVY-41] Let users access sessions by session name

meisam:
This commit  enables Livy users to access sessions either by names or by auto-generated sessiond id's.

It also prevents users from creating sessions that have the same name.

This commit keeps API change  minimal. These are places that API change
is needed:
- `Session` and its sub-classes adds a new field, `name`.
- `RecoveryMetadata` and its sub-classes adds a new field, `name`.
- `SessionManager` adds a new method `getSession(name: String)` which looks sessions up by name.

Task-url: https://issues.apache.org/jira/browse/LIVY-41


server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -214,16 +215,23 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
   private def doWithSession(fn: (S => Any),
       allowAll: Boolean,
       checkFn: Option[(String, HttpServletRequest) => Boolean]): Any = {
-    val sessionId = params("id").toInt
-    sessionManager.get(sessionId) match {
+    val idParam: String = params("id")
```
ajbozarth: I don't think re-using the id param for name is a good idea. It should have it's own optional param instead. I believe this is already how you are doing session creation anyway.

server/src/main/scala/org/apache/livy/server/batch/BatchSessionServlet.scala
```diff
@@ -43,8 +44,23 @@ class BatchSessionServlet(
   override protected def createSession(req: HttpServletRequest): BatchSession = {
     val createRequest = bodyAs[CreateBatchRequest](req)
     val proxyUser = checkImpersonation(createRequest.proxyUser, req)
+    val sessionId = sessionManager.nextId()
+    val sessionName: String = createRequest.name match {
+      case Some(name) if sessionManager.get(name).isEmpty =>
+        name
+      case Some(name) =>
```
ajbozarth: I think this should be an else statement in the first case instead. This usage of match is a bit hacky.

server/src/main/scala/org/apache/livy/server/batch/BatchSessionServlet.scala
```diff
@@ -43,8 +44,23 @@ class BatchSessionServlet(
...
+        // this does NOT guarantee that by the time this session is ready to be registered in
+        // sessionManager, another with the same name is not registered. But in most cases,
+        // it prevents Livy from submitting applications to Spark.
+        val msg = s"Session $name already exists! " +
+          s"Choose a different name or delete the existing session."
+        throw new IllegalArgumentException(msg)
+      case None =>
+        s"INTERACTIVE-SESSION-$sessionId"
```
ajbozarth: This wasn't changed in the copy-paste. Also I'm not a fan of this default name, it's a bit verbose. Just `batch-$sessionId` would suffice.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSessionServlet.scala
```diff
@@ -53,8 +53,23 @@ class InteractiveSessionServlet(
...
+        name
+      case Some(name) =>
+        // this does NOT guarantee that by the time this session is ready to be registered in
+        // sessionManager, another with the same name is not registered. But in most cases,
+        // it prevents Livy from submitting applications to Spark.
+        val msg = s"Session $name already exists! " +
+          s"Choose a different name or delete the existing session."
+        throw new IllegalArgumentException(msg)
+      case None =>
+        s"INTERACTIVE-SESSION-$sessionId"
```
ajbozarth: As above, just `session-$sessionId` would suffice.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSessionServlet.scala
```diff
@@ -53,8 +53,23 @@ class InteractiveSessionServlet(
   override protected def createSession(req: HttpServletRequest): InteractiveSession = {
     val createRequest = bodyAs[CreateInteractiveRequest](req)
     val proxyUser = checkImpersonation(createRequest.proxyUser, req)
+    val sessionId: Int = sessionManager.nextId()
+    val sessionName: String = createRequest.name match {
+      case Some(name) if sessionManager.get(name).isEmpty =>
+        name
+      case Some(name) =>
```
ajbozarth: same as above

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
```diff
@@ -92,13 +94,29 @@ class SessionManager[S <: Session, R <: RecoveryMetadata : ClassTag](
   def register(session: S): S = {
     info(s"Registering new session ${session.id}")
     synchronized {
-      sessions.put(session.id, session)
+      // in InteractiveSessionServlet.createSession() and BatchSessionServlet.createSession(),
+      // Livy checks to make sure another session with the same name does not exist already.
+      // This case should not happen rarely.
+      // already exists. But looking up a session name and adding it to the set of existing sessions
+      // should happen atomically.
```
ajbozarth: Theres some confusing wording this comment block, might be artifacts from changes during development, but it makes the comment difficult to understand.

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
```diff
@@ -92,13 +94,29 @@ class SessionManager[S <: Session, R <: RecoveryMetadata : ClassTag](
...
+      if (sessionsByName.contains(session.name)) {
+        val msg = s"Session ${session.name} already exists!"
+        error(msg)
+        delete(session)
```
ajbozarth: **Discussion Point**: Do we want to fail a request with a "bad" name or should we instead give a warning message and create a session with a default name instead? Just an idea, I'm still not 100% sure of this PR's use case.

jerryshao:
@meisam , do you need to change Java and python job API to support this feature? As I remembered it can specify session id to reuse the existing session, so for now we should also support session name.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
meisam: I don't have any strong opinion for or against reusing `id param`. Here is why:

If we do not reuse `id param`, we have to create a new REST endpoint for each existing one. Reusing `id param` allows Livy users access a session seamlessly by its ID or its name through the same endpoint:

`sessions/{id}` or `sessions/{name}`
`sessions/{id}/statements` or `sessions/{name}/statements`
`sessions/{id}/state` or `sessions/{name}/state`
`sessions/{id}/log` or `sessions/{name}/log`

If we do not reuse `id param`,  we have to add separate endpoints for accessing sessions by name. Something like
`named-sessions/{name}`
`named-sessions/{name}/statements`
`named-sessions/{name}/state`
`named-sessions/{name}/log`
or
`sessions/by-name/{name}`
`sessions/by-name/{name}/statements`
`sessions/by-name/{name}/state`
`sessions/by-name/{name}/log`

or
`sessions?name={name}`
`sessions/statements?name={name}`
`sessions/state?name={name}`
`sessions/log?name={name}`

@ajbozarth, what are your thoughts on adding a new endpoint?

server/src/main/scala/org/apache/livy/server/batch/BatchSessionServlet.scala
meisam: 👍  I'll update it.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSessionServlet.scala
meisam: 👍

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
meisam: This is an example use case we have here. I'll ask Prabhu to provide more.

Team AAA has an ETL pipeline everyday at 1pm. The pipeline is a chain of jobs that work on the same dataset. They want to create a session on Livy and use it to run the pipeline. They also want to push the metrics and stats of their pipeline to altering and monitoring services.
This PR allows them to access the session, manage it, and monitor it with a human readable name like team_aaa_daily_etl.

server/src/main/scala/org/apache/livy/server/batch/BatchSessionServlet.scala
meisam: 👍  I'll update this even though I am an advocate of if-guards in case statements.

meisam:
@jerryshao Yes, changing the Java and python API is a good idea.

server/src/main/scala/org/apache/livy/server/batch/BatchSessionServlet.scala
ajbozarth: I was thinking an if/else inside the `Some(name)` case.

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
ajbozarth: Your example use case does help me with understanding this PR's purpose, which I am liking more, but not this particular code decision. Is there a good reason we should be failing session creation on a "bad" session name rather than displaying a warning that the name was bad then creating a session with the default naming? I'm not saying either way is better, I am just wondering you reason for doing it this way.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
ajbozarth: I checked the code to followup my argument and realized that the way these end-points are written the way you did it is actually the best way. The param name `id` is not user-facing so the clarity I want can be made in the Docs rather than the code. Given this change to the API though I would like you to double check that the Web UI will support these changes, both that it still works with the id, but also works by using a session name in the url. Actually adding the Session name to the UI can be done in a followup task (which I'm willing to do).

ajbozarth:
IMPO, for quicker pr reviews, I think adding Java and python API support can be done in a followup PR

server/src/main/scala/org/apache/livy/server/batch/BatchSessionServlet.scala
```diff
@@ -43,8 +44,23 @@ class BatchSessionServlet(
...
+    val sessionName: String = createRequest.name match {
+      case Some(name) if sessionManager.get(name).isEmpty =>
+        name
+      case Some(name) =>
+        // this does NOT guarantee that by the time this session is ready to be registered in
+        // sessionManager, another with the same name is not registered. But in most cases,
+        // it prevents Livy from submitting applications to Spark.
+        val msg = s"Session $name already exists! " +
+          s"Choose a different name or delete the existing session."
+        throw new IllegalArgumentException(msg)
```
jerryshao: Is the exception here OK, shall we return some HTTP code to rest client about session name duplication?

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSessionServlet.scala
```diff
@@ -53,8 +53,23 @@ class InteractiveSessionServlet(
   override protected def createSession(req: HttpServletRequest): InteractiveSession = {
     val createRequest = bodyAs[CreateInteractiveRequest](req)
     val proxyUser = checkImpersonation(createRequest.proxyUser, req)
+    val sessionId: Int = sessionManager.nextId()
+    val sessionName: String = createRequest.name match {
+      case Some(name) if sessionManager.get(name).isEmpty =>
```
jerryshao: We'd better put this code into common place for both interactive and batch session.

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
```diff
@@ -169,5 +174,5 @@ class BatchSession(
   override def infoChanged(appInfo: AppInfo): Unit = { this.appInfo = appInfo }
 
   override def recoveryMetadata: RecoveryMetadata =
-    BatchRecoveryMetadata(id, appId, appTag, owner, proxyUser)
+    BatchRecoveryMetadata(id, name, appId, appTag, owner, proxyUser)
```
jerryshao: @meisam what if old recovery metadata which doesn't have `name` field, I think we should consider the compatibility of old metadata file with new Livy server, would you please check it?

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
```diff
@@ -73,6 +73,8 @@ class SessionManager[S <: Session, R <: RecoveryMetadata : ClassTag](
 
   protected[this] final val idCounter = new AtomicInteger(0)
   protected[this] final val sessions = mutable.LinkedHashMap[Int, S]()
+  private[this] final val sessionsByName = mutable.Map[String, S]()
```
jerryshao: I think we can use `mutable.HashMap` to be clear.

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
jerryshao: I think fail the request and return user the specific bad http code would be better then choosing another name for user. Because user want to create a session with name "A" for example, if Livy cannot satisfy user's requirement, then user should choose another name. It is better to let user to decide than letting Livy to handle it.

server/src/main/scala/org/apache/livy/server/batch/BatchSessionServlet.scala
ajbozarth: nice catch, I not sure this would be surfaced to the submitter or not

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
meisam: 👍 I'll rename it.

integration-test/src/main/scala/org/apache/livy/test/framework/LivyRestClient.scala
```diff
@@ -205,12 +205,14 @@ class LivyRestClient(val httpClient: AsyncHttpClient, val livyEndpoint: String)
   }
 
   def startBatch(
+      name: String,
```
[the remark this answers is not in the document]
meisam: I'll update it to avoid explicit `null` values.

server/src/main/scala/org/apache/livy/server/batch/BatchSessionServlet.scala
meisam: One case that we should keep in mind about returning an error code. 
Assume the existing session `s` with name `daily-job`.
If the owner `s` is requesting a new session with name `daily-job`, we will reach this point and the user gets a http 400 bad request.
But if a different user is requesting a session `s2` with name `daily-job`, the user gets a 403 Forbidden before reaching this point, because he/she does not have access to `/sessions/daily-job/`.

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
meisam: What do you suggest? Should we assign such sessions auto-generated names on recovery? Or should we leave them without a name?

meisam:
> I think fail the request and return user the specific bad http code would be better then choosing another name for user. Because user want to create a session with name "A" for example, if Livy cannot satisfy user's requirement, then user should choose another name. It is better to let user to decide than letting Livy to handle it.

I like failing the request option more. The only caveat is that with current approach, sometimes we submit the spark App and then kill it later (if two requests arrive at almost the same time, we launch both, but later when we want to add them to sessionManager, we have to keep only one and kill and and discard the other one).

One way to avoid that is to refactor `Session`, `InteractiveSesion`, and `BatchSession` so it does not start the Spark process immediately, so we can start the Spark process after registering them in `SessionManager`.

Are you okay with this approach?

ajbozarth:
> Are you okay with this approach?

I would have to see the change first but I think this could potentially be ok

jerryshao:
> I like failing the request option more. The only caveat is that with current approach, sometimes we submit the spark App and then kill it later (if two requests arrive at almost the same time, we launch both, but later when we want to add them to sessionManager, we have to keep only one and kill and and discard the other one).

That's the problem of concurrency, I think we can first check the validity of session name, if it is conflicted, then directly return bad code. We should also postpone such check until session is created.

server/src/main/scala/org/apache/livy/server/batch/BatchSessionServlet.scala
jerryshao: Yeah, that's true.

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
jerryshao: I'm saying with your changes, if new Livy server tries to recover from old data, what will be happened? Will it be failed? 

Since the old recovery data doesn't have a session name field, so we could offer an auto-generated name, but we should not fail such scenario.

jerryshao:
> One way to avoid that is to refactor Session, InteractiveSesion, and BatchSession so it does not start the Spark process immediately, so we can start the Spark process after registering them in SessionManager.

@meisam I would suggest to leave this work to another PR for the simplicity of review.

jerryshao:
@meisam can you please address the comments, thanks!

meisam:
This is the output from last build:
>No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated.

Closing and re-opening this pull request to trigger a new build.

meisam:
> @meisam can you please address the comments, thanks!

@jerryshao I addressed comments in the previous commit.

integration-test/src/test/scala/org/apache/livy/test/BatchIT.scala
```diff
@@ -161,14 +161,14 @@ class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
...
-    val s = livyClient.startBatch(scriptPath, None, args, sparkConf)
+    val s = livyClient.startBatch(null, scriptPath, None, args, sparkConf)
     withSession(s)(f)
   }
 
   private def withTestLib[R]
     (testClass: Class[_], args: List[String], sparkConf: Map[String, String] = Map.empty)
     (f: (LivyRestClient#BatchSession) => R): R = {
-    val s = livyClient.startBatch(testLibPath, Some(testClass.getName()), args, sparkConf)
+    val s = livyClient.startBatch(null, testLibPath, Some(testClass.getName()), args, sparkConf)
```
jerryshao: Also here.

integration-test/src/test/scala/org/apache/livy/test/InteractiveIT.scala
```diff
@@ -191,7 +191,7 @@ class InteractiveIT extends BaseIntegrationTestSuite {
       waitForIdle: Boolean = true,
       heartbeatTimeoutInSecond: Int = 0)
     (f: (LivyRestClient#InteractiveSession) => R): R = {
-    withSession(livyClient.startSession(kind, sparkConf, heartbeatTimeoutInSecond)) { s =>
+    withSession(livyClient.startSession(null, kind, sparkConf, heartbeatTimeoutInSecond)) { s =>
```
jerryshao: Here, too.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
```diff
@@ -378,66 +383,13 @@ class InteractiveSession(
   sessionStore.save(RECOVERY_SESSION_TYPE, recoveryMetadata)
   heartbeat()
 
-  private val app = mockApp.orElse {
```
jerryshao: What's the purpose of below change?

integration-test/src/test/scala/org/apache/livy/test/BatchIT.scala
```diff
@@ -161,14 +161,14 @@ class BatchIT extends BaseIntegrationTestSuite with BeforeAndAfterAll {
   private def withScript[R]
     (scriptPath: String, args: List[String], sparkConf: Map[String, String] = Map.empty)
     (f: (LivyRestClient#BatchSession) => R): R = {
-    val s = livyClient.startBatch(scriptPath, None, args, sparkConf)
+    val s = livyClient.startBatch(null, scriptPath, None, args, sparkConf)
```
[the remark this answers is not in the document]
meisam: That's right. I prefer passing `None` to passing `null`. I'll fix it.

integration-test/src/test/scala/org/apache/livy/test/BatchIT.scala
meisam: I'll fix this one too.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
meisam: Before this change, Livy immediately submits the Spark app upon receiving the request . After this change, Livy has to call `Session.start()` to submit the Spark app, which means Livy can make sure another session with the same name does not exist before submitting the Spark app.

Without this change, making session names unique is complicated, because if two session names collide, we have only two choices:
- Kill one of the sessions.
- Come up with a logic that gives one of the sessions an autogenerated name and guarantees the autogenerated name does not collide with any of the existing/future session names.

jerryshao: Thanks for the explanation, let me review again.

meisam:
Is there a plan to merge this PR? If so, I can fix the recent merge conflicts. Otherwise we should close it for good.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
```diff
@@ -451,6 +403,64 @@ class InteractiveSession(
     }
   }
 
+  override def start(): Unit = {
+    app = mockApp.orElse {
+      val driverProcess = client.flatMap { c => Option(c.getDriverProcess) }
+        .map(new LineBufferedProcess(_, livyConf.getInt(LivyConf.SPARK_LOGS_SIZE)))
+      driverProcess.map { _ => SparkApp.create(appTag, appId, driverProcess, livyConf, Some(this))
+      }
```
vanzin: either in previous line or, if it doesn't fit, use multi-line syntax for the whole closure.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
```diff
@@ -451,6 +403,64 @@ class InteractiveSession(
...
+
+    if (client.isEmpty) {
+      transition(Dead())
+      val msg = s"Cannot recover interactive session $id because its RSCDriver URI is unknown."
+      info(msg)
+      sessionLog = IndexedSeq(msg)
+    } else {
+      val uriFuture = Future { client.get.getServerUri.get() }
+
+      uriFuture onSuccess { case url =>
```
vanzin: `uriFuture.onSuccess`, also in other places

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
```diff
@@ -92,13 +94,27 @@ class SessionManager[S <: Session, R <: RecoveryMetadata : ClassTag](
   def register(session: S): S = {
     info(s"Registering new session ${session.id}")
     synchronized {
+      session.name.map { sessionName =>
```
vanzin: s/map/foreach

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -214,16 +214,23 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
   private def doWithSession(fn: (S => Any),
       allowAll: Boolean,
       checkFn: Option[(String, HttpServletRequest) => Boolean]): Any = {
-    val sessionId = params("id").toInt
-    sessionManager.get(sessionId) match {
+    val idOrNameParam: String = params("id")
+    val session = if (idOrNameParam.forall(_.isDigit)) {
```
vanzin: You're assuming that a string with all digits is an ID and not a name. I don't see that enforced when creating a session?

meisam: I changed this to force session names to have at least one non-digit character.

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
meisam: Addressed 👍

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
meisam: Addressed 👍

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
meisam: In these lines, I only changed the indentation. I think formatting issues should be handled in a separate PR if needed.

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
meisam: Addressed 👍

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
meisam: I'll add test cases to confirm that Livy can recover from old data.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSessionServlet.scala
meisam: It is refactored into one place now.

meisam:
close/reopen to trigger the Travis CI...

meisam:
@vanzin This PR is ready for another review. I addressed all the issues.

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
```diff
@@ -140,31 +144,38 @@ object BatchSession extends Logging {
...
 
   override def stopSession(): Unit = {
-    app.kill()
+    app.foreach(_.kill())
   }
 
   override def appIdKnown(appId: String): Unit = {
-    _appId = Option(appId)
+    _appId = Option(
+      appId)
```
vanzin: undo

server/src/main/scala/org/apache/livy/sessions/Session.scala
```diff
@@ -135,13 +134,24 @@ object Session {
   }
 }
 
-abstract class Session(val id: Int, val owner: String, val livyConf: LivyConf)
+abstract class Session(val id: Int, val name: Option[String], val owner: String,
```
vanzin: one parameter per line when you go multi-line.

server/src/main/scala/org/apache/livy/sessions/Session.scala
```diff
@@ -135,13 +134,24 @@ object Session {
...
+                       val livyConf: LivyConf)
   extends Logging {
 
   import Session._
 
   protected implicit val executionContext = ExecutionContext.global
 
+  private def isSessionId(str: String): Boolean = {
```
vanzin: I'd just inline this in an `if` inside a `name.foreach` call, instead of using `filter`. Less code.

server/src/main/scala/org/apache/livy/sessions/Session.scala
```diff
@@ -135,13 +134,24 @@ object Session {
...
+    str.nonEmpty && str.forall(_.isDigit)
+  }
+
+  name.filter(isSessionId).foreach { idStr =>
+    val errMsg = s"Session name cannot be a number: $idStr"
+    error(errMsg)
```
vanzin: Not a big fan of scary error messages in logs caused by bad user input.

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
```diff
@@ -92,13 +94,27 @@ class SessionManager[S <: Session, R <: RecoveryMetadata : ClassTag](
   def register(session: S): S = {
     info(s"Registering new session ${session.id}")
     synchronized {
+      session.name.foreach { sessionName =>
+        if (sessionsByName.contains(sessionName)) {
+          val msg = s"Session ${session.name} already exists!"
+          error(msg)
```
vanzin: Same here.

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
```diff
@@ -92,13 +94,27 @@ class SessionManager[S <: Session, R <: RecoveryMetadata : ClassTag](
...
+          delete(session)
```
vanzin: You haven't registered the session, so do you need this?

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
```diff
@@ -92,13 +94,27 @@ class SessionManager[S <: Session, R <: RecoveryMetadata : ClassTag](
...
+          throw new IllegalArgumentException(msg)
+        } else {
+          // info(s"sessionNames = ${sessionsByName.keys.mkString}")
```
vanzin: remove

server/src/test/scala/org/apache/livy/server/batch/BatchSessionSpec.scala
```diff
@@ -100,11 +102,25 @@ class BatchSessionSpec
...
+      batch.appIdKnown("appId")
+      verify(sessionStore, atLeastOnce()).save(
+        Matchers.eq(BatchSession.RECOVERY_SESSION_TYPE), anyObject())
+    }
+
+    it("should recover session with no name ") {
+      val conf = new LivyConf()
+      val req = new CreateBatchRequest()
+      val mockApp = mock[SparkApp]
+      val m = BatchRecoveryMetadata(999, None, None, "appTag", null, None)
```
vanzin: Is this the test for Saisai's question about what happens with old recovery data?

I don't think this covers it. For that you need to test what happens when the serialized data doesn't have a `name` field. I'm almost sure that the deserialized `BatchRecoveryMetadata` in that case will contain a `null` (instead of `None`) for that field. But it would be good to have a test for that.

Similarly later for `InteractiveRecoveryMetadata`.

server/src/main/scala/org/apache/livy/sessions/Session.scala
meisam: fixed

server/src/main/scala/org/apache/livy/sessions/Session.scala
meisam: fixed

server/src/main/scala/org/apache/livy/sessions/Session.scala
meisam: fixed

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
meisam: Calling delete() cleans up session store. As soon as a session object is created (and before it is registered), the session stores its recovery metadata in session store.

server/src/main/scala/org/apache/livy/sessions/Session.scala
```diff
@@ -135,13 +134,25 @@ object Session {
...
 
   import Session._
 
   protected implicit val executionContext = ExecutionContext.global
 
+  private def isSessionId(str: String): Boolean = {
+    str.nonEmpty && str.forall(_.isDigit)
+  }
+
+  name.foreach { idStr =>
```
vanzin: `idStr` is a misleading name since this is not an id.

server/src/main/scala/org/apache/livy/sessions/Session.scala
```diff
@@ -135,13 +134,25 @@ object Session {
...
+    val owner: String,
+    val livyConf: LivyConf)
   extends Logging {
 
   import Session._
 
   protected implicit val executionContext = ExecutionContext.global
 
+  private def isSessionId(str: String): Boolean = {
+    str.nonEmpty && str.forall(_.isDigit)
```
vanzin: An empty name should be invalid too, right?

I think it would be clearer to have this code inlined below, as I suggested before. Then that whole block becomes the clear place where session name validation occurs.

server/src/test/scala/org/apache/livy/server/batch/BatchServletSpec.scala
```diff
@@ -141,6 +144,36 @@ class BatchServletSpec extends BaseSessionServletSpec[BatchSession, BatchRecover
...
       view.id shouldEqual id
+      view.name.isDefined shouldBe true
+      view.name shouldEqual Some(name)
+      view.state shouldEqual state.toString
+      view.appId shouldEqual Some(appId)
+      view.appInfo shouldEqual appInfo
+      view.log shouldEqual log
+    }
+
+    it("should show session properties for sessions without a name") {
```
vanzin: This looks very much like the test above. You could have:

```
Seq(
  (Some(id), None),
  (None, Some(name)
).foreach { case (id, name) =>
  it("should show session properties (id = $id, name = $name)") {
    methodWithTestImpl(id, name)
  }
}
```

server/src/test/scala/org/apache/livy/server/batch/BatchSessionSpec.scala
```diff
@@ -100,11 +102,25 @@ class BatchSessionSpec
...
+    it("should recover named session") {
       val conf = new LivyConf()
       val req = new CreateBatchRequest()
       val mockApp = mock[SparkApp]
-      val m = BatchRecoveryMetadata(99, None, "appTag", null, None)
+      val m = BatchRecoveryMetadata(99, Some("Test Batch Session"), None, "appTag", null, None)
+      val batch = BatchSession.recover(m, conf, sessionStore, Some(mockApp))
+
+      batch.state shouldBe (SessionState.Recovering)
+
```
vanzin: Check that the recovered session is named?

server/src/test/scala/org/apache/livy/server/interactive/InteractiveSessionServletSpec.scala
```diff
@@ -177,6 +180,45 @@ class InteractiveSessionServletSpec extends BaseInteractiveServletSpec {
...
+    view.owner shouldEqual owner
+    view.proxyUser shouldEqual proxyUser
+    view.state shouldEqual state.toString
+    view.kind shouldEqual kind.toString
+    view.appInfo should contain (Entry(AppInfo.DRIVER_LOG_URL_NAME, appInfo.driverLogUrl.get))
+    view.appInfo should contain (Entry(AppInfo.SPARK_UI_URL_NAME, appInfo.sparkUiUrl.get))
+    view.log shouldEqual log.asJava
+  }
+
+  it("should show session properties for sessions without a name") {
```
vanzin: Same thing here. Better to avoid copy & pasting the existing test and changing a couple of lines.

server/src/test/scala/org/apache/livy/server/interactive/InteractiveSessionSpec.scala
```diff
@@ -242,15 +243,32 @@ class InteractiveSessionSpec extends FunSpec
...
+      s.start()
+
+      s.state shouldBe (SessionState.Recovering)
+
+      s.appIdKnown("appId")
+      verify(sessionStore, atLeastOnce()).save(
+        MockitoMatchers.eq(InteractiveSession.RECOVERY_SESSION_TYPE), anyObject())
+    }
+
+    it("should recover sessions with no name") {
```
vanzin: Another one that looks like copy & paste of the existing test.

server/src/test/scala/org/apache/livy/server/recovery/BlackholeStateStoreSpec.scala
```diff
@@ -43,5 +44,24 @@ class BlackholeStateStoreSpec extends FunSpec with LivyBaseUnitTestSuite {
...
+          |  "id": 408107,
+          |  "appId": "application_1541532370353_1465148",
+          |  "state": "running",
+          |  "appTag": "livy-batch-408107-2jAOFzDy",
+          |  "owner": "batch_admin",
+          |  "proxyUser": "batch_opts",
+          |  "version": 1
+          |}
+        """.stripMargin
+      val batchRecoveryMetadata = stateStore.deserialize[BatchRecoveryMetadata](json.getBytes())
```
vanzin: nit: `getBytes(UTF_8)`

server/src/test/scala/org/apache/livy/sessions/SessionManagerSpec.scala
```diff
@@ -54,10 +54,48 @@ class SessionManagerSpec extends FunSpec with Matchers with LivyBaseUnitTestSuit
...
+        mock[SessionStore],
+        "test",
+        Some(Seq.empty))
+      val session1 = new MockSession(manager.nextId(), null, livyConf, Some(name))
+      val session2 = new MockSession(manager.nextId(), null, livyConf, Some(name))
+      manager.register(session1)
+      an[IllegalArgumentException] should be thrownBy manager.register(session2)
+      manager.get(session1.id).isDefined should be(true)
+      manager.get(session2.id).isDefined should be(false)
+      eventually(timeout(5 seconds), interval(100 millis)) {
```
vanzin: This test is not needed here, right? It has nothing to do with the session name.

server/src/test/scala/org/apache/livy/sessions/SessionManagerSpec.scala
```diff
@@ -54,10 +54,48 @@ class SessionManagerSpec extends FunSpec with Matchers with LivyBaseUnitTestSuit
       }
     }
 
+    it("should create sessions with names") {
+      val livyConf = new LivyConf()
+      val name = "Mock-session"
+      livyConf.set(LivyConf.SESSION_TIMEOUT, "100ms")
+      val manager = new SessionManager[MockSession, RecoveryMetadata](
```
vanzin: This is now in 3 different places in this file. Should probably be a helper method.

alex-the-man:
Have we considered the potential impact of this change for Livy multi-master HA?
To support multi-master HA without this feature, the only thing we need is a distributed atomic counter (which ZK offers).
With this change, we have to enforce session names are unique across multiple masters. This makes multi-master HA more complicated.

server/src/main/scala/org/apache/livy/sessions/Session.scala
meisam: > An empty name should be invalid too, right?
Right.

While testing what is a good way to separate session names from session IDs, I saw two interesting behaviors in Livy (and Scala/Java).
- `Char.isDigit` returns `true` for digits in all alphabets systems. For example:
```
scala> "᧑᧒᧓᧔᧕".forall(_.isDigit)
res18: Boolean = true

scala> "᧑᧒᧓᧔᧕".toInt
res19: Int = 12345

scala> "᧑᧒᧓᧔᧕".map(c => Character.getName(c.toInt))
res20: scala.collection.immutable.IndexedSeq[String] = Vector(NEW TAI LUE DIGIT ONE, NEW TAI LUE DIGIT TWO, NEW TAI LUE DIGIT THREE, NEW TAI LUE DIGIT FOUR, NEW TAI LUE DIGIT FIVE)
```

- If I create a session with `sessionID=1` all these are valid endpoints to the session:
```
${LIVY_SERVER}:${LIVY_PORT}/sessions/1
${LIVY_SERVER}:${LIVY_PORT}/sessions/01
${LIVY_SERVER}:${LIVY_PORT}/sessions/001
${LIVY_SERVER}:${LIVY_PORT}/sessions/00001
${LIVY_SERVER}:${LIVY_PORT}/sessions/۱
${LIVY_SERVER}:${LIVY_PORT}/sessions/0۱
${LIVY_SERVER}:${LIVY_PORT}/sessions/۰۰۰۱
${LIVY_SERVER}:${LIVY_PORT}/sessions/0۰۰00۰᧑
```

Should we keep this behavior as is?

meisam:
> To support multi-master HA without this feature, the only thing we need is a distributed atomic counter (which ZK offers).
With this change, we have to enforce session names are unique across multiple masters. This makes multi-master HA more complicated.

Very good point.

server/src/test/scala/org/apache/livy/sessions/SessionManagerSpec.scala
meisam: Refactored!

server/src/test/scala/org/apache/livy/sessions/SessionManagerSpec.scala
meisam: removed the garbage collection section.

server/src/test/scala/org/apache/livy/server/recovery/BlackholeStateStoreSpec.scala
meisam: fixed.

server/src/test/scala/org/apache/livy/server/interactive/InteractiveSessionSpec.scala
meisam: refactored into one helper method.

server/src/test/scala/org/apache/livy/server/interactive/InteractiveSessionServletSpec.scala
meisam: refactored into one helper method.

server/src/test/scala/org/apache/livy/server/batch/BatchSessionSpec.scala
meisam: Added the check.

server/src/test/scala/org/apache/livy/server/batch/BatchServletSpec.scala
meisam: refactored into one helper method.

server/src/main/scala/org/apache/livy/sessions/Session.scala
meisam: For now, disallowing empty strings as session names.

server/src/main/scala/org/apache/livy/sessions/Session.scala
meisam: renamed it to sessionName

vanzin:
Just to refresh my memory: is multi-master HA a thing? I looked at how session IDs are created today and it doesn't really seem that atomic when you have multiple servers looking at the store. In fact it doesn't seem that atomic even with a single server:

```
    val id = idCounter.getAndIncrement()
    sessionStore.saveNextSessionId(sessionType, idCounter.get())
```

That can write the wrong id to the session store if two threads race in just the wrong way.

Anyway, if multi-master is not a thing today, then this looks ok to me.

meisam:
> I looked at how session IDs are created today and it doesn't really seem that atomic ...

There is a separate PR to make sessionIDs atomic across multiple servers.
https://github.com/apache/incubator-livy/pull/103

vanzin:
So I guess pushing this change will require changes in that PR, or vice-versa. Seems this one is farther along.

meisam:
> So I guess pushing this change will require changes in that PR, or vice-versa. Seems this one is farther along.

No. #103 is not needed for this PR. It is part of a bigger PR to implement multi-server HA.

vanzin:
I didn't mean it's needed. I'm saying that the "how to have unique names with multiple servers" issue becomes a problem when either PR goes in.

So when I push this, that problem will have to be solved in the other PR. Or vice-versa, if the other PR went in first.

vanzin:
Merging to master.

PR #49, opened 2017-09-21
[LIVY-399][TEST] Enable real test for PySpark and SparkR interpreters

jerryshao:
Currently because we lack pyspark and sparkr dependencies in our environment, so we neglect the pyspark and sparkr integration tests, here propose a way to enable these integration tests. The solution is mainly borrowed from Zeppelin - to download Spark binary before integration tests, so that we will have a complete Spark environment to do the test. Because of this several modules like minicluster-dependencies are not required, so here removed such modules. 

Besides, we proposed to remove real test implementations, since we never used real test before and it mainly did what scripts should do with Scala code, which makes it hard to maintain. Also for now mini-cluster test is quite close to real test, so it is not so useful to still keep real test.


.travis.yml
```diff
@@ -57,17 +57,17 @@ before_install:
   - sudo apt-get -y install python3-pip python-dev
   - sudo apt-get -y install libkrb5-dev
   - sudo apt-get -y remove python-setuptools
-  - pip2 install --user --upgrade pip "setuptools < 36"
-  - pip3 install --user --upgrade pip "setuptools < 36"
-  - pip2 install --user codecov cloudpickle
-  - pip3 install --user cloudpickle
+  - sudo pip2 install --upgrade pip "setuptools < 36"
+  - sudo pip3 install --upgrade pip "setuptools < 36"
+  - sudo pip2 install codecov cloudpickle
```
jerryshao: It is weird that pyspark on yarn doesn't honor cloudpickle installed under user's directory, and will throw `no module named cloudpickle.cloudpickle`, and after changing to install these packages to system path, the issue is gone.

jerryshao:
@zjffdu @ajbozarth please help to review, thanks!

ajbozarth:
I don't see any glaring issues so if @zjffdu is ok with this then I am too.

jerryshao:
close and reopen to test again.

zjffdu:
LGTM

jerryshao:
Some pyspark test looks quite flaky, I'm still digging on it.

jerryshao:
Test various times, seems not it is more stable, merging to master, and will fix any flaky test if it occurs again.

PR #50, opened 2017-09-22
[LIVY-406][SERVER] Fix Livy cannot app id in yarn client mode issue

jerryshao:
In our `SparkYarnApp` logic, we have to wait for process to exit before to query app id. Since the process is never exited in yarn client mode, so it will block follow-up logic to get app id and other yarn application information. 

So here propose to remove this logic, also because now we will query app id ever since app is launched, so we should increase the look-up time to avoid timeout.

### DIFF
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -107,7 +107,7 @@
 # livy.server.recovery.state-store.url =
 
 # If Livy can't find the yarn app within this time, consider it lost.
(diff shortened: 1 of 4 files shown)


jerryshao:
CC @zjffdu to help to review, thanks!

bikassaha:
No tests seem to have been added or modified?

jerryshao:
Merge to Master and branch 0.4.

PR #52, opened 2017-09-26
[LIVY-406][FOLLOWUP][SERVER] Fix LIVY-406 UT failure in branch 0.4 and apply to Master branch for consistency

jerryshao:
Because we changed the AppState logic in #39 , so the UT which passed in master branch will be failed in branch 0.4, here to fix this issue.

Also apply this PR to master branch for the consistency, it should also be worked in master branch.


jerryshao:
Merging to master and branch 0.4.

PR #53, opened 2017-09-28
[LIVY-408][RSC] Update Netty version to 4.0.37.Final

jerryshao:
Netty version below 4.0.37.Final has some potential security issue (https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-4970), it is fixed in this version. So propose to upgrade Netty to avoid such issue.


jerryshao:
Merging to master.

PR #54, opened 2017-10-04
[Minor] Adding a PR Template

ajbozarth:
Added a pull request template like Spark's to help new Livy devs


ajbozarth:
@jerryshao do you think this would be useful? If it looks good to you I'll use this to test my new committer privileges.

jerryshao:
Yes, it looks good to me. Sorry for late response, I was off last week.

jerryshao:
I think you need to add this file to rat exclude.

ajbozarth:
I noticed you were off sometime around Thursday >.< and I'll add the rat exclude then push this after it passes tests

PR #55, opened 2017-10-05
[LIVY-409] Livy shell UX improvements

ericjperry:
I'm not sure how large of an audience the livy-shell has, but I've come to use it quite a bit in my testing/use of Livy and added a few small improvements that have made it a bit easier to use for me (and may help others):

* Replaced the use of httplib with requests to fix dropped connection issues (and a few other robustness issues).
* Added some context to the REPL prompt, by including the session `kind` and ID.
* Ignore `SIGINT` signal so that users can type `CTRL-C` to cancel out a command.

I know the use of `requests` is pretty ubiquitous but I included `livy-shell-requirements.txt` just in case 1) someone doesn't have it installed or 2) more dependencies are added down the road.

### DIFF
--- a/dev/livy-shell
+++ b/dev/livy-shell
@@ -37,14 +37,10 @@ import sys
 import time
 import urlparse
 
-try:
-  import requests
-except ImportError:
-  print "Unable to import 'requests' module, which is required by livy-shell."
-  sys.exit(1)
 
 class ControlCInterrupt(Exception):
   pass
+
 
 def check(condition, msg, *args):
   if not condition:
@@ -57,7 +53,14 @@ def check(condition, msg, *args):
 def message(msg, *args):
   if args:
     msg = msg % args
-  print msg
+  print(msg)
+
+
+try:
+  import requests
+except ImportError:
+  message("Unable to import 'requests' module, which is required by livy-shell.")
+  sys.exit(1)
 
 
 class LiteralDict(dict):
@@ -175,6 +178,7 @@ check(len(sys.argv) > 1, "Missing arguments.")
 
 url = urlparse.urlparse(sys.argv[1])
 sid = -1
+
 try:
   message("Creating new session...")
   session = create_session()
--- a/dev/livy-shell
+++ b/dev/livy-shell
@@ -67,8 +67,7 @@ def request(method, uri, body):
   if body:
     kwargs['json'] = body
   resp = requests.request(method.upper(), urlparse.urljoin(url.geturl(), uri), **kwargs)
-  if resp.status_code < requests.codes.ok or resp.status_code >= requests.codes.bad_request:
-    raise httplib.HTTPException, (resp.status_code, resp.reason, data)
+  resp.raise_for_status()
   if resp.status_code < requests.codes.multiple_choices and resp.status_code != requests.codes.no_content:
     return resp.json()
   return None
--- a/dev/livy-shell
+++ b/dev/livy-shell
@@ -43,6 +43,9 @@ except ImportError:
   print "Unable to import 'requests' module, which is required by livy-shell."
   sys.exit(1)
 
+class ControlCInterrupt(Exception):
+  pass
+
 def check(condition, msg, *args):
   if not condition:
     if args:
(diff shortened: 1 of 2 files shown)


ajbozarth:
A few notes and issues:
- If you could please open a JIRA for this and update the title according to the contributor guidelines (https://livy.incubator.apache.org/community/)
- I like your idea of using the requests library, but since I only have passing knowledge of python I'll defer to others on whether it makes sense from a python choice @jerryshao 
- I don't think adding the `livy-shell-requirements.txt` file makes sense, adding a inline comments or a `help` print line would be better options.
- I have mixed feelings on handling `SIGINT` but I lean toward agreeing with you
- I like the addition of showing kind and id in the prompt given our recent addition of shared language support.

ericjperry:
@ajbozarth thank you for the quick response

* Sorry about that, I have created https://issues.apache.org/jira/browse/LIVY-409 and updated the title of this PR to reflect the link.
* I removed `livy-shell-requirements.txt` and exit with a message if that module cannot be imported. Happy to take suggestions if you'd like to see something different.

ericjperry:
Running build again.

ericjperry:
@ajbozarth I fixed another issue I missed and re-ran the build a few times to get it to pass. This should be ready to review again.

jerryshao:
Thanks @ericjperry for your contribution, I think we seldom use Livy shell for testing and others, but it is better to keep this up-to-date, I think it is useful in local test. Will spend time to review this.

ajbozarth:
If @jerryshao signs off on this then LGTM

ericjperry:
@ajbozarth @jerryshao apologies, but I found an issue with the SIGINT handling in my code, which was concatenating lines of input together into a single command. I updated the loop to continue and reissue the `raw_input` call if a user submitted the ctrl-C command. This works far better and doesn't involve the raw writes to stdout which is a little nicer.

dev/livy-shell
```diff
@@ -56,33 +65,30 @@ class LiteralDict(dict):
...
-  conn.request(method, uri, body=body, headers=headers)
-
-  resp = conn.getresponse()
-  data = resp.read()
-  if resp.status < 200 or resp.status >= 400:
-    raise httplib.HTTPException, (resp.status, resp.reason, data)
-  if resp.status < 300 and resp.status != httplib.NO_CONTENT:
-    return json.loads(data)
+def request(method, uri, body):
+  kwargs = { 'headers': { 'Content-Type' : 'application/json' } }
```
jerryshao: I would suggest you to add CSRF header also, you can refer to python-api module to know the details.

dev/livy-shell
```diff
@@ -30,13 +30,22 @@
...
 import readline
+import signal
 import sys
 import time
 import urlparse
 
+try:
+  import requests
+except ImportError:
+  print "Unable to import 'requests' module, which is required by livy-shell."
```
jerryshao: Can you please change the script to make it python3 compatible, I think this syntax can only be worked in python2.

dev/livy-shell
ericjperry: Added the `X-Requested-By` header to match the other clients.

dev/livy-shell
ericjperry: I updated this line to reuse the `message` method and updated the `message` method to use a python3 compatible print call.

ajbozarth:
The latest update still look fine to me, but I was wondering why there are so many extra blank lines in the file? Is it a python thing? Cause I've alway thought multiple empty lines in a row was bad style no matter the language.

ericjperry:
@ajbozarth I may have gone a bit overboard but PEP8's guidance on blank lines says to surround top level functions and class definitions with two blank lines: http://legacy.python.org/dev/peps/pep-0008/#blank-lines.

Since there aren't any class members in here I split everything up with two blank lines.

jerryshao:
I think @ericjperry is right, python code follows two blank lines convention between class definitions and top level functions, you can refer to pyspark code @ajbozarth .

ajbozarth:
Thanks, good to know more python style conventions :)

ericjperry:
@ajbozarth @jerryshao is there any trick here to getting the tests to pass?

ajbozarth:
Usually just trying it again, I've never seen it fail this much before

ericjperry:
@ajbozarth @jerryshao Tests passed!

jerryshao:
Thanks for the contribution, merging to master.

PR #56, opened 2017-10-18
[LIVY-411][REPL] Fix session cannot start issue when python or r package is missing

jerryshao:

https://issues.apache.org/jira/browse/LIVY-411

In Livy 0.5.0, we supported multiple languages in one session, but it requires that all the packages should be available before session creation, such as R package and Python package, otherwise session will be failed to create. However, in some cases python or R package may be missing in Spark distro, this will make Livy fail to creation interactive session.

To fix this issue, we should not force such restriction on session creation, but delay the check until related interpreter is used. If such packaging is missing, we should make the related execution failure and return user the cause of issue, but don't affect other correctly started interpreters.

## How was this patch tested?

Existing test and manually verification on local cluster.

### DIFF
--- a/repl/src/main/scala/org/apache/livy/repl/ReplDriver.scala
+++ b/repl/src/main/scala/org/apache/livy/repl/ReplDriver.scala
@@ -93,24 +93,24 @@ class ReplDriver(conf: SparkConf, livyConf: RSCConf)
 
   override protected def createWrapper(msg: BaseProtocol.BypassJobRequest): BypassJobWrapper = {
     Kind(msg.jobType) match {
-      case PySpark() =>
+      case PySpark() if session.interpreter(PySpark()).isDefined =>
         new BypassJobWrapper(this, msg.id,
           new BypassPySparkJob(msg.serializedJob,
-            session.interpreter(PySpark()).asInstanceOf[PythonInterpreter]))
+            session.interpreter(PySpark()).get.asInstanceOf[PythonInterpreter]))
       case _ => super.createWrapper(msg)
     }
   }
 
   override protected def addFile(path: String): Unit = {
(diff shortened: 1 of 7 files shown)


ajbozarth:
On the code side this LGTM, but I didn't test out the functionality

repl/src/main/scala/org/apache/livy/repl/ReplDriver.scala
```diff
@@ -93,24 +93,24 @@ class ReplDriver(conf: SparkConf, livyConf: RSCConf)
...
+            session.interpreter(PySpark()).get.asInstanceOf[PythonInterpreter]))
       case _ => super.createWrapper(msg)
     }
   }
 
   override protected def addFile(path: String): Unit = {
-    if (!ClientConf.TEST_MODE) {
-      session.interpreter(PySpark()).asInstanceOf[PythonInterpreter].addFile(path)
+    if (!ClientConf.TEST_MODE && session.interpreter(PySpark()).isDefined) {
+      session.interpreter(PySpark()).get.asInstanceOf[PythonInterpreter].addFile(path)
```
meisam: Just a suggestion. I like this more:
```scala
if (!ClientConf.TEST_MODE) {
  session.interpreter(PySpark()).foreach { case interpreter: PythonInterpreter => interpreter.addFile(path) }
}
```

jerryshao: Sure, will change it.

jerryshao:
@ajbozarth @meisam do you have further comments? If not, I'm going to merge this, since it blocks our internal release.

jerryshao:
Merging to master.

PR #58, opened 2017-10-23
[LIVY-412][Server] Reject create session request if too more sessions are creating.

shenh062326:

In our cluster, livy server run with spark yarn cluster mode, when createSession request is too frequently, livyServer will start too more spark-submit child process, it will cause the machine oom. 

Reject the create session request when there are too many spark-submit child process. 

## How was this patch tested?

Add two testcase, test create interactive and batch session when reach the max creating sessions, the request should failed.


shenh062326:
The testcase org.apache.livy.repl.ReplDriverSuite is passed in my own machine.

rsc/src/main/java/org/apache/livy/rsc/ContextLauncher.java
```diff
@@ -374,6 +374,7 @@ public ChildProcess(RSCConf conf, Promise<?> promise, final Process childProc, F
         @Override
         public void run() {
           try {
+            RSCClientFactory.interactive_child_process += 1;
```
meisam: I think accesses to `RSCClientFactory.interactive_child_process` are causing a data race.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -117,14 +120,23 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
...
   post("/") {
-    val session = sessionManager.register(createSession(request))
-    // Because it may take some time to establish the session, update the last activity
-    // time before returning the session info to the client.
-    session.recordActivity()
-    Created(clientSessionView(session, request),
-      headers = Map("Location" ->
-        (getRequestPathInfo(request) + url(getSession, "id" -> session.id.toString))))
+    if (shouldRejectCreateSession) {
+      BadRequest("Reject, too more creating sessions!")
```
meisam: This might be a better message:
"Reject, too many sessions are being created!"

rsc/src/main/java/org/apache/livy/rsc/ContextLauncher.java
meisam: Also, `interactive_child_process` does not conform to Java naming standards. `interactiveSessionCount` could be a better name.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -117,14 +120,23 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
     }
   }
 
+  def shouldRejectCreateSession(): Boolean = {
+    val creatingSession = RSCClientFactory.interactive_child_process + SessionServlet.batch_child_process
+    creatingSession > livyConf.getInt(LivyConf.MAX_CREATE_SESSION)
```
meisam: I don't think this is a big issue, but I am pointing it out anyways:

`shouldRejectCreateSession` is not executed atomically, which means by the time the `creatingSession > livyConf.getInt(LivyConf.MAX_CREATE_SESSION)` is being executed, the values for `interactive_child_process` and `interactive_child_process` can change.

shenh062326:
@meisam , thanks for comment, In our cluster, we can't  simply throttle down create session requests. 
 I will update the code.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -117,14 +121,24 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
     }
   }
 
+  def shouldRejectCreateSession(): Boolean = {
+    val creatingSession = RSCClientFactory.interactiveSessionChildProcess.get() +
```
ajbozarth: this naming is confusing, maybe a better name could be `totalChildProceses`

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -117,14 +121,24 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
     }
   }
 
+  def shouldRejectCreateSession(): Boolean = {
```
ajbozarth: This isn't necessarily a bad name, but I feel like it could be better. Like `tooManySessions` or `overMaxSessionCount` could describe the actual return value better.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -117,14 +121,24 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
...
   post("/") {
-    val session = sessionManager.register(createSession(request))
-    // Because it may take some time to establish the session, update the last activity
-    // time before returning the session info to the client.
-    session.recordActivity()
-    Created(clientSessionView(session, request),
-      headers = Map("Location" ->
-        (getRequestPathInfo(request) + url(getSession, "id" -> session.id.toString))))
+    if (shouldRejectCreateSession) {
+      BadRequest("Reject, too many sessions are being created!")
```
ajbozarth: The english here is off, this should either be `Rejected, too many ...` or just `Too many ...`

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
```diff
@@ -85,6 +86,18 @@ object BatchSession extends Logging {
       val file = resolveURIs(Seq(request.file), livyConf)(0)
       val sparkSubmit = builder.start(Some(file), request.args)
 
+      Utils.startDaemonThread(s"ContextLauncher-$id") {
+        SessionServlet.batchSessionChildProcess.incrementAndGet()
+        try {
+          sparkSubmit.waitFor() match {
```
ajbozarth: Will adding this `waitFor()` cause any delay? I'm interested in how adding this chunk of code affects this function.

server/src/main/scala/org/apache/livy/LivyConf.scala
```diff
@@ -148,6 +148,8 @@ object LivyConf {
   val SESSION_TIMEOUT = Entry("livy.server.session.timeout", "1h")
   // How long a finished session state will be kept in memory
   val SESSION_STATE_RETAIN_TIME = Entry("livy.server.session.state-retain.sec", "600s")
+  // Max creating session in livyServer
+  val MAX_CREATE_SESSION = Entry("livy.server.max.create.session", 100)
```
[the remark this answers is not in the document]
shenh062326: In our own cluster, when I send about 120 create session requestes at once, the livyServer machine oom.  In fact, it's depend on the machine performance.

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
shenh062326: The code is like in SparkProcApp:
https://github.com/apache/incubator-livy/blob/bef5b4f6d8ac931613027cf50cbc44be3be2e107/server/src/main/scala/org/apache/livy/utils/SparkProcApp.scala#L53

shenh062326:
@ajbozarth thanks for your review, I will update the code according to the comment.

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
shenh062326: The code is like in SparkProcApp.
https://github.com/apache/incubator-livy/blob/bef5b4f6d8ac931613027cf50cbc44be3be2e107/server/src/main/scala/org/apache/livy/utils/SparkProcApp.scala#L53

jerryshao:
@shenh062326 are you using yarn cluster mode? spark-submit process is a short-lived process, it may not last very long.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -17,17 +17,21 @@
...
 import scala.concurrent.duration._
 
 import org.apache.livy.{LivyConf, Logging}
+import org.apache.livy.rsc.RSCClientFactory
 import org.apache.livy.sessions.{Session, SessionManager}
 import org.apache.livy.sessions.Session.RecoveryMetadata
 
-object SessionServlet extends Logging
+object SessionServlet extends Logging {
+  val batchSessionChildProcesses = new AtomicInteger
```
jerryshao: Why dont we maintain this in `BatchSession`?

jerryshao:
@meisam rejection is another kind of throttle, what's your idea of throttle?

@codingcat, I saw you have a similar JIRA about this issue, what do you think about the current proposal?

shenh062326:
@jerryshao , I use yarn cluster mode.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
shenh062326: Maintain in BatchSession looks more suitable.

jerryshao:
Beside this solution, what I can think of is to delay the session creation for throttling. Since POST request is async, so we could delay the session creation if reaching to threshold. Client side should wait and query until and session is created, what do you think of this solution?

shenh062326:
@jerryshao The solution seems more friendly to client. But acturally we have have more than one livyServer, if too many create request send to one livy server, client may wait more time.

jerryshao:
I get your point. I'm OK with current solution.

jerryshao:
Please address the comments and add unit test to verify your code.

shenh062326:
OK, I will add a unit test.

meisam:
> rejection is another kind of throttle, what's your idea of throttle?

@jerryshao 
I was thinking we could delay the session creation, because, at least for our workloads, there are short bursts in requests to Livy. Delaying the session creation jut by 1 or 2 minutes solves those problems.

jerryshao:
Yes, but for the scenario mentioned by @shenh062326 , they have multiple Livy instances,  if one request is rejected by one Livy instance, it could quickly shifted to another instance. That seems better than unnecessary waiting.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -24,9 +24,12 @@ import scala.concurrent._
 import scala.concurrent.duration._
 
 import org.apache.livy.{LivyConf, Logging}
+import org.apache.livy.rsc.RSCClientFactory
+import org.apache.livy.server.batch.BatchSession
 import org.apache.livy.sessions.{Session, SessionManager}
 import org.apache.livy.sessions.Session.RecoveryMetadata
 
+
```
ajbozarth: unnecessary extra line

server/src/test/scala/org/apache/livy/server/batch/BatchServletSpec.scala
```diff
@@ -80,6 +80,11 @@ class BatchServletSpec extends BaseSessionServletSpec[BatchSession, BatchRecover
         batch should be (defined)
       }
 
+      val tmp = servlet.livyConf.getInt(LivyConf.MAX_CREATING_SESSION)
```
ajbozarth: I think it might be better to just place this test somewhere we don't have to save and re-set the conf value

rsc/src/main/java/org/apache/livy/rsc/RSCClientFactory.java
```diff
@@ -36,6 +36,7 @@
 
   private final AtomicInteger refCount = new AtomicInteger();
   private RpcServer server = null;
+  public static AtomicInteger interactiveSessionChildProcesses = new AtomicInteger();
```
ajbozarth: If we're putting the batch version of this in `BatchSession`, why wouldn't this be in `InteractiveSession`? Is there a reason to keep this in `RSCClientFactory`?

server/src/test/scala/org/apache/livy/server/interactive/FailedInteractiveSessionServletSpec.scala
```diff
@@ -0,0 +1,107 @@
...
+    val accessManager = new AccessManager(livyConf)
+    new BatchSessionServlet(
+      new BatchSessionManager(livyConf, sessionStore, Some(Seq.empty)),
+      sessionStore,
+      livyConf,
+      accessManager)
+  }
+
+  describe("Interactive Servlet") {
+    it("should failed to create interactive session") {
```
ajbozarth: I there a reason this test couldn't have been put in `InteractiveSessionServletSpec` rather than creating a whole new class for it?

rsc/src/main/java/org/apache/livy/rsc/RSCClientFactory.java
shenh062326: RSCClientFactory.java in project livy-rsc, InteractiveSession in project livy-server, project livy-server depend on livy-rsc, but livy-rsc not depend on livy-server, so if we keep interactiveSessionChildProcesses in InteractiveSession, we can't call it in RSCClientFactory.java.

shenh062326:
@ajbozarth thanks for you comment, I will update the testcase.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
shenh062326: I will remove it.

server/src/main/scala/org/apache/livy/LivyConf.scala
```diff
@@ -148,6 +148,8 @@ object LivyConf {
   val SESSION_TIMEOUT = Entry("livy.server.session.timeout", "1h")
   // How long a finished session state will be kept in memory
   val SESSION_STATE_RETAIN_TIME = Entry("livy.server.session.state-retain.sec", "600s")
+  // Max creating session in livyServer
+  val MAX_CREATING_SESSION = Entry("livy.server.max.creating.session", 100)
```
jerryshao: Can we please rename to "livy.server.session.max-creation"?

rsc/src/main/java/org/apache/livy/rsc/RSCClientFactory.java
jerryshao: I would suggest you don't make class variable public, instead adding a method to call this, like `int childProcesses()`. Also can you shorten the name of this variable?

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
```diff
@@ -42,6 +44,7 @@ case class BatchRecoveryMetadata(
 
 object BatchSession extends Logging {
   val RECOVERY_SESSION_TYPE = "batch"
+  val batchSessionChildProcesses = new AtomicInteger
```
jerryshao: Also don't expose this variable, instead using a method.

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
```diff
@@ -85,6 +88,18 @@ object BatchSession extends Logging {
       val file = resolveURIs(Seq(request.file), livyConf)(0)
       val sparkSubmit = builder.start(Some(file), request.args)
 
+      Utils.startDaemonThread(s"ContextLauncher-$id") {
```
jerryshao: Can we rename this name of thread to `batch-session-process-$id`?

server/src/test/scala/org/apache/livy/server/batch/BatchServletSpec.scala
```diff
@@ -151,6 +146,29 @@ class BatchServletSpec extends BaseSessionServletSpec[BatchSession, BatchRecover
       view.appInfo shouldEqual appInfo
       view.log shouldEqual log
     }
-  }
 
-}
+    it("should failed create session when too many creating session") {
```
ajbozarth: grammar nit: `should fail session creation when max session creation is hit

shenh062326:
@jerryshao @meisam thanks for your comment, I will update the code.

server/src/main/scala/org/apache/livy/LivyConf.scala
```diff
@@ -148,6 +148,8 @@ object LivyConf {
   val SESSION_TIMEOUT = Entry("livy.server.session.timeout", "1h")
   // How long a finished session state will be kept in memory
   val SESSION_STATE_RETAIN_TIME = Entry("livy.server.session.state-retain.sec", "600s")
+  // Max creating session in livyServer
+  val MAX_CREATING_SESSION = Entry("livy.server.session.max-creation", 100)
```
ajbozarth: I think the val name should match the conf name: `SESSION_MAX_CREATION`

jerryshao:
@shenh062326 can you please make the travis pass the unit test.

jerryshao:
You can close and reopen PR to trigger the travis test.

shenh062326:
@jerryshao  👌

jerryshao:
I think your added test is a bit flaky, can you please fix it, thanks!

shenh062326:
The testcase always succeeded in my own machine, I will add some log to find the failed reason.
BatchServletSpec:
Batch Servlet
- should create and tear down a batch (1 second, 146 milliseconds)
- should respect config black list (4 milliseconds)
- should show session properties (10 milliseconds)
- should fail session creation when max session creation is hit (20 milliseconds)
InteractiveSessionServletSpec:
- should setup and tear down an interactive session (68 milliseconds)
- should show session properties (11 milliseconds)
- should failed create session when too many creating session  (20 milliseconds)
CreateBatchRequestSpec:
CreateBatchRequest
- should have default values for fields after deserialization (15 milliseconds)
AccessManagerSuite:
- access permission (3 milliseconds)
- wildcard access permission (1 millisecond)
- default allowed users (2 milliseconds)
- configured users are not in the allowed list (2 milliseconds)
Run completed in 1 minute, 6 seconds.
Total number of tests run: 133
Suites: completed 23, aborted 0
Tests: succeeded 133, failed 0, canceled 0, ignored 0, pending 0
All tests passed.

rsc/src/main/java/org/apache/livy/rsc/ContextLauncher.java
```diff
@@ -374,6 +374,8 @@ public ChildProcess(RSCConf conf, Promise<?> promise, final Process childProc, F
         @Override
         public void run() {
           try {
+            RSCClientFactory.childProcesses().incrementAndGet();
+            System.out.println("processes:" + RSCClientFactory.childProcesses().get());
```
jerryshao: Can you please remove all the `println` if the flaky test is fixed, thanks!

shenh062326: Yes, I have remove debug log.

rsc/src/main/java/org/apache/livy/rsc/RSCClientFactory.java
```diff
@@ -36,6 +36,12 @@
 
   private final AtomicInteger refCount = new AtomicInteger();
   private RpcServer server = null;
+  // interactive session child processes number
+  public static AtomicInteger iscpn = new AtomicInteger();
```
jerryshao: nit: I think you don't have to make it `public` now.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -117,14 +119,24 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
     }
   }
 
+  def tooManySessions(): Boolean = {
+    val totalChildProceses = RSCClientFactory.childProcesses().get() +
+      BatchSession.childProcesses.get()
+    totalChildProceses >= livyConf.getInt(LivyConf.SESSION_MAX_CREATION)
```
jerryshao: Should it be better to change to `>`?

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
```diff
@@ -42,6 +44,12 @@ case class BatchRecoveryMetadata(
 
 object BatchSession extends Logging {
   val RECOVERY_SESSION_TYPE = "batch"
+  // batch session child processes number
+  val bscpn = new AtomicInteger
```
jerryshao: ditto.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
shenh062326: If change to '>' , totalChildProceses can reach SESSION_MAX_CREATION+1.

jerryshao: I see, I understand your point.

jerryshao:
Thank @shenh062326 , let me merge to master.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -117,14 +119,24 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
...
   post("/") {
-    val session = sessionManager.register(createSession(request))
-    // Because it may take some time to establish the session, update the last activity
-    // time before returning the session info to the client.
-    session.recordActivity()
-    Created(clientSessionView(session, request),
-      headers = Map("Location" ->
-        (getRequestPathInfo(request) + url(getSession, "id" -> session.id.toString))))
+    if (tooManySessions) {
+      BadRequest("Rejected, too many sessions are being created!")
```
taoliseki: Hi @shenh062326,

I just started looking at this throttling change and have a quick question here. Isn't TooManyRequests better than BadRequest here?

PR #63, opened 2017-11-14
[LIVY-414][SERVER] Support setting env variables during session creation

jerryshao:

In the current Livy, we don't support setting environment variables per session, Livy server by default will honor env variables set in spark-env.sh during SparkSubmit process launches. This is not flexible enough if we want to set different env variables per session. So here propose to support this feature.

## How was this patch tested?

new UTs.

### DIFF
--- a/api/src/main/java/org/apache/livy/LivyClientBuilder.java
+++ b/api/src/main/java/org/apache/livy/LivyClientBuilder.java
@@ -23,6 +23,7 @@ import java.io.Reader;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
 import java.util.ServiceLoader;
@@ -36,6 +37,8 @@ public final class LivyClientBuilder {
   public static final String LIVY_URI_KEY = "livy.uri";
 
   private final Properties config;
+
+  private final Map<String, String> env;
 
   /**
    * Creates a new builder that will automatically load the default Livy and Spark configuration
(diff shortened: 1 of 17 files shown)


vanzin:
This kinda assumes Livy will always start client applications as child processes, which seems like an assumption that should be avoided.

Also, is this safe against command injections? If I set an env variable to `$(command to upload /some/sensitive/file)`, what will happen?

Finally, the change in `LivyClientFactory.java` also breaks API compatibility.

jerryshao:
The purpose of this PR is to support different environment variables per session, but without this I cannot figure out another solution.

Yes it's true if we support launching SparkSubmit in one JVM then this will not be worked.

>Finally, the change in LivyClientFactory.java also breaks API compatibility.

`LivyClientFactory` is Private, user will not touch it AFAIK.

vanzin:
> LivyClientFactory is Private, user will not touch it AFAIK.

Right, forgot about that.

Another question: what env variables are you thinking about here?

e.g., if a user sets `HADOOP_CONF_DIR` and Livy respects that, it's possible that the spark-submit launched by Livy will upload files to HDFS that Livy has access to, but the user shouldn't. That's another potential security risk.

jerryshao:
For example some Python or R related env variables, like different PYTHONPATH/PYSPARK_PYTHON  per session.

vanzin:
Spark 2 at least has `spark.pyspark.python` which can be set in `SparkConf`. For others, one could use `spark.yarn.appMasterEnv.*` and `spark.executorEnv.*`, but the former assumes YARN cluster mode.

If really exposing this, you'll probably need to add some kind of configurable whitelist, and sanitize the user input, otherwise it's a bit risky.

jerryshao:
Yes, that's what I currently did, which makes me think of supporting env variables per session.

Thanks a lot for your comment, let me think a bit about this.

jerryshao:
Close this for now.

PR #66, opened 2017-11-22
[LIVY-299][REPL] Output multiple lines in one statement block

jerryshao:

Livy Scala Interpreter doesn't support output multiple lines in one code block, this will miss some outputs, so here change the logic to support it.

## How was this patch tested?

Add new unit tests.

### DIFF
--- a/repl/src/main/scala/org/apache/livy/repl/AbstractSparkInterpreter.scala
+++ b/repl/src/main/scala/org/apache/livy/repl/AbstractSparkInterpreter.scala
@@ -274,7 +274,7 @@ abstract class AbstractSparkInterpreter extends Interpreter with Logging {
                   val lastRet = s.values.getOrElse(TEXT_PLAIN, "").asInstanceOf[String]
                   val currRet = e.values.getOrElse(TEXT_PLAIN, "").asInstanceOf[String]
                   if (lastRet.nonEmpty && currRet.nonEmpty) {
-                    Interpreter.ExecuteSuccess(TEXT_PLAIN -> s"$lastRet\n$currRet")
+                    Interpreter.ExecuteSuccess(TEXT_PLAIN -> s"$lastRet$currRet")
                   } else if (lastRet.nonEmpty) {
                     Interpreter.ExecuteSuccess(TEXT_PLAIN -> lastRet)
                   } else if (currRet.nonEmpty) {
@@ -344,7 +344,7 @@ abstract class AbstractSparkInterpreter extends Interpreter with Logging {
   }
 
   private def readStdout() = {
-    val output = outputStream.toString("UTF-8").trim
+    val output = outputStream.toString("UTF-8")
     outputStream.reset()
 
     output
--- a/repl/src/main/scala/org/apache/livy/repl/AbstractSparkInterpreter.scala
+++ b/repl/src/main/scala/org/apache/livy/repl/AbstractSparkInterpreter.scala
@@ -262,8 +262,34 @@ abstract class AbstractSparkInterpreter extends Interpreter with Logging {
           case Interpreter.ExecuteError(_, _, _) =>
             result
 
-          case _ =>
-            executeLines(tail, result)
+          case Interpreter.ExecuteAborted(_) =>
+            result
+
+          case Interpreter.ExecuteSuccess(e) =>
+            val mergedRet = resultFromLastLine match {
+              case Interpreter.ExecuteSuccess(s) =>
+                // Because of SparkMagic related specific logic, so we will only merge text/plain
+                // result. For any magic related output, still follow the old way.
+                if (s.values.contains(TEXT_PLAIN) && e.values.contains(TEXT_PLAIN)) {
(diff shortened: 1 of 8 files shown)


jerryshao:
@zjffdu , would you please take a chance to review this PR, thanks!

repl/src/test/scala/org/apache/livy/repl/PythonInterpreterSpec.scala
```diff
@@ -65,6 +65,17 @@ abstract class PythonBaseInterpreterSpec extends BaseInterpreterSpec {
...
   }
 
+  it should "get multiple outputs in one block" in withInterpreter { interpreter =>
+    val response = interpreter.execute(
+      """
+        |print("1")
+        |print("2")
+      """.stripMargin)
+    response should equal(Interpreter.ExecuteSuccess(
+      TEXT_PLAIN -> "1\n2"
```
zjffdu: Shouldn't it be `12` ?, otherwise code `println(1)\nprintln(2)` would result produce 3 `\n`

repl/src/main/scala/org/apache/livy/repl/AbstractSparkInterpreter.scala
```diff
@@ -262,8 +262,34 @@ abstract class AbstractSparkInterpreter extends Interpreter with Logging {
...
+          case Interpreter.ExecuteSuccess(e) =>
+            val mergedRet = resultFromLastLine match {
+              case Interpreter.ExecuteSuccess(s) =>
+                // Because of SparkMagic related specific logic, so we will only merge text/plain
+                // result. For any magic related output, still follow the old way.
+                if (s.values.contains(TEXT_PLAIN) && e.values.contains(TEXT_PLAIN)) {
+                  val lastRet = s.values.getOrElse(TEXT_PLAIN, "").asInstanceOf[String]
+                  val currRet = e.values.getOrElse(TEXT_PLAIN, "").asInstanceOf[String]
+                  if (lastRet.nonEmpty && currRet.nonEmpty) {
+                    Interpreter.ExecuteSuccess(TEXT_PLAIN -> s"$lastRet\n$currRet")
```
zjffdu: I think we don't need to add `\n`

repl/src/test/scala/org/apache/livy/repl/PythonInterpreterSpec.scala
jerryshao: what's the meaning?

repl/src/main/scala/org/apache/livy/repl/AbstractSparkInterpreter.scala
jerryshao: If we don't have `\n`, how to separate different line's output?

repl/src/test/scala/org/apache/livy/repl/PythonInterpreterSpec.scala
jerryshao: For the python outputs, it is naturally supported by python itself, we didn't change anything.

zjffdu: Sorry, my mistake. python's `print` would add `\n` at the end of output, just like scala's `println`. Could you add the same test for scala. The following scala code should output `12` in my opinion.
```
print(1)
print(2)
```

jerryshao: I don't think so, naturally you separate this code block into two lines, so it should print out `1\n2`, unless you change to `print(1);print(2)`.

jerryshao: @zjffdu can you please review again about the new changes, will it break Zeppelin?

zjffdu:
LGTM

jerryshao:
Thanks @zjffdu , merging to master.

gm310509:
IMHO, I think that this severely breaks Zeppelin,
My organisation has just upgraded to livy 0.5 and now the output of every line of scala code is being echoed to the browser.

for example this %livy code:
```scala
val x = 1
val y = 2
println(s"x + y = ${x + y}")
```
produces this output:
```
x: Int = 1
y: Int = 2
x + y = 3
```

The end result required to get a clean output is a cumbersome set of editing to create multiple paragraphs that set up output for display in subsequent paragraphs that can only contain println statements and hiding of the output of the "setup" paragraphs. Also, refreshing the display requires running both the "setup" and the "output" paragraphs. Previously, the "setup output" and "display output" logic could be combined in one paragraph. i.e. there was no need to create nor run separate paragraphs to get a clean output.

zjffdu:
Please file ticket to disable the repl output like zeppelin spark interpreter.

gm310509:
Sure, happy to do so, but I need a bit of guidance. Where can I go to file a ticket?

gm310509:
Never mind, someone else appears to have already done it: [Livy-507](https://issues.apache.org/jira/browse/LIVY-507)

PR #67, opened 2017-11-28
[LIVY-415][CORE][Followup] Fix static object creation issue for SessionState

jerryshao:

In LIVY-415, we changed `SessionState` to static object instead of case class. This is OK for `SessionState`, but for `FinishedSessionState`, it requires `time` field to be set during object creation. Using static object will never reflect the actual `time` when object is created. So here propose to fix it.

CC @aa8y please take a review, this is introduced in your changes.

## How was this patch tested?

Existing tests.

### DIFF
--- a/core/src/main/scala/org/apache/livy/sessions/SessionState.scala
+++ b/core/src/main/scala/org/apache/livy/sessions/SessionState.scala
@@ -37,9 +37,9 @@ object SessionState {
     case "running" => Running
     case "busy" => Busy
     case "shutting_down" => ShuttingDown
-    case "error" => Error
-    case "dead" => Dead
-    case "success" => Success
+    case "error" => Error()
+    case "dead" => Dead()
+    case "success" => Success()
     case _ => throw new IllegalArgumentException(s"Illegal session state: $s")
   }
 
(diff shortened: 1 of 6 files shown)


aa8y:
@jerryshao I am sorry I did not see this before, but I am not sure whether the fix was required. The way we had it before, we could have created a different instance with a different time and we had a singleton with the default time.

jerryshao:
A singleton's time cannot actually reflect the time of state transition, we don't know when this singleton is created, also in your previous patch, you change all the `SessionState` to singleton, it is not correct, once singleton is created, it will never change the time when you reuse it, it is the only copy in JVM, so using old state of reflect current session is not correct.

aa8y:
I understand that a singleton is a one time creation and does not reflect changes. That was the whole reason why I made that change. But that never restricted any of the `FinishedSessionState` objects being instantiated without the default values for the arguments. As an example, let's look at `Error`.

```
scala> :paste
// Entering paste mode (ctrl-D to finish)

sealed abstract class SessionState(val state: String, val isActive: Boolean) {
  override def toString: String = state
}
class FinishedSessionState(
  override val state: String,
  override val isActive: Boolean,
  val time: Long
) extends SessionState(state, isActive)
case class Error(override val time: Long) extends FinishedSessionState("error", true, time)
object Error extends Error(System.nanoTime)

// Exiting paste mode, now interpreting.

defined class SessionState
defined class FinishedSessionState
defined class Error
defined object Error

scala> Error
res0: Error.type = error

scala> Error.time
res1: Long = 2456992903279170

scala> Error(123L)
res2: Error = error

scala> res2.time
res3: Long = 123
```
Here you can see we have a default singleton for `Error` and we created a new object with a different value for `time`.

jerryshao:
Yes, true. You class and object definition is correct. But the way you're using it is not correct. For example: [here](https://github.com/jerryshao/incubator-livy/blob/761e574f6609b760c59f0ad890b0d895e85b5226/server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala#L181) you're using singleton object `Success` instead of class object `Success()`, so the time of `Success` state is not as expected, this singleton object may already be initialized before, using the `time` from this object cannot actually reflect the session's success time.

aa8y:
Okay. Now I get it. Makes sense.

PR #71, opened 2018-01-02
[LIVY-429] preserve url fragment in spark.yarn.dist.archives

broartem:

Preserve url fragment identifier in spark.yarn.dist.archives setting when there is no schema specified. Current code truncates url fragment, as a result Spark just copies archives to job local dir without extracting it.

## How was this patch tested?

Added case to existing unit test


jerryshao:
LGTM, thanks for the fix.

CC @ajbozarth would you please also take a look.

ajbozarth:
LGTM too, also @jerryshao now that I'm back if there's anything else I should take a look at tag me

jerryshao:
Merging to master. Thanks for the fix.

PR #72, opened 2018-01-11
[LIVY-104] [Core] Update Livy build to Scala 2.11

ajbozarth:
[LIVY-104](https://issues.apache.org/jira/browse/LIVY-104)

Livy currently builds using Scala 2.10 which has been EOL for a couple years. This update switches the build to use 2.11 instead.
This does not drop Livy support for Scala 2.10, it only updates the version used to build Livy itself.
We are holding off updating to Scala 2.12 since Livy still supports Spark 1.6, which doesn't support 2.12 and such an update would break UTs for 1.6, any update to 2.12 will have to wait until Spark 1.6 support is dropped.

This update also addresses any new deprecation warning raised in the update as well as updates various ITs to pass.

This locally passes both UTs and ITs build withe all supported versions of Spark.

### DIFF
--- a/integration-test/src/main/scala/org/apache/livy/test/framework/BaseIntegrationTestSuite.scala
+++ b/integration-test/src/main/scala/org/apache/livy/test/framework/BaseIntegrationTestSuite.scala
@@ -54,7 +54,7 @@ abstract class BaseIntegrationTestSuite extends FunSuite with Matchers with Befo
   }
 
   protected def restartLivy(): Unit = {
-    val f = future {
+    val f = Future {
       cluster.stopLivy()
       cluster.runLivy()
     }
--- a/integration-test/src/test/scala/org/apache/livy/test/InteractiveIT.scala
+++ b/integration-test/src/test/scala/org/apache/livy/test/InteractiveIT.scala
@@ -75,7 +75,7 @@ class InteractiveIT extends BaseIntegrationTestSuite {
       // This is important because if YARN app state is killed, Spark history is not archived.
       val appId = s.appId()
       s.stop()
-      eventually(timeout(15 seconds), interval(5 seconds)) {
+      eventually(timeout(5 seconds), interval(1 seconds)) {
         val appReport = cluster.yarnClient.getApplicationReport(appId)
         appReport.getYarnApplicationState() shouldEqual YarnApplicationState.FINISHED
       }
--- a/integration-test/src/test/scala/org/apache/livy/test/InteractiveIT.scala
+++ b/integration-test/src/test/scala/org/apache/livy/test/InteractiveIT.scala
@@ -34,6 +34,7 @@ class InteractiveIT extends BaseIntegrationTestSuite {
   test("basic interactive session") {
     withNewSession(Spark) { s =>
       s.run("val sparkVersion = sc.version").result().left.foreach(info(_))
+      s.run("val scalaVersion = util.Properties.versionString").result().left.foreach(info(_))
       s.run("1+1").verifyResult("res0: Int = 2\n")
       s.run("""sc.getConf.get("spark.executor.instances")""").verifyResult("res1: String = 1\n")
       s.run("val sql = new org.apache.spark.sql.SQLContext(sc)").verifyResult(
@@ -74,8 +75,10 @@ class InteractiveIT extends BaseIntegrationTestSuite {
       // This is important because if YARN app state is killed, Spark history is not archived.
       val appId = s.appId()
       s.stop()
-      val appReport = cluster.yarnClient.getApplicationReport(appId)
-      appReport.getYarnApplicationState() shouldEqual YarnApplicationState.FINISHED
+      eventually(timeout(15 seconds), interval(5 seconds)) {
+        val appReport = cluster.yarnClient.getApplicationReport(appId)
+        appReport.getYarnApplicationState() shouldEqual YarnApplicationState.FINISHED
+      }
     }
   }
 
--- a/integration-test/src/test/scala/org/apache/livy/test/JobApiIT.scala
+++ b/integration-test/src/test/scala/org/apache/livy/test/JobApiIT.scala
@@ -168,10 +168,10 @@ class JobApiIT extends BaseIntegrationTestSuite with BeforeAndAfterAll with Logg
 
     val jobs = Seq(
       new ScalaEcho("abcde"),
-      new ScalaEcho(Seq(1, 2, 3, 4)),
+      new ScalaEcho(IndexedSeq(1, 2, 3, 4)),
       new ScalaEcho(Map(1 -> 2, 3 -> 4)),
       new ScalaEcho(ValueHolder("abcde")),
-      new ScalaEcho(ValueHolder(Seq(1, 2, 3, 4))),
+      new ScalaEcho(ValueHolder(IndexedSeq(1, 2, 3, 4))),
       new ScalaEcho(Some("abcde"))
     )
 
--- a/server/src/main/scala/org/apache/livy/server/JsonServlet.scala
+++ b/server/src/main/scala/org/apache/livy/server/JsonServlet.scala
@@ -133,7 +133,7 @@ abstract class JsonServlet extends ScalatraServlet with ApiFormats with FutureSu
   }
 
   private def toJson(obj: Any): Any = {
-    if (obj != null && obj != ()) {
+    if (obj != null && obj != (())) {
       mapper.writeValueAsBytes(obj)
     } else {
       null
--- a/pom.xml
+++ b/pom.xml
@@ -98,8 +98,8 @@
     <py4j.version>0.9</py4j.version>
     <scala-2.10.version>2.10.4</scala-2.10.version>
     <scala-2.11.version>2.11.8</scala-2.11.version>
-    <scala.binary.version>2.10</scala.binary.version>
-    <scala.version>${scala-2.10.version}</scala.version>
+    <scala.binary.version>2.11</scala.binary.version>
+    <scala.version>${scala-2.11.version}</scala.version>
     <scalatest.version>2.2.4</scalatest.version>
     <scalatra.version>2.3.0</scalatra.version>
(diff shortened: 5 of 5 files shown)


server/src/main/scala/org/apache/livy/server/JsonServlet.scala
```diff
@@ -133,7 +133,7 @@ abstract class JsonServlet extends ScalatraServlet with ApiFormats with FutureSu
   }
 
   private def toJson(obj: Any): Any = {
-    if (obj != null && obj != ()) {
+    if (obj != null && obj != (())) {
```
ajbozarth: addresses deprecation warning

integration-test/src/main/scala/org/apache/livy/test/framework/BaseIntegrationTestSuite.scala
```diff
@@ -54,7 +54,7 @@ abstract class BaseIntegrationTestSuite extends FunSuite with Matchers with Befo
   }
 
   protected def restartLivy(): Unit = {
-    val f = future {
+    val f = Future {
```
ajbozarth: addresses deprecation warning

integration-test/src/test/scala/org/apache/livy/test/InteractiveIT.scala
```diff
@@ -34,6 +34,7 @@ class InteractiveIT extends BaseIntegrationTestSuite {
   test("basic interactive session") {
     withNewSession(Spark) { s =>
       s.run("val sparkVersion = sc.version").result().left.foreach(info(_))
+      s.run("val scalaVersion = util.Properties.versionString").result().left.foreach(info(_))
```
ajbozarth: Added as a followup to LIVY-399 ( #49 ) which removed the double run of ITs on both 2.10 and 2.11, now only running the tests with the version that corresponds to the build's Spark version (2.10 for 1.6 and 2.11 for 2.X).

integration-test/src/test/scala/org/apache/livy/test/InteractiveIT.scala
```diff
@@ -74,8 +75,10 @@ class InteractiveIT extends BaseIntegrationTestSuite {
       // This is important because if YARN app state is killed, Spark history is not archived.
       val appId = s.appId()
       s.stop()
-      val appReport = cluster.yarnClient.getApplicationReport(appId)
-      appReport.getYarnApplicationState() shouldEqual YarnApplicationState.FINISHED
+      eventually(timeout(15 seconds), interval(5 seconds)) {
```
ajbozarth: @jerry I added this to address a consistently failing test on my local machine, I did not see this failure in your previous PR but got it consistently on my local machine after the update. I'm open to removing this or changing the retry if you have suggestions

integration-test/src/test/scala/org/apache/livy/test/JobApiIT.scala
```diff
@@ -168,10 +168,10 @@ class JobApiIT extends BaseIntegrationTestSuite with BeforeAndAfterAll with Logg
 
     val jobs = Seq(
       new ScalaEcho("abcde"),
-      new ScalaEcho(Seq(1, 2, 3, 4)),
+      new ScalaEcho(IndexedSeq(1, 2, 3, 4)),
       new ScalaEcho(Map(1 -> 2, 3 -> 4)),
       new ScalaEcho(ValueHolder("abcde")),
-      new ScalaEcho(ValueHolder(Seq(1, 2, 3, 4))),
+      new ScalaEcho(ValueHolder(IndexedSeq(1, 2, 3, 4))),
```
ajbozarth: @vanzin I know it'd been a while since you were in the Livy code, but since you originally wrote this test I wanted your input. You originally had this test only run on Scala 2.10 compiles and not 2.11 (back when we ran both for all builds), but never gave a reason for that choice outside of needing to match the Scala version between Livy build and Spark build. 
If left as `Seq` after the version update these line consistently cause the following job to catastrophically fail on a `head of empty list` exception. I'm not sure why this happens only in this IT and not actually use and only when using a `LinearSeq`

Is there a reason not to change this? I worked with @jodersky on debugging this last spring without ever finding a root cause.

jerryshao:
Thanks @ajbozarth , I will review it soon.

integration-test/src/test/scala/org/apache/livy/test/InteractiveIT.scala
jerryshao: let's rerun this on travis several times to see if it is flaky. We can increase the timeout or remove it if it is too flaky.

integration-test/src/test/scala/org/apache/livy/test/InteractiveIT.scala
jerryshao: LGTM.

integration-test/src/test/scala/org/apache/livy/test/InteractiveIT.scala
ajbozarth: with the eventually it is always successful, there was just no retry or wait before, should I remove this and let the tests rerun then?

jerryshao: If `eventually` can fix the issue, then let's leave it. One concern is that 5 seconds `interval` seems to long to wait, we can make the check more frequently.

ajbozarth: ok, I'll update it in the morning then

ajbozarth: updated to 5s t/o at 1s intervals, still passes

integration-test/src/test/scala/org/apache/livy/test/JobApiIT.scala
vanzin: I might be totally off here since I really don't remember details. But:

> outside of needing to match the Scala version between Livy build and Spark build

That makes a lot of sense if this test does serialization between Spark and Livy, because the serialized format for these classes might have changed between 2.10 and 2.11.

Just a wild guess though.

vanzin: Although from other comments of yours you are not mixing versions during ITs anymore, so I can't really explain the problem without seeing more info.

ajbozarth: Based on your comment about serialization changes I'll take another look and see if the Spark code might be using 2.10 still instead of respecting the new build version.

ajbozarth: Ok so after some checking I'll have to re-look at this issue from scratch. At some point the version check that the tests do stopped doing its job. Currently we can turn off the version check and the test passes on all Spark versions with 2.10 and none of them with 2.11. In theory it should only pass with 1.6 on 2.10 and 2.x on 2.1, thus why the check stops the other tests from running. So somewhere it's using a Scala or Spark version different that the one it thinks it's using (according to the props/confs)

ajbozarth: So I did some digging into the whole versioning thing and found no new leads or info. @vanzin as for a short detail on how to reproduce the issue:
1. Update base pom.xml to Scala 2.11
2. Run ITs
3. This test fails with the following error: 
```
Reporter completed abruptly with an exception after receiving event: SuiteAborted(Ordinal(1, 73),Exception encountered when invoking run on a nested suite - head of empty list,JobApiIT,org.apache.livy.test.JobApiIT,Some(org.apache.livy.test.JobApiIT),Some(java.util.NoSuchElementException: head of empty list),Some(46036),Some(IndentedText(Exception encountered when invoking run on a nested suite - head of empty list,Exception encountered when invoking run on a nested suite - head of empty list,0)),Some(SeeStackDepthException),Some(org.apache.livy.test.JobApiIT),None,ScalaTest-main-running-DiscoverySuite,1516068713776).
java.util.NoSuchElementException: head of empty list
	at scala.collection.immutable.Nil$.head(List.scala:420)
	at scala.collection.immutable.Nil$.head(List.scala:417)
	at scala.collection.immutable.List.map(List.scala:277)
	at org.scalatest.tools.StringReporter$.stackTrace$1(StringReporter.scala:970)
	at org.scalatest.tools.StringReporter$.getStackTrace$1(StringReporter.scala:1002)
	at org.scalatest.tools.StringReporter$.stringsToPrintOnError(StringReporter.scala:1008)
	at org.scalatest.tools.StringReporter$.fragmentsForEvent(StringReporter.scala:760)
	at org.scalatest.tools.StringReporter.apply(StringReporter.scala:169)
	at org.scalatest.tools.PrintReporter.apply(PrintReporter.scala:141)
	at org.scalatest.DispatchReporter$Propagator$$anonfun$run$1.apply(DispatchReporter.scala:240)
	at org.scalatest.DispatchReporter$Propagator$$anonfun$run$1.apply(DispatchReporter.scala:239)
	at scala.collection.immutable.List.foreach(List.scala:381)
	at org.scalatest.DispatchReporter$Propagator.run(DispatchReporter.scala:239)
```
This seems to happen regardless of Spark version and (from some digging) is caused when Spark runs for the `Seq` jobs (which are created as `List`) but not when they are `IndexedSeq`s

If just changing it to `IndexedSeq` as a permanent workaround is ok with you guys then we should merge this, otherwise I can keep digging, but after the time I've already put in I'm not sure where else to dig

vanzin: Yeah, nothing pops up. Could be some weird issue with serializing some Scala classes with Kryo in Livy's environment. If changing the type works here go for it.

ajbozarth:
Thanks for the feedback @jerryshao and @vanzin if there no other comments I'll merge this later today once tests re-pass

ajbozarth:
Since Travis is down I may have to merge this tomorrow

ajbozarth:
Now that tests passed I'll merge this

PR #75, opened 2018-01-24
[MINOR] Addressed Docs dependency security vulnerability

ajbozarth:
Addressed issue raised on mailing list:
https://www.mail-archive.com/dev@livy.incubator.apache.org/msg00295.html

More info on the vulnerability:
https://snyk.io/blog/nokogiri-xxe-vulnerabilities/

### DIFF
--- a/docs/Gemfile.lock
+++ b/docs/Gemfile.lock
@@ -71,7 +71,7 @@ GEM
       typhoeus (~> 0.7)
     html-pipeline (2.6.0)
       activesupport (>= 2)
-      nokogiri (>= 1.4)
+      nokogiri (>= 1.5.4)
     i18n (0.8.6)
     jekyll (3.4.5)
       addressable (~> 2.4)


ajbozarth:
@jerryshao this should address that issue that was brought to you. Once Travis passes, I'll merge this and open rc2. I'll also merge this to branch-0.4 but given it's an issue in the Docs I don't think it's worth pushing out a 0.4.1 release

jerryshao:
In the mail, it says "Known * critical severity* security vulnerability detected in nokogiri <
1.8.1 defined in Gemfile.lock", do we need to upgrade to 1.8.1?

ajbozarth:
1.8.1 is just the latest version and Gemfile.lock already imports 1.8.0 directly, the issue is that html-pipeline depends on nokogiri version 1.4 or greater. My solution just makes sure that dependency uses 1.5.4 or greater instead (the version the vulnerability was fixed in). Given it was a dependency of a dependency I wanted to cause the smallest change possible, but I can alway change both lines to use 1.8.1 instead.

jerryshao:
I see, thanks for explain.

PR #76, opened 2018-01-26
[MINOR][BUILD] Improve release-build script

ajbozarth:
Improved release-build.sh based on the 0.5.0-incubating release process.

Includes:

- Instead of forcing user to set env it prompts for required info, this also protects password inout
- Creates temp dir to hold files created in release process and makes sure git ignores it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ reports/
 metastore_db/
 derby.log
 dependency-reduced-pom.xml
+release-staging/
 
 # For python setup.py, which pollutes the source dirs.
 python-api/dist
--- a/dev/release-build.sh
+++ b/dev/release-build.sh
@@ -50,19 +50,36 @@ if [[ $@ == *"help"* ]]; then
   exit_with_usage
 fi
 
-for env in ASF_USERNAME ASF_PASSWORD GPG_PASSPHRASE RELEASE_RC; do
-  if [ -z "${!env}" ]; then
-    echo "ERROR: $env must be set to run this script"
-    exit_with_usage
-  fi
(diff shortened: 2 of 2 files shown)


ajbozarth:
@jerryshao I made some update to your release script while doing 0.5, could you take a look. Also if you could try it in your own env, I did extra testing by inserting a `exit 0` before the svn block of code. Note I hav't tested the `publish-release` portion, so this should be merged until after I test that in the final release.

jerryshao:
LGTM, thanks for the fix.

ajbozarth:
@jerryshao made a small updated to fix an issue I found while running `publish-release`

jerryshao:
I wouldn't expect release package to be built on mac environment, so it is not necessary to fix.

ajbozarth:
But I just had to run it on a mac, my dev environment is my MacBook Pro.

jerryshao:
You can install a virtualbox or something else. My concern about building on Mac is that some behavior maybe different regarding to native library. Since Mac will never be used in production. We'd better to keep the same environment for building and running.

ajbozarth:
(left this alone for a bit since it was low priority)

I disagree about releasing from a Mac, many Spark releases are done from a Mac (at least those done by people I know). Also IMO scripts should be written to be platform agnostic, so they should work in both a prod and dev env, not just prod or dev (and Mac a primary dev env in Silicon Valley currently).

ajbozarth:
@jerryshao ^

jerryshao:
I see. I'm OK with it. Sorry for late response.

PR #77, opened 2018-02-04
[LIVY-438] support upload resources which name contain non-ascii char.

fishjoy:
## support upload resources which name contain non-ascii char
https://issues.apache.org/jira/browse/LIVY-438

## How was this patch tested?
I write codes to test the patch.


jerryshao:
Maybe you also need to fix python client API, please double check it.

ajbozarth:
I'll defer to @jerryshao on this since he would be more familiar with how to make sure non-ascii works

chengqiangboy:
@jerryshao  You are right, there is something wrong with python api when upload resource which have non-ascii-char name. Python api use `requests` to post file to livy server. File name is encoded by utf-8 if contails non-ascii-char. See detail: https://github.com/requests/requests/issues/2117
So, I add a function to decode file name after receive the http body in `InteractiveSessionServlet`.

jerryshao:
Hi @chengqiangboy , thanks for the fix. I think it would be better to add a unit test to verify your changes.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSessionServlet.scala
```diff
@@ -247,4 +247,21 @@ class InteractiveSessionServlet(
...
+
+  /**
+    * This function used to decode fileName which is uploaded to spark through python api.
+    * see detail: https://github.com/requests/requests/issues/2117
+    * @param utf8EncodeName
+    * @return
+    */
+  private def resolveResourceName(utf8EncodeName: String): String = {
+    val utf8EncodeTag = "utf-8\'\'"
+    if (utf8EncodeName != null && utf8EncodeName.startsWith(utf8EncodeTag)) {
```
jerryshao: Do you mean that python uploaded non-ascii file name will start with "utf-8"? Is it a requests library specific feature, or it is a http standard?

chengqiangboy: According to the owner of `requests` said, there is a standard, and `requests` implement it. It is defined in https://tools.ietf.org/html/rfc5987#section-3.2.1

jerryshao: So this means that Java http client will also include such prefix?

chengqiangboy: This depends on http client implements, java http client has no prefix. Actually, `requests-1.2.3` has no prefix too. But I tried to change the `request`'s version of python-client, leads to other errors.

jerryshao: I see, thanks for explanation.

chengqiangboy: That is my pleasure. I will add unit test for this change.

chengqiangboy:
```Failed to execute goal org.scalastyle:scalastyle-maven-plugin:0.8.0:check (default) on project livy-client-http: Failed during scalastyle execution: You have 2 Scalastyle violation(s). -> [Help 1]```
There is a problem when add unit test, since there is a scalastyle check rule:
`<check level="error" class="org.scalastyle.scalariform.NonASCIICharacterChecker" enabled="true"></check>`
I think this check rule is necessary, so we believe we cannot remove this rule.
@jerryshao

**Comment by jerryshao** [comment]
Ohh, I see. Can you please check if there's some tags like `scalastyle:off xxx scalastyle:on xxx` available which can be used around your test case?

**Comment by chengqiangboy** [comment]
I add unit tests for java api and python api. The last CI test has errors in  393.6, 393.7. But I don't know why. Would you like to help me?

**Comment by chengqiangboy** [comment]
@jerryshao  I have added unit tests for java api and python api. Please review the patch.

File: python-api/src/test/python/livy-tests/client_test.py [hunk 2]
```diff
@@ -197,6 +208,16 @@ def test_upload_file():
     assert upload_file_future.done()
 
 
+@responses.activate
+def test_upload_non_ascii_file():
+    file_path = os.path.dirname(os.path.abspath(__file__)) + \
+        u"/resources/text_файл.txt"
```
jerryshao: I would suggest we create files in the runtime in test suite, rather than adding files to resource folder.

chengqiangboy:
@jerryshao  In the last commit, I added unit tests code in `integration-test` for python-api. So we can create non-ascii tmp file. But there has a problem in unit test. In this patch, I upload non-ascii files to spark throws python api, and get the file content through `pyspark code`. There is a bug in pyspark.

```
   source : spark-2.2.0-bin-hadoop2.7/python/pyspark/files.py

    @classmethod
    def get(cls, filename):
        """
        Get the absolute path of a file added through C{SparkContext.addFile()}.
        """
        path = os.path.join(SparkFiles.getRootDirectory(), filename)
        return os.path.abspath(path)
```
Then call `SparkFiles.get(nonASCIIfileName)`, the function in `spark-2.2.0-bin-hadoop2.7/python/pyspark/files.p` throws exception at `os.path.join` since the default codec is `ASCII`, but  filename is unicode. The exeception is:
```
path = os.path.join(SparkFiles.getRootDirectory(), filename)\n  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/posixpath.py", line 73, in join\n    path += \'/\' + b\nUnicodeDecodeError: \'ascii\' codec can\'t decode byte 0xd1 in position 42: ordinal not in range(128)\n\norg.apache.livy.repl.BypassPySparkJob.call(BypassPySparkJob.scala:32)
```
But, this patch is still works, we can upload non-ascii resources to spark correctly, but we cannot get the content through pyspark. So, I think the unit test for python-api should be simple: upload resources, but don't check it's content. What's your suggestion? Please let me know.

chengqiangboy:
@jerryshao  I had moved unit tests code in `integration-test` for python api. So we can create temp files in the runtime. Besides, I am sorry to make so many commits since the problems in my develop env.

PR #85, opened 2018-04-09
[LIVY-455][REPL] Fix json4s doesn't support java.math.BigDecimal issue

jerryshao:

Livy's SQLInterpreter will throw exception when rows contain java.math.BigDecimal data. This is because current version of json4s doesn't treat java.math.BigDecimal type as primitive type. On the contrary, json4s supports Scala BigDecimal as primitive type. So the fix is to convert java BigDecimal to Scala BigDecimal.

## How was this patch tested?

Unit test is added.

### DIFF
--- a/repl/src/main/scala/org/apache/livy/repl/SQLInterpreter.scala
+++ b/repl/src/main/scala/org/apache/livy/repl/SQLInterpreter.scala
@@ -97,7 +97,16 @@ class SQLInterpreter(
       val rows = result.getClass.getMethod("take", classOf[Int])
         .invoke(result, maxResult: java.lang.Integer)
         .asInstanceOf[Array[Row]]
-          .map(_.toSeq)
+        .map(_.toSeq)
+        .map { seq =>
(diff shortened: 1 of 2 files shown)


repl/src/main/scala/org/apache/livy/repl/SQLInterpreter.scala
```diff
@@ -97,7 +97,16 @@ class SQLInterpreter(
       val rows = result.getClass.getMethod("take", classOf[Int])
         .invoke(result, maxResult: java.lang.Integer)
         .asInstanceOf[Array[Row]]
-          .map(_.toSeq)
+        .map(_.toSeq)
+        .map { seq =>
+          seq.map {
```
vanzin: You could merge both maps by doing `row => row.toSeq.map {...` here.

jerryshao: Thanks for the comment, fixed.

PR #86, opened 2018-04-17
[LIVY-457][REPL] Fix SQLContext is not initialized correctly issue

jerryshao:

The signature of SQLContext's constructor is changed in Spark2, but we're still using the Spark1's signature, which will throw an exception when using this object.

## How was this patch tested?

UT and local verification.


jerryshao:
Test again.

jerryshao:
Merging to master and branch 0.5.

PR #88, opened 2018-04-20
[LIVY-456] IllegalStateException when trying to execute a statement while interactive session is still starting

IngoSchuster:
Fix timing issue when InteractiveSession is still starting and a request to execute a statement comes in. The change is that the check that ensures that the session is running does not immediately throw an exception in case that the state is "starting". If that's the case, we'll wait up to 30 seconds for the session to reach "running" state.
https://issues.apache.org/jira/browse/LIVY-456

## How was this patch tested?
Executed InteractiveSessionSpec to ensure that it still runs.
Other than that a manual test on a system where the Session startup takes about 15 seconds. Before the fix, I got an exception, with the fix, the exception is gone.

### DIFF
--- a/server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
+++ b/server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
@@ -499,7 +499,7 @@ class InteractiveSession(
   }
 
   def executeStatement(content: ExecuteRequest): Statement = {
-    ensureRunning()
+    ensureRunningWithRetryWhileStarting()
     recordActivity()
 
     val id = client.get.submitReplCode(content.code, content.kind.orNull).get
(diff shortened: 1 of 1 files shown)


server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
```diff
@@ -590,6 +590,25 @@ class InteractiveSession(
...
+  }
+
+  private def ensureRunningWithRetryWhileStarting(): Unit = {
+    val sleepMillis: Int = 500
+    val maxRetries: Int = 60
+    var numRetries: Int = 0
+    while (isStarting() && (numRetries < maxRetries)) {
+      info(s"Session still starting. Check status again in $sleepMillis msecs")
+      numRetries+=1
+      Thread.sleep(sleepMillis)
```
jerryshao: The loop here will block the REST query, seems this disobey the design principal of Livy REST API. Livy's API is designed to be non-block. If the session is not ready, user could wait and poll again, or using some async ways. Here this changes introduce unnecessary blocks, which looks not so intuitive.

CC @ajbozarth @zjffdu , what's your opinion?

IngoSchuster: Hm. Are you suggesting that when a client starts an interactive session, it has to poll by submitting code until the session is started? I would argue that this is not very intuitive either. And each unsuccessful check will generate an exception with stack trace in the logs.

I understand the principle of a non-blocking REST API. However, in case of starting a new session, what is the benefit of this non-blocking behaviour if the client cannot submit any other command until the session is started anyway? On the contrary - rather then Livy waiting until everything is ready, all clients will have to know about this and implement it.

So I wonder if it does really make sense to follow the non-blocking principle in this specific case?

zjffdu: I don't mean to submitting code to verify the session status. You just need to pull the session status via rest api.  See the rest api here. http://livy.incubator.apache.org/docs/latest/rest-api.html

jerryshao: I think maybe we can buffer the statement and execute them once session is ready, like what we do for JobAPI.

PR #90, opened 2018-05-02
[LIVY-466][RSC] Fix RSCDriver exception during RPC shutdown

jerryshao:

During RSCDriver's shutdown, it will first shutdown RPC server, and then all the RPC clients. When RPC client is closed, it will register a timeout to avoid orphaned RSCDriver, but this is not necessary during RSCDriver's shutdown, so here fixing this issue. The details can be seen in [JIRA](https://issues.apache.org/jira/browse/LIVY-466).

## How was this patch tested?

Local verification.


jerryshao:
Retest.

jerryshao:
Merging to master and branch-0.5.

PR #92, opened 2018-05-09
[LIVY-452] Differentiate FAILED and KILLED states

aromanenko-dev:

Currently, it's not possible to distinguish between two states - SparkApp.State.KILLED and SparkApp.State.FAILED. In both cases the session state will be SessionState.Dead(). This patch adds new SessionState.Killed() which will be used when job was actually killed by user.

https://issues.apache.org/jira/browse/LIVY-452

## How was this patch tested?

- Corresponding unit test was adjusted
- Tested manually on YARN cluster with `livy.spark.deploy-mode=cluster`


ajbozarth:
I don't have any problems with this, it LGTM. @jerryshao since you've been in this section of the code a bit more than me does it look good to you. If so, we can merge.

core/src/main/scala/org/apache/livy/sessions/SessionState.scala
```diff
@@ -55,8 +56,13 @@ object SessionState {
 
   object Busy extends SessionState("busy", true)
 
+  object Killed extends SessionState("killed", true)
```
jerryshao: Why do you need to add this `object Killed`, I don't think it is required as you already have `case class Killed`.

server/src/main/scala/org/apache/livy/utils/SparkProcApp.scala
```diff
@@ -53,8 +56,15 @@ class SparkProcApp (
     process.waitFor() match {
       case 0 => changeState(SparkApp.State.FINISHED)
       case exitCode =>
-        changeState(SparkApp.State.FAILED)
-        error(s"spark-submit exited with code $exitCode")
+        val message = if (killed) {
+          changeState(SparkApp.State.KILLED)
+          "job was killed by user"
+        }
+        else {
```
jerryshao: The style is not correct.

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
```diff
@@ -179,8 +179,8 @@ class BatchSession(
           info(s"Batch session $id created [appid: ${appId.orNull}, state: ${state.toString}, " +
             s"info: ${appInfo.asJavaMap}]")
         case SparkApp.State.FINISHED => _state = SessionState.Success()
-        case SparkApp.State.KILLED | SparkApp.State.FAILED =>
-          _state = SessionState.Dead()
+        case SparkApp.State.KILLED => _state = SessionState.Killed()
+        case SparkApp.State.FAILED => _state = SessionState.Dead()
```
jerryshao: I think you should also has a similar fix in `InteractiveSession`.

aromanenko-dev:
@jerryshao I addressed your comments and pushed updated commit.

jerryshao:
Please also add new session state to doc (http://livy.incubator.apache.org/docs/latest/rest-api.html#session).

aromanenko-dev:
@jerryshao Good point, it was added.

ajbozarth:
@aromanenko-dev for future PRs I would recommend not amending your commits but creating new ones for each update. By having a new commit for each update it allow us reviewers to see what you've changed in a diff.

aromanenko-dev:
@ajbozarth Ok, makes sense for me, I'll do. Thank you for letting me know!

PR #93, opened 2018-05-11
[LIVY-468] Reverse proxy support

m-wcislo:
Implements reverse proxy support, by adding basePath/URL concept, configurable using livy.conf

Resolves: [LIVY-468](https://issues.apache.org/jira/browse/LIVY-468)

## How was this patch tested?
Tested manually with Kong 0.11 and Spark 2.1.0 in cluster mode (using spark dispatcher).

### DIFF
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -29,6 +29,10 @@
 # What port to start the server on.
 # livy.server.port = 8998
 
+# What base path ui should work on. By default UI is mounted on "/".
+# E.g.: livy.ui.basePath = /my_livy - result in mounting UI on /my_livy/
+# livy.ui.basePath = ""
+
 # What spark master Livy sessions should use.
 # livy.spark.master = local
 
--- a/server/src/main/resources/org/apache/livy/server/ui/static/js/livy-ui.js
+++ b/server/src/main/resources/org/apache/livy/server/ui/static/js/livy-ui.js
@@ -27,6 +27,8 @@ var entityMap = {
   '=': '&#x3D;'
 };
 
+var basePath = "";
+
 function escapeHtml(string) {
   return String(string).replace(/[&<>"'`=\/]/g, function fromEntityMap (s) {
     return entityMap[s];
@@ -34,7 +36,7 @@ function escapeHtml(string) {
 }
 
 function uiLink(relativePath, inner) {
-  return anchorLink("/ui/" + relativePath, inner);
+  return anchorLink(prependBasePath("/ui/") + relativePath, inner);
 }
 
 function anchorLink(link, inner) {
@@ -89,8 +91,17 @@ function progressBar(double) {
 
 function getPathArray() {
   var pathArr = location.pathname.split("/");
-  pathArr.splice(0, 2);
+  var baseUrlEnd = 2 + (basePath.match(/\//g) || []).length;
+  pathArr.splice(0, baseUrlEnd);
   return pathArr;
+}
+
+function setBasePath(path) {
+  basePath = path;
+}
+
+function prependBasePath(path) {
+  return basePath + path;
 }
 
 $.extend( $.fn.dataTable.defaults, {
--- a/server/src/main/scala/org/apache/livy/LivyConf.scala
+++ b/server/src/main/scala/org/apache/livy/LivyConf.scala
@@ -61,6 +61,7 @@ object LivyConf {
 
   val SERVER_HOST = Entry("livy.server.host", "0.0.0.0")
   val SERVER_PORT = Entry("livy.server.port", 8998)
+  val SERVER_BASE_PATH = Entry("livy.ui.basePath", "")
 
   val UI_ENABLED = Entry("livy.ui.enabled", true)
 
--- a/server/src/main/scala/org/apache/livy/server/LivyServer.scala
+++ b/server/src/main/scala/org/apache/livy/server/LivyServer.scala
@@ -62,6 +62,7 @@ class LivyServer extends Logging {
 
     val host = livyConf.get(SERVER_HOST)
     val port = livyConf.getInt(SERVER_PORT)
+    val basePath = livyConf.get(SERVER_BASE_PATH)
     val multipartConfig = MultipartConfig(
         maxFileSize = Some(livyConf.getLong(LivyConf.FILE_UPLOAD_MAX_SIZE))
       ).toMultipartConfigElement
@@ -198,12 +199,12 @@ class LivyServer extends Logging {
             mount(context, batchServlet, "/batches/*")
 
             if (livyConf.getBoolean(UI_ENABLED)) {
-              val uiServlet = new UIServlet
+              val uiServlet = new UIServlet(basePath)
               mount(context, uiServlet, "/ui/*")
               mount(context, staticResourceServlet, "/static/*")
-              mount(context, uiRedirectServlet("/ui/"), "/*")
+              mount(context, uiRedirectServlet(basePath + "/ui/"), "/*")
             } else {
-              mount(context, uiRedirectServlet("/metrics"), "/*")
+              mount(context, uiRedirectServlet(basePath + "/metrics"), "/*")
             }
 
             context.mountMetricsAdminServlet("/metrics")
--- a/server/src/main/scala/org/apache/livy/server/ui/UIServlet.scala
+++ b/server/src/main/scala/org/apache/livy/server/ui/UIServlet.scala
@@ -21,7 +21,7 @@ import scala.xml.Node
 
 import org.scalatra.ScalatraServlet
 
-class UIServlet extends ScalatraServlet {
+class UIServlet(val basePath: String) extends ScalatraServlet {
   before() { contentType = "text/html" }
 
   sealed trait Page { val name: String }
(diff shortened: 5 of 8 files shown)


jerryshao:
CC @ajbozarth , would you please take a review? I'm not so familiar with UI part.

ajbozarth:
I'll take a look when I have time (hopefully today but definitely this week)

jerryshao:
@m-wcislo is this work done? If yes then remove the "WIP" in the title.

server/src/main/scala/org/apache/livy/server/LivyServer.scala
```diff
@@ -198,12 +199,12 @@ class LivyServer extends Logging {
             mount(context, batchServlet, "/batches/*")
 
             if (livyConf.getBoolean(UI_ENABLED)) {
-              val uiServlet = new UIServlet
+              val uiServlet = new UIServlet(basePath)
               mount(context, uiServlet, "/ui/*")
               mount(context, staticResourceServlet, "/static/*")
-              mount(context, uiRedirectServlet("/ui/"), "/*")
+              mount(context, uiRedirectServlet(basePath + "/ui/"), "/*")
```
ajbozarth: I'm not sure these redirect mounts should be changed. Did you find this change necessary in your testing?

server/src/main/scala/org/apache/livy/server/ui/UIServlet.scala
```diff
@@ -37,23 +37,30 @@ class UIServlet extends ScalatraServlet {
...
-      <script src="/static/js/livy-ui.js"></script>
+      <link rel="stylesheet"
+            href={basePath + "/static/css/bootstrap.min.css"}
+            type="text/css"/>
+      <link rel="stylesheet"
+            href={basePath + "/static/css/dataTables.bootstrap.min.css"}
+            type="text/css"/>
+      <link rel="stylesheet" href={basePath + "/static/css/livy-ui.css"} type="text/css"/>
+      <script type="text/javascript">
+        sessionStorage.setItem('basePath', {"'" + basePath + "'"});
```
ajbozarth: I don't think we need to use session storage here. We can just use a js var with a setter like in Spark (https://github.com/apache/spark/blob/master/core/src/main/resources/org/apache/spark/ui/static/webui.js)

server/src/main/scala/org/apache/livy/server/LivyServer.scala
m-wcislo: Yes, without it, entering my.domain/mylivy would redirect to my.domain/ui instead of my.domain/mylivy/ui.

server/src/main/scala/org/apache/livy/server/ui/UIServlet.scala
m-wcislo: Good point. Changed already.

ajbozarth:
Thanks for the fixes, I don't know when I'll have time to check out and run your code though. @jerryshao would you be able to check this out and try running it? The code LGTM but I want a second env to run it and don't want this blocking on me.

server/src/main/resources/org/apache/livy/server/ui/static/js/livy-ui.js
```diff
@@ -89,10 +91,19 @@ function progressBar(double) {
 
 function getPathArray() {
   var pathArr = location.pathname.split("/");
-  pathArr.splice(0, 2);
+  var extra = (basePath.match(/\//g) || []).length;
+  pathArr.splice(0, 2+extra);
```
jerryshao: nit: `2 + extra`?

jerryshao:
Should you also support the reverse proxy for REST APIs?

server/src/main/resources/org/apache/livy/server/ui/static/js/livy-ui.js
m-wcislo: Right, also I think below would be more descriptive?
```
var baseUrlEnd = 2 + (basePath.match(/\//g) || []).length;
pathArr.splice(0, baseUrlEnd);
```

m-wcislo:
Yes reverse proxy mainly causes problems in UI,  because html/js is executed on client side (browser). Until any ui paths or something like that are returned by REST API only UI needs changes.

ajbozarth:
@m-wcislo is correct. This was why I double checked about the redirects above, because I had assumed they worked like the API mounts.

server/src/main/resources/org/apache/livy/server/ui/static/js/livy-ui.js
jerryshao: Ok, why don't you fix it.

server/src/main/scala/org/apache/livy/LivyConf.scala
```diff
@@ -61,6 +61,7 @@ object LivyConf {
 
   val SERVER_HOST = Entry("livy.server.host", "0.0.0.0")
   val SERVER_PORT = Entry("livy.server.port", 8998)
+  val SERVER_BASE_PATH = Entry("livy.server.basePath", "")
```
jerryshao: If it is only related to UI, I would suggest to rename the configuration to "livy.ui.basePath".

Beside also adding this configuration and description to `livy.conf.template`.

m-wcislo: Done.

server/src/main/resources/org/apache/livy/server/ui/static/js/livy-ui.js
m-wcislo: Done

conf/livy.conf.template
```diff
@@ -29,6 +29,9 @@
 # What port to start the server on.
 # livy.server.port = 8998
 
+# What base path ui should work on. By default it is "/"
+# livy.ui.basePath =
```
ajbozarth: Based on the code below the default is "" not "/" and the default should be on this line rather that specified in a comment above.

server/src/main/resources/org/apache/livy/server/ui/static/js/livy-ui.js
```diff
@@ -89,10 +91,19 @@ function progressBar(double) {
...
 function getPathArray() {
   var pathArr = location.pathname.split("/");
-  pathArr.splice(0, 2);
+  var baseUrlEnd = 2 + (basePath.match(/\//g) || []).length;
+  pathArr.splice(0, baseUrlEnd);
   return pathArr;
 }
 
+function setBasePath(path) {
+	basePath = path;
```
ajbozarth: this is a tab rather than 2 spaces

server/src/main/resources/org/apache/livy/server/ui/static/js/livy-ui.js
```diff
@@ -89,10 +91,19 @@ function progressBar(double) {
...
+}
+
+function prependBasePath(path) {
+	return basePath + path;
```
ajbozarth: tab again

server/src/main/scala/org/apache/livy/server/ui/UIServlet.scala
```diff
@@ -68,12 +75,13 @@ class UIServlet extends ScalatraServlet {
...
       case sessionPage: SessionPage => {
-        <li><a href="/ui">Sessions</a></li> ++
+        <li><a href={basePath + "/ui"}>Sessions</a></li> ++
           <li class="active"><a href="#">{sessionPage.name}</a></li>
       }
       case logPage: LogPage => {
-        val sessionLink = if (logPage.sessionType == "Session") "/ui/session/" + logPage.id else "#"
-        <li><a href="/ui">Sessions</a></li> ++
+        val sessionLink = if (logPage.sessionType == "Session") basePath +
+                             "/ui/session/" + logPage.id else "#"
```
ajbozarth: Since this doesn't fit on one line this should be written in a format like:
```
if () {
} else {
}
```

conf/livy.conf.template
m-wcislo: The thing is that by default ui is mounted on "/". If we pass something to livy.ui.basePath like "/my_livy" ui will be mounted on "/my_livy/". Also I'd like to avoid any mandatory value here. I could better describe it like ".. By default ui is mounted on '/'" - not the default value is "/".

server/src/main/resources/org/apache/livy/server/ui/static/js/livy-ui.js
m-wcislo: Fixed.

server/src/main/resources/org/apache/livy/server/ui/static/js/livy-ui.js
m-wcislo: Fixed.

ajbozarth:
@m-wcislo for future PRs I would recommend not amending your commits but creating new ones for each update. By having a new commit for each update it allow us reviewers to see what you've changed in a diff.

conf/livy.conf.template
ajbozarth: What I meant by the default being `""` is that in `LivyConf` it's set to `""` so it should be
```
# livy.ui.basePath = ""
```
As for your explanation about your comment, I agree that it would be clearer to say it "mounts on `/` by default" so as to differentiate the value of the config from the explanation of the default mount point.

m-wcislo: OK. Fixed.

m-wcislo:
Right, I didn't want to make commit for every tiny change, but since there were couple of,  it ended up with little mess, sorry for that. Still amending until new discussion is opened.

ajbozarth:
It's ok to amend and squash commits locally, it's just good practice to not do a force pushes to GitHub.com

PR #94, opened 2018-05-15
[LIVY-450] Override config options by environment variables.

aromanenko-dev:

Add a possibility to override all config options by environment variables. 
For example:
`LIVY_SPARK_MASTER` automatically overrides `livy.spark.master` and `LIVY_SPARK_DEPLOY_MODE` overrides `livy.spark.deploy-mode`.

It can be useful in case of using the same Docker image of Livy in different scenarios that requires different configuration options.

https://issues.apache.org/jira/browse/LIVY-450

## How was this patch tested?

- Added corresponding unit test
- Tested manually on virtual Spark2/Livy cluster

### DIFF
--- a/client-common/src/main/java/org/apache/livy/client/common/ClientConf.java
+++ b/client-common/src/main/java/org/apache/livy/client/common/ClientConf.java
@@ -38,6 +38,9 @@ public abstract class ClientConf<T extends ClientConf>
 
   protected Logger LOG = LoggerFactory.getLogger(getClass());
 
+  /** Option to configure the list of options that can be overridden by environment variables. */
+  static final String LIVY_CONF_ENV_WHITELIST_PROPERTY = "livy.conf.env.whitelist";
+
   public static interface ConfEntry {
 
     /**
@@ -70,6 +73,9 @@ public abstract class ClientConf<T extends ClientConf>
 
   protected final ConcurrentMap<String, String> config;
 
+  /** Used to keep a set of whitelisted options that can be overridden by environment variables. */
+  private Set<String> whitelistEnvOptions;
+
   /** Used to retrieve configuration value from environment variable. */
   private EnvironmentService environmentService;
 
@@ -84,13 +90,34 @@ public abstract class ClientConf<T extends ClientConf>
     }
   }
 
+  /**
+   * Split whitelist config option by commas, trim spaces and create new set of whitelisted options.
+   */
+  private void initWhitelistEnvOptions() {
+    String whitelist = config.get(LIVY_CONF_ENV_WHITELIST_PROPERTY);
+    if (whitelist == null) {
+      whitelistEnvOptions = new HashSet<>();
+    } else {
+      whitelistEnvOptions = new HashSet<String>(Arrays.asList(whitelist.split("\\s*,\\s*")));
+    }
+  }
+
   public String get(String key) {
-    // Environment variable overrides the correspondent configuration option if it was set before.
+    String val;
+
+    // lazy initialization
+    if (whitelistEnvOptions == null) {
+      initWhitelistEnvOptions();
+    }
+    // Environment variable overrides the correspondent configuration option if it was set before
+    // and this option was found in whitelist.
     // For example, option "livy.spark.master" will be overridden by LIVY_SPARK_MASTER environment
     // variable.
-    String val = environmentService.getVariable(key.toUpperCase().replaceAll("[.-]", "_"));
-    if (val != null) {
-      return val;
+    if (!key.equals(LIVY_CONF_ENV_WHITELIST_PROPERTY) && whitelistEnvOptions.contains(key)) {
+      val = environmentService.getVariable(key.toUpperCase().replaceAll("[.-]", "_"));
+      if (val != null) {
+        return val;
+      }
     }
 
     val = config.get(key);
--- a/client-common/src/main/java/org/apache/livy/client/common/ClientConf.java
+++ b/client-common/src/main/java/org/apache/livy/client/common/ClientConf.java
@@ -70,7 +70,11 @@ public abstract class ClientConf<T extends ClientConf>
 
   protected final ConcurrentMap<String, String> config;
 
+  /** Used to retrieve configuration value from environment variable. */
+  private EnvironmentService environmentService;
+
   protected ClientConf(Properties config) {
+    this.environmentService = new ClientEnvironmentService();
     this.config = new ConcurrentHashMap<>();
     if (config != null) {
       for (String key : config.stringPropertyNames()) {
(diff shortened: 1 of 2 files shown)


aromanenko-dev:
@jerryshao Thank you for response, I fully understand your concerns. In fact, there was a discussion about that in [LIVY-450](https://issues.apache.org/jira/browse/LIVY-450), where me and @RyanSkraba explained our scenario and why we need this feature. 

Perhaps, it would make sense to add a possibility to configure a list of config options that can be overridden by environment variables. In this case. the scope can be reduced and controlled by Livy administrator. What do you think?

jerryshao:
Yes, I think we should have a white list to define which configuration can be overridden by env variables.

Besides we should define a order (who overrides who), in Spark configuration will override env variable, maybe here we should keep the consistency with Spark.

aromanenko-dev:
@jerryshao Agree on adding a whitelist, so I'll work on this.

Regarding the ordering. For our use case, it would make sense to have a default configuration value, set in config file, which can be overridden by corresponding environment variable if it was added into whitelist. In this case, it's still under control of system administrator, who does a setup of Livy and decides what should be user-specific configuration, and it's configurable by user, who runs a job. So, it means that environment variable will have a priority over configuration option. 
Are any issues related with this?

jerryshao:
>Regarding the ordering. For our use case, it would make sense to have a default configuration value, set in config file, which can be overridden by corresponding environment variable if it was added into whitelist. In this case, it's still under control of system administrator, who does a setup of Livy and decides what should be user-specific configuration, and it's configurable by user, who runs a job. So, it means that environment variable will have a priority over configuration option.

I'm OK with this.

aromanenko-dev:
@jerryshao Great! So I'm going to update my PR with adding this new functionality.

aromanenko-dev:
@jerryshao I added a whitelist option as it was discussed above. Could you take a look on it?

aromanenko-dev:
R: @jerryshao Could you take a look on my last changes?

jerryshao:
Sorry for the late response, I'm a little busy these days, will review this PR some day this week. Sorry about it.

aromanenko-dev:
@jerryshao Thank you for response! Please, take your time

client-common/src/main/java/org/apache/livy/client/common/ClientConf.java
```diff
@@ -80,8 +90,37 @@ protected ClientConf(Properties config) {
...
+    // lazy initialization
+    if (whitelistEnvOptions == null) {
+      initWhitelistEnvOptions();
+    }
+    // Environment variable overrides the correspondent configuration option if it was set before
+    // and this option was found in whitelist.
+    // For example, option "livy.spark.master" will be overridden by LIVY_SPARK_MASTER environment
+    // variable.
+    if (!key.equals(LIVY_CONF_ENV_WHITELIST_PROPERTY) && whitelistEnvOptions.contains(key)) {
+      val = environmentService.getVariable(key.toUpperCase().replaceAll("[.-]", "_"));
```
jerryshao: This seems not a good idea to mapping all the configurations to env variables with a simple "." to "-". I actually don't want all the configurations have the ability to use env variable to replace.

jerryshao:
I started to think if this is the only solution to bring flexibility to Livy docker image. I think it is a common problem for all the docker based applications, user requires to change parameters time to time even for docker based application. What is the common practice for such usage?

jerryshao:
For example, Spark has a k8s support which requires a docker image, this Spark docker image doesn't have any configurations preset. When user runs Spark on k8s, user could configure the different properties to let Spark to honor. This seems a like similar scenario, I'm wondering why do have to use env variables for docker image?

client-common/src/main/java/org/apache/livy/client/common/ClientConf.java
aromanenko-dev: Well, to avoid default mappings of all configuration options to env variables I added a whitelist option `livy.conf.env.whitelist` which is empty by default and can be configured only with an access to Livy config file. Using this option Livy administrator can limit the set of options that can be overridden. 
This is what we discussed before. Do you see any other objections with this?

RyanSkraba:
Hello -- there are a number of solutions for configuring a docker image!  I took a very quick look at some of the most top docker images:

* by environment variable (like mysql),
* by mounting a config file (like nginx and httpd -- which also require mounting content to be useful!),
* by building a new docker image containing your customized config file and/or content (almost all of them, but not the flexibility we're looking for),
* by providing the config on the command line when you run the image (redis, memcached have examples).

The list isn't exclusive -- the image author can pick and choose one or more.  I'm intrigued by nginx, which optionally provides an "envsubst" approach to its config.

Another solution would be to permit substitution for environment variables in the Livy config parser directly, something HOCON-ish: https://github.com/lightbend/config/blob/master/HOCON.md#goals--background  For the community, it might be a plus to use an existing config library instead of maintaining it's own!

To be honest, by environment variable is by far the simplest for most of our deployments!  We use a variety of techniques with different products in dev and prod: one image per machine, or docker-compose, or (in some but not all cases) k8s.  Our devops discourage us from mounting volumes, although IIRC this is simpler if everything is deployed via k8s.

I'm not entirely sure about the rationale for the whitelist -- the system administrator or devops that has control over the whitelist is the same person that has control over setting environment variables for the image!

RyanSkraba:
> Another solution would be to permit substitution for environment variables in the Livy config parser directly

Sorry -- on reflection, this is what @aromanenko-dev has done in this PR!

aromanenko-dev:
@jerryshao I'm sorry for delay with response.

> When user runs Spark on k8s, user could configure the different properties to let Spark to honor. 

Do you know if we can do the same with k8s/docker and Livy now? Probably, I was not aware about that, It would be great if we can.

> I'm wondering why do have to use env variables for docker image?

@RyanSkraba perfectly explained above our use case, different ways how we could go and, actually, it was simplest and only possible solution for that moment but probably we missed something.  

What are the disadvantages and obstacles of proposed solution with env variables, that prevents this PR to be merged, from your point of you?

jerryshao:
Sorry I'm not familiar with k8s. I just saw Spark could support it.

My concern, a) I don't want to the configurations to support env variables, just several picked configurations. b) I don't want the name to be simply defined as upper case and hyphen, a proper name should be defined.

aromanenko-dev:
Unfortunately, for our use case, the solution with a limited set of "hardcoded" env variables, that can be used for configuration, won't work well. In this case, it would be easier and more generic to have default livy.conf template in the docker image which contains configuration options in the way like `livy.spark.master=$LIVY_SPARK_MASTER` and then can be substituted by env vars values with a help of `envsubst` on docker startup.
This PR, I think, should be closed then.

PR #96, opened 2018-05-23
[LIVY-472][Server] Improve the logs for fail-to-create session

jerryshao:

Livy currently doesn't give a very clear log about the fail-to-create session, it only says that session related app tag cannot be found in RM, but doesn't tell user how to search and get the true root cause. So here change the logs to make it more clear.

## How was this patch tested?

Local verification.

### DIFF
--- a/server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
+++ b/server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
@@ -25,6 +25,7 @@ import scala.concurrent._
 import scala.concurrent.duration._
 import scala.language.postfixOps
 import scala.util.Try
+import scala.util.control.NonFatal
 
 import org.apache.hadoop.yarn.api.records.{ApplicationId, ApplicationReport, FinalApplicationStatus, YarnApplicationState}
 import org.apache.hadoop.yarn.client.api.YarnClient
@@ -33,6 +34,7 @@ import org.apache.hadoop.yarn.exceptions.ApplicationAttemptNotFoundException
 import org.apache.hadoop.yarn.util.ConverterUtils
 
 import org.apache.livy.{LivyConf, Logging, Utils}
+
 
 object SparkYarnApp extends Logging {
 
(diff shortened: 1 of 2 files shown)


jerryshao:
@zjffdu Would you please help to review? Thanks.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -179,9 +180,11 @@ class SparkYarnApp private[utils] (
...
           process.foreach(_.destroy())
           leakedAppTags.put(appTag, System.currentTimeMillis())
-          throw new Exception(s"No YARN application is found with tag $appTagLowerCase in " +
-            livyConf.getTimeAsMs(LivyConf.YARN_APP_LOOKUP_TIMEOUT)/1000 + " seconds. " +
-            "Please check your cluster status, it is may be very busy.")
+          throw new IllegalStateException(s"No YARN application is found with tag" +
+            s" $appTagLowerCase in ${livyConf.getTimeAsMs(LivyConf.YARN_APP_LOOKUP_TIMEOUT)/1000}" +
+            " seconds. This may be because 1) spark-submit fail to submit application to YARN; " +
+            "or 2) YARN cluster doesn't have enough resources to start the application in time. " +
+            "Please check Livy log and YARN log to know the details.")
```
zjffdu: Could we include the yarn appId or yarn app webl url in this log ?

jerryshao: No, we can't. Usually when we met such issue, yarn doesn't actually launch the application, so there's no web URL, or app id is invalid.

jerryshao:
x'x'x

jerryshao:
Any more comments? @zjffdu .

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
zjffdu: Then could we include the livy session web link so that user that check this livy session to see why session is failed ?

jerryshao: I can try to add (but I don't want to change a lot for just a link), also Livy session web link is no different from the returning json data.

jerryshao: Double check the code, seems this is not easy to get a session UI link here, this will involve unnecessary parameter passing between different classes.

zjffdu: OK, it's fine

zjffdu:
LGTM

jerryshao:
Thanks for the review @zjffdu !

Merging to master and branch 0.5.

PR #98, opened 2018-05-30
[LIVY-474] Livy SparkRInterpreter should support RBackend.init() return tuple

yanboliang:

Spark RBackend.init() return a tuple after https://github.com/GulajavaMinistudio/spark/pull/352 , we should update Livy SparkRInterpreter to adapt this change.


jerryshao:
Note. This issue only happens when using Spark 2.3.1+

PR #99, opened 2018-05-30
LIVY-475 Support of Hadoop CredentialProvider API

idzikovsky:
In this PR I've added following option to livy.conf:
```
# Hadoop Credential Provider Path to get "livy.keystore.password" and "livy.key-password"
# livy.hadoop.security.credential.provider.path =
```
to allow to specify path to Hadoop Credential Provider, which than used in `WebServer.scala` to set Keystore password and Key password to JKS that used to enable SSL encryption.

Also before trying to use Hadoop CredentialProvider API I'm checking if it available (as it was not available in Hadoop < 2.6) using the same [method that used in Oozie](https://github.com/apache/oozie/commit/6a731f9926158da38d1e3b518671ada95a544fe8#diff-800f95e605f21c5aaf5edef13039c9b9R124).

To use this, you will need to generate Credential Provider containing "livy.keystore.password" and/or "livy.key-password" in the common way:
```bash
hadoop credential create "livy.keystore.password" -value "keystore_secret" -provider jceks://hdfs@nn1.example.com/my/path/livy_creds.jceks
hadoop credential create "livy.key-password" -value "key_secret" -provider jceks://hdfs@nn1.example.com/my/path/livy_creds.jceks
```

### DIFF
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -23,7 +23,9 @@
 # Specify the key password.
 # livy.key-password =
 
-# Hadoop Credential Provider Path to get "livy.keystore.password" and "livy.key-password"
+# Hadoop Credential Provider Path to get "livy.keystore.password" and "livy.key-password".
+# Credential Provider can be created using command as follow:
+# hadoop credential create "livy.keystore.password" -value "secret" -provider jceks://hdfs/path/to/livy.jceks
 # livy.hadoop.security.credential.provider.path =
 
 # What host address to start the server on. By default, Livy will bind to all network interfaces.
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -23,6 +23,9 @@
 # Specify the key password.
 # livy.key-password =
 
+# Hadoop Credential Provider Path to get "livy.keystore.password" and "livy.key-password"
+# livy.hadoop.security.credential.provider.path =
+
 # What host address to start the server on. By default, Livy will bind to all network interfaces.
 # livy.server.host = 0.0.0.0
 
--- a/server/src/main/scala/org/apache/livy/server/WebServer.scala
+++ b/server/src/main/scala/org/apache/livy/server/WebServer.scala
@@ -58,12 +58,12 @@ class WebServer(livyConf: LivyConf, var host: String, var port: Int) extends Log
 
       val keyStorePassword = Option(livyConf.get(LivyConf.SSL_KEYSTORE_PASSWORD))
         .orElse {
-          Option(hadoopConf.getPassword(LivyConf.SSL_KEYSTORE_PASSWORD.key).mkString)
+          Option(hadoopConf.getPassword(LivyConf.SSL_KEYSTORE_PASSWORD.key)).map(_.mkString)
         }
 
       val keyPassword = Option(livyConf.get(LivyConf.SSL_KEY_PASSWORD))
         .orElse {
-          Option(hadoopConf.getPassword(LivyConf.SSL_KEY_PASSWORD.key).mkString)
+          Option(hadoopConf.getPassword(LivyConf.SSL_KEY_PASSWORD.key)).map(_.mkString)
         }
 
       keyStorePassword
--- a/server/src/main/scala/org/apache/livy/server/WebServer.scala
+++ b/server/src/main/scala/org/apache/livy/server/WebServer.scala
@@ -20,6 +20,7 @@ package org.apache.livy.server
 import java.net.InetAddress
 import javax.servlet.ServletContextListener
 
+import org.apache.hadoop.conf.Configuration
 import org.eclipse.jetty.server._
 import org.eclipse.jetty.server.handler.{HandlerCollection, RequestLogHandler}
 import org.eclipse.jetty.servlet.{DefaultServlet, ServletContextHandler}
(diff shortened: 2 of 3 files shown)


jerryshao:
Please explain what/why you did in PR description.

idzikovsky:
@jerryshao sorry for that. I've added description.

jerryshao:
We don't support Hadoop 2.6-, so I don't think you need to consider that.

idzikovskyi:
@jerryshao removed that check.

server/src/main/scala/org/apache/livy/server/WebServer.scala
```diff
@@ -48,9 +49,31 @@ class WebServer(livyConf: LivyConf, var host: String, var port: Int) extends Log
...
+
+      if (credentialProviderPath != null) {
+        val hadoopConf = new Configuration()
+        hadoopConf.set("hadoop.security.credential.provider.path", credentialProviderPath)
+
+        val credentialKeyStorePassword = hadoopConf.getPassword(LivyConf.SSL_KEYSTORE_PASSWORD.key)
+        val credentialKeyPassword = hadoopConf.getPassword(LivyConf.SSL_KEY_PASSWORD.key)
+
+        if (credentialKeyStorePassword != null) {
+          keyStorePassword = credentialKeyStorePassword.mkString
```
jerryshao: Why do we need to `mkString` here.

server/src/main/scala/org/apache/livy/server/WebServer.scala
```diff
@@ -48,9 +49,31 @@ class WebServer(livyConf: LivyConf, var host: String, var port: Int) extends Log
 
       val sslContextFactory = new SslContextFactory()
       sslContextFactory.setKeyStorePath(keystore)
-      Option(livyConf.get(LivyConf.SSL_KEYSTORE_PASSWORD))
+
+      var keyStorePassword = livyConf.get(LivyConf.SSL_KEYSTORE_PASSWORD)
```
jerryshao: Can we change the code more scala-ish here and below, for example:

```
val keyStorePassword = Option(livyConf.get(xxxx))
  .orElse {
      Option(hadoopConf.getPassword(xxx))
  }
```

conf/livy.conf.template
```diff
@@ -23,6 +23,9 @@
 # Specify the key password.
 # livy.key-password =
 
+# Hadoop Credential Provider Path to get "livy.keystore.password" and "livy.key-password"
```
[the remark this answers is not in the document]
idzikovsky: Added more description in last comment.

server/src/main/scala/org/apache/livy/server/WebServer.scala
idzikovsky: Because [getPassword returns](https://hadoop.apache.org/docs/r2.8.3/api/org/apache/hadoop/conf/Configuration.html#getPassword(java.lang.String)) Array[Char].

server/src/main/scala/org/apache/livy/server/WebServer.scala
idzikovsky: Changed this in last commit.

idzikovsky:
I'm checking now Livy build from latest PR and share results if everything is fine.

server/src/main/scala/org/apache/livy/server/WebServer.scala
```diff
@@ -48,9 +49,26 @@ class WebServer(livyConf: LivyConf, var host: String, var port: Int) extends Log
...
+
+      val credentialProviderPath = livyConf.get(LivyConf.HADOOP_CREDENTIAL_PROVIDER_PATH)
+      val hadoopConf = new Configuration()
+      if (credentialProviderPath != null) {
+        hadoopConf.set("hadoop.security.credential.provider.path", credentialProviderPath)
+      }
+
+      val keyStorePassword = Option(livyConf.get(LivyConf.SSL_KEYSTORE_PASSWORD))
+        .orElse {
+          Option(hadoopConf.getPassword(LivyConf.SSL_KEYSTORE_PASSWORD.key).mkString)
```
jerryshao: `hadoopConf.getPassword(xxx).mkString` will throw NPE if password is null, I think you should change to:

```
Option(xxx).map(_.mkString)
```

server/src/main/scala/org/apache/livy/server/WebServer.scala
```diff
@@ -48,9 +49,26 @@ class WebServer(livyConf: LivyConf, var host: String, var port: Int) extends Log
...
+      val keyPassword = Option(livyConf.get(LivyConf.SSL_KEY_PASSWORD))
+        .orElse {
+          Option(hadoopConf.getPassword(LivyConf.SSL_KEY_PASSWORD.key)).map(_.mkString)
+        }
+
+      keyStorePassword
         .foreach(sslContextFactory.setKeyStorePassword)
-      Option(livyConf.get(LivyConf.SSL_KEY_PASSWORD))
+      keyPassword
         .foreach(sslContextFactory.setKeyManagerPassword)
```
jerryshao: not necessary to change to another line for now.

server/src/main/scala/org/apache/livy/server/WebServer.scala
```diff
@@ -48,9 +49,26 @@ class WebServer(livyConf: LivyConf, var host: String, var port: Int) extends Log
...
+          Option(hadoopConf.getPassword(LivyConf.SSL_KEYSTORE_PASSWORD.key)).map(_.mkString)
+        }
+
+      val keyPassword = Option(livyConf.get(LivyConf.SSL_KEY_PASSWORD))
+        .orElse {
+          Option(hadoopConf.getPassword(LivyConf.SSL_KEY_PASSWORD.key)).map(_.mkString)
+        }
+
+      keyStorePassword
         .foreach(sslContextFactory.setKeyStorePassword)
```
jerryshao: Also here.

jerryshao:
Can you please verify it locally again to see if the newest change works, thanks!

server/src/main/scala/org/apache/livy/server/WebServer.scala
idzikovsky: Yeah, I've found that when tried to test build from this branch, but have troubles on fixing this in scala-way:)
Thank you for helping with this.

server/src/main/scala/org/apache/livy/server/WebServer.scala
idzikovsky: fixed

server/src/main/scala/org/apache/livy/server/WebServer.scala
idzikovsky: fixed

idzikovsky:
I've just rebuild jar for livy-server, and tried with livy.keystore.password set in livy.conf, set in credentials provider in HDFS, and set to wrong password in livy.conf (Livy was not able to start in last case which is expected), and in all cases livy.key-password was empty and I didn't get any npe. Everything seems fine for me now.

jerryshao:
LGTM, merging to master branch.

idzikovsky:
@jerryshao thank you for your help, and sorry for being newbie in Scala.

PR #101, opened 2018-06-21
[LIVY-479] Enable livy.rsc.launcher.address configuration

taoliseki:
In current code we are setting livy.rsc.launcher.address to RPC server and ignoring whatever is specified in configs. However in some scenarios there is a need for the user to be able to explicitly configure it. For example, the IP address for an active Livy server might change over the time, so rather than using the fixed IP address, we need to specify this setting to a more generic host name. For this reason, we want to enable the capability of user side configurations.

Jira: https://issues.apache.org/jira/browse/LIVY-479

## How was this patch tested?
Manual tests by changing "livy.rsc.launcher.address" and verified the setting took effect

### DIFF
--- a/rsc/src/main/java/org/apache/livy/rsc/ContextLauncher.java
+++ b/rsc/src/main/java/org/apache/livy/rsc/ContextLauncher.java
@@ -97,7 +97,13 @@ class ContextLauncher {
       String replMode = conf.get("repl");
       boolean repl = replMode != null && replMode.equals("true");
 
-      conf.set(LAUNCHER_ADDRESS, factory.getServer().getAddress());
+      // In some scenarios the user may need to configure this endpoint setting explicitly.
+      String address = conf.get(LAUNCHER_ADDRESS);
+      // If not specified, use the RPC server address; otherwise use the specified address.
+      if (address == null) {
+        address = factory.getServer().getAddress();
+      }
+      conf.set(LAUNCHER_ADDRESS, address);
       conf.set(LAUNCHER_PORT, factory.getServer().getPort());
       conf.set(CLIENT_ID, clientId);
       conf.set(CLIENT_SECRET, secret);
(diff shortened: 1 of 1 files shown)


rsc/src/main/java/org/apache/livy/rsc/ContextLauncher.java
```diff
@@ -97,7 +97,13 @@ private ContextLauncher(RSCClientFactory factory, RSCConf conf) throws IOExcepti
       String replMode = conf.get("repl");
       boolean repl = replMode != null && replMode.equals("true");
 
-      conf.set(LAUNCHER_ADDRESS, factory.getServer().getAddress());
+      // In some scenarios the user may need to configure this endpoint setting explicitly.
+      String address = conf.get(LAUNCHER_ADDRESS);
```
jerryshao: How do you set this `LAUNCHER_ADDRESS` by user?

jerryshao:
> the IP address for an active Livy server might change over the time

In your case, shall we also fix the "rpc.server.address" to use hostname? This "rpc.server.address" might also has similar issue.

rsc/src/main/java/org/apache/livy/rsc/ContextLauncher.java
taoliseki: You can specify "livy.rsc.launcher.address" in livy-client.conf file.

taoliseki:
The existing logic is setting "rpc.server.address" to the host name where the RSC driver process is running. In a yarn cluster (for example), this driver process can be running in any node. So it's pointless for a user to configure that setting and the existing logic is good enough.

vanzin:
Looks fine. The difference between RPC_SERVER_ADDRESS and LAUNCHER_ADDRESS is not 100% clear to me (it's been a while) but this doesn't seem to make it worse.

Merging to master.

PR #103, opened 2018-07-17
[LIVY-239] Generate session IDs in SessionStore, not in SessionManager

meisam:

Move the logic to create session IDs from SessionManager to SessionStore

## Why this change is needed?
Generating session IDs at session stores that they are unique across all Livy instances that connect to a session store. Right now only one Livy instance connects to the session store, but to implement multimode HA, generating session IDs in session store prevents inconsistent/colliding values.

## How was this patch tested?

- Unit tests are added.
- This patch is working in production at PayPal for about a year now.

JIRA ticket: https://issues.apache.org/jira/browse/LIVY-239


meisam:
```
sudo -E apt-get ...
```
failed on one of the travis instances. Closing this PR and opening it again to rebuild.

server/src/main/scala/org/apache/livy/server/recovery/StateStore.scala
```diff
@@ -70,6 +70,14 @@ abstract class StateStore(livyConf: LivyConf) extends JsonMapper {
...
    */
   def remove(key: String): Unit
+
+  /**
+    * Gets the Long value for the given key, increments the value
+    * and stores the new value before returning the value.
+    *
+    * @return incremented value
+    */
+  def increment(key: String): Long
```
vanzin: This method name is weird, since you're not incrementing the `StateStore` itself... maybe `nextValue` instead.

vanzin (commented):
Looks reasonable from what I remember of this code.

server/src/main/scala/org/apache/livy/server/recovery/SessionStore.scala
```diff
@@ -67,15 +61,14 @@ class SessionStore(
...
+   * Return 0 if no value is stored in the state store.
+   * Save the next unused session ID to the session store before returning the current value.
    *
-   * @throws Exception If SessionManagerState stored is corrupted, it throws an error.
+   * @throws Exception If session store is corrupted or unreachable, it throws an error.
    */
   def getNextSessionId(sessionType: String): Int = {
-    store.get[SessionManagerState](sessionManagerPath(sessionType))
-      .map(_.nextSessionId).getOrElse(0)
+    store.increment(sessionManagerPath(sessionType)).toInt
```
alex-the-man: This's backward incompatible on an existing cluster as the `sessionManagerPath(sessionType)` is a JSON object.
Do you want to use a different path instead? May be suffix it with something?

meisam: @alex-the-man I changed the path.

server/src/main/scala/org/apache/livy/server/recovery/StateStore.scala
meisam: > Can we add some migration code so user can seamlessly upgrade from older Livy? Thanks!

Should the migration code be part of Livy? Or should be a separate script that users run once?

Also should we change the version in zookeeper path from `/v1/...` to `/v2/...`?

meisam:
> Can we add some migration code so user can seamlessly upgrade from older Livy? Thanks!

Should the migration code be part of Livy? Or should be a separate script that users run once?

Also should we change the version in zookeeper path from `/v1/...` to `/v2/...`?

alex-the-man:
> > Can we add some migration code so user can seamlessly upgrade from older Livy? Thanks!
> 
> Should the migration code be part of Livy? Or should be a separate script that users run once?
In this case it seems simple enough to be a part of the startup code.

> 
> Also should we change the version in zookeeper path from `/v1/...` to `/v2/...`?
I think you are right, we should.

server/src/main/scala/org/apache/livy/server/recovery/BlackholeStateStore.scala
```diff
@@ -26,11 +28,16 @@ import org.apache.livy.LivyConf
  * Livy will use this when session recovery is disabled.
  */
 class BlackholeStateStore(livyConf: LivyConf) extends StateStore(livyConf) {
+
+  private val atomicLong: AtomicLong = new AtomicLong(-1L)
```
yiheng: Looks like this cannot guarantee the global uniqueness across many Livy services, right? As the `atomicLong` cannot be shared by processes.

server/src/main/scala/org/apache/livy/server/recovery/FileSystemStateStore.scala
```diff
@@ -129,4 +129,10 @@ class FileSystemStateStore(
   }
 
   private def absPath(key: String): Path = new Path(fsUri.getPath(), key)
+
+  override def nextValue(key: String): Long = synchronized {
+    val incrementedValue = 1 + get[Long](key).getOrElse(-1L)
```
yiheng: Is the `get` or `set` method atomic?

PR #104, opened 2018-08-15
[LIVY-491][LIVY-492][LIVY-493] Add Thriftserver module and implementation

mgaido91:

The PR contains an implementation of a JDBC API for Livy server based on the Hive Thriftserver. The implementation is based on the version 3.0 of Hive Thriftserver.

This initial PR contains the thriftserver module added to Livy and its implementation. It doesn't contain any binding for starting it, this will be added later.

Some Hive classes have been ported here because they needed come modifications in order to works properly in Livy. Long term solution is to get of them all by re-implementing the needed parts in the Livy thriftserver itself, without relying on Hive code (other than the PRC interface) anymore. Those classes/changes can be summarized in three categories:

 1. Changes to make the Hive classes easy to extend: for instance, some visibility modifiers were changes (moving from `private` or `package private` to `protected` or `public`);
 2. Changes in order to reduce the dependencies on Hive modules/classes: for instance all the classes in the `operation` package were modified in order to get rid of the usage of `HiveSession` and the `HiveServer2` was changed in order not to use `CLIService`.
 3. The UGI management which is currently performed in Livy is definitely very different from the Hive one, this required changes to the `HiveAuthFactory` in order not to interfere with the existing codebase.

## How was this patch tested?

added integration tests as UTs


mgaido91:
cc @jerryshao

mgaido91:
travis failure now is intermittent, but I am not sure how to retrigger a build. Anyway it is a timeout in a job which passed in the commit before, and the last commit modifies only the timeout for the tests in the thriftserver module, so it completely unrelated to the timeout which occurs in the last failed job.

pom.xml
```diff
@@ -1077,6 +1078,25 @@
   </distributionManagement>
 
   <profiles>
+    <profile>
+      <id>thriftserver</id>
+      <activation>
+        <jdk>1.8</jdk>
+        <property>
+          <name>thriftserverEnabled</name>
```
vanzin: I'm not sure this is the best way to do this.

If I understand what's going on, you're saying that the profile should be enabled if that property is set and the java version is 8. Meaning that if you say `-DthriftserverEnabled` but you're using java 7, it will do nothing silently.

I think it would be more correct to enable the profile the usual way (`-Pthriftserver`) and use `maven-enforcer-plugin` to ensure that you're using java 8 if this profile is enabled.

pom.xml
```diff
@@ -1077,6 +1078,25 @@
...
+        </property>
+      </activation>
+      <modules>
+        <module>thriftserver</module>
+      </modules>
+      <properties>
+        <jetty.version>9.3.8.v20160314</jetty.version>
```
vanzin: Hmm. Why not, separately, update the jetty version for the whole of Livy instead of doing this? Sounds like a better idea to me.

thriftserver/pom.xml
```diff
@@ -0,0 +1,303 @@
...
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
```
vanzin: All other pom files seem to use 2-space indentation.

thriftserver/pom.xml
```diff
@@ -0,0 +1,303 @@
...
+        <artifactId>livy-main</artifactId>
+        <groupId>org.apache.livy</groupId>
+        <version>0.6.0-incubating-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>livy-thriftserver</artifactId>
+
+    <properties>
+        <hive.group>org.apache.hive</hive.group>
```
vanzin: Do we need this? I hope no one is planning to go down Spark's path of forking things...

thriftserver/pom.xml
```diff
@@ -0,0 +1,303 @@
...
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <outputDirectory>${project.build.directory}/jars</outputDirectory>
+                </configuration>
+            </plugin>
+
+            <plugin>
+                <groupId>com.googlecode.maven-download-plugin</groupId>
```
vanzin: Doesn't this mean that a single Livy build will download the same Spark archive twice? Sounds a little wasteful...

thriftserver/pom.xml
```diff
@@ -0,0 +1,303 @@
...
+
+    <properties>
+        <hive.group>org.apache.hive</hive.group>
+        <hive.version>3.0.0</hive.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>${hive.group}</groupId>
+            <artifactId>hive-cli</artifactId>
```
vanzin: Could you add a comment about why these Hive dependencies are needed?

I'd expect that aside from the Thrift stuff that is being forked, everything else would be just re-implemented in Livy, and there wouldn't be a need for any Hive stuff. But you've looked at this way more than I have, so I'm probably missing things still.

The main worry I have is that once you do this, you're bringing a whole lot of code into the Livy server's classpath; you already have a bunch of dependency exclusions, and who knows what other issues might arise. So if possible it would be better to avoid bringing these dependencies into the Livy server.

pom.xml
mgaido91: Thanks, I like the proposal, I'll give a try to it.

pom.xml
mgaido91: The problem is that since 9.3 Jetty supports only Java 8. So I cannot update it otherwise Java 7 builds would fail. Having it here enforces that we have Java 8 and we have the right Jetty version for the thriftserver to work. I don't see other viable options until Livy supports Java 7

thriftserver/pom.xml
mgaido91: sorry, I'll fix this.

thriftserver/pom.xml
mgaido91: it is not needed indeed. I can remove it.

thriftserver/pom.xml
mgaido91: The plugin uses cache, so actually the second time the download is attempted, the file is retrieved from the cache. Thus I don't think this is an issue. Do you agree?

thriftserver/pom.xml
mgaido91: This is actually a bit complicated. I tried to get rid of most Hive dependencies I could without breaking things and this seemed to be the minimal set. Actually it has been more a trial and error work, so there may be parts that I am missing, but unfortunately the Hive thriftserver is quite integrated with all the other Hive modules. Probably (but I am not sure, just a guess), deleting some part of the forked code which we are not really using we may be able to get rid of some of those, but I am not sure.

Trying to answer also to your main question, forking the code was needed mainly for 2 reasons (but they may become 3):

 - avoid to rewrite some Hive methods using reflection as it was done in STS;
 - avoiding to access the Hive metastore, which is done automatically and cannot be avoided without some code changes;
 - the third point I am not still sure about, since I still need to look more into it as I am not an expert on it, and I have to understand better all the implications, is the UGI management. The management which is performed in Hive may break also the Livy server itself.

I agree that the best solution is to use Hive's artifact avoiding this code forking and involving the Hive community in order to achieve this. But if point 1 is kind of trivial to achieve (it just means moving some methods to protected/public, which I think may be not a big deal), point 2 and - if needed - 3 are much more critical. So my proposal which you can find [here](https://docs.google.com/document/d/18HAR_VnQLegbYyzGg8f4zwD4GtDP5q_t3K21eXecZC4/edit) is to set this as a goal for the future, so this can start and in parallel we can work with the Hive community in order to make the Hive thriftserver easier to extend.

mgaido91:
thank you very much for looking at this @vanzin , really appreciated.

pom.xml
vanzin: Ah, true. We should probably at some point start discussing the removal of Java 7 / old Spark support (there's even talk of making Spark 2.1 EOL now), but that's a separate discussion.

thriftserver/pom.xml
vanzin: > but unfortunately the Hive thriftserver is quite integrated with all the other Hive modules

That's the thing I'd like to get a better grasp of but haven't fully had the time to do yet. I've only briefly looked at the Hive code around this, and my initial understanding is that it might be simpler to re-implement the `TCLIService` thrift service without borrowing that code from Hive.

There's some code that we could borrow there (e.g. auth stuff and maybe some socket / config initialization stuff from `ThriftCLIService` and `ThriftBinaryCLIService`), but I think in the long run getting rid of Hive baggage would be a better approach.

Also, you're forking both the `service-rpc` and parts of the `service` modules in Hive. Can't you just pull the `service-rpc` module as is? That's just the compiled thrift interfaces, from what I can see here:

https://repo1.maven.org/maven2/org/apache/hive/hive-service-rpc/3.0.0/

It has some extra dependencies (such as some tomcat stuff) that at first glance don't seem necessary, so maybe they can be excluded. But it would avoid forking the source code.

mgaido91: yes, thanks, I'll check the feasibility of using `service-rpc` in order to get rid of some of these classes. Unfortunately I think it will take some time, so I am not sure when I'll be able to provide a definitive answer on this topic, but I'll do my best.

The point is that actually we are not reusing only the thrift/RPC layer (with authentication and so on) but also most of the logic for session and operation management. This is probably the trickiest part also in term of dependencies needed.  It is true that we are overriding the core part of it, but most of the stuff is still handled by the Hive classes (namely `OperationManager`, `SessionManager` and so on).

I definitely do agree with you about getting rid of the Hive classes on the long run. I really appreciate any suggestion like yours and I'll try it as soon as I can. Reimplementing from the RPC interfaces is an option, too but it requires a lot of effort, as many things which are handled by the Hive code now, need to be reimplemented. I am not saying it is not doable, but it may be a huge effort which I'd suggest to do in a progressive way if possible.

jerryshao:
@mgaido91 close and reopen this PR could trigger the travis to run again.

jerryshao:
@mgaido91 what's the key purpose to copy thrift related codes to Livy? In Spark2 because we don't want to rely on Hive thrift library, also don't want to be to hacky to change the Hive related codes, so we copied the code to Spark. But here in Livy do we have a such requirement?

server/src/main/scala/org/apache/livy/LivyConf.scala
```diff
@@ -99,6 +99,10 @@ object LivyConf {
   val LAUNCH_KERBEROS_REFRESH_INTERVAL = Entry("livy.server.launch.kerberos.refresh-interval", "1h")
   val KINIT_FAIL_THRESHOLD = Entry("livy.server.launch.kerberos.kinit-fail-threshold", 5)
 
+  val THRIFT_INCR_COLLECT_ENABLED = Entry("livy.server.thrift.incrementalCollect", false)
+  val THRIFT_SESSION_CREATION_TIMEOUT = Entry("livy.server.thrift.session.creationTimeout", "10m")
+  val THRIFT_SERVER_JAR_LOCATION = Entry("livy.server.thrift.jarLocation", null)
```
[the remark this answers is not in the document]
mgaido91: this is used in order to add the JAR containing Livy thriftserver classes to the Spark sessions which are created or used by the thriftserver. This is needed because the RPC operations need some of the classes present in the JAR. Unfortunately, the local JAR cannot be used when we are using yarn-cluster mode, so the only working way which I cloud think of was to add the JAR to HDFS and add to the Spark session from there. So this would be the location on HDFS of the `livy-thriftserver` JAR. The other option I think would be to put the JAR in the `repl-jars` folder, but this would mean that all the sessions (also the ones which are not used by the thriftserver) would contain the thriftserver JAR, which I preferred to avoid.

thriftserver/pom.xml
```diff
@@ -0,0 +1,294 @@
...
+    </dependency>
+
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.thrift</groupId>
+      <artifactId>libthrift</artifactId>
+      <version>0.9.3</version>
```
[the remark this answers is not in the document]
mgaido91: sure, will do

thriftserver/src/main/java/org/apache/hive/service/AbstractService.java
```diff
@@ -0,0 +1,193 @@
...
+
+import org.apache.hadoop.hive.conf.HiveConf;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * AbstractService.
+ *
+ */
+public abstract class AbstractService implements Service {
```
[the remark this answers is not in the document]
mgaido91: yes

mgaido91:
@jerryshao thanks for you review.

The main reason for copying the classes from Hive are:

 1. We need to override private methods, so either we do it using reflection and kind of rewriting some methods using reflection or copying the Hive code and doing some very small changes about visibility and very small changes in order to make the code easier to extend (this is the least important reason);
 2. Avoid metastore access from the server, which is otherwise intrinsic in Hive codebase;
 3. This needs further investigation, but the UGI management which is performed automatically by Hive may need to be removed/adjusted in order not to interfere with the one performed by Livy.

I think that the 1 issue can be easily addressed by cooperating with the Hive community, but the other 2 points may be more critical to address. I'd consider the current situation as a workaround which will be replaced by using only Hive's artifacts once we can  solve the 3 points above with the help of the Hive community.

thriftserver/pom.xml
vanzin: Actually, since you mentioned `OperationManager`... your spec (IIRC) doesn't have a lot of details about how this stuff actually works. It mentions that RPCs will be sent to the session, but that's about it.

How do you see things like `OperationManager` even making sense here?

The way I'm seeing things, the thrift server side would take care of starting sessions, and once that's done, everything should be just delegated to them. That makes a class like `OperationManager` something that makes more sense on the session side, not on the server side.

To give a more concrete example: a call to `TCLIServer.ExecuteStatement` would just map the handle in the request to a session, and dispatch an RPC to that session, then return its result.

What other state do you plan to keep on the server side that makes things like Hive's `OperationManager` and `SessionManager` useful?

mgaido91: Well, they handle the management of a JDBC session creation, generating a new Handle for it. We can definitely rewrite them from scratch and we can get rid of most of the things they do.

There isn't much state which is really needed, it is more about the logic for the session lifecycle which is reused. So they are handling all the mapping between an incoming request to doing the proper operation on the proper session.

Overriding the `CLIService`, `ThriftCLIService` is indeed an option. It just takes much more time.

vanzin: The part I don't understand is how those classes help Livy. They're Hive classes, Livy sessions are a different thing. The external protocol is defined by the Thrift interfaces, not those classes.

At some point you will have to send an RPC to the actual Livy session to actually parse and execute statements. And that's where I don't see where the Hive classes help on the server side. You'll need state on the session side, not the server side, which only needs to know to which session to dispatch incoming requests. That's basically a hash table lookup.

For example, `TExecuteStatementReq` has a `sessionHandle` member. Livy just needs to map that session handle to a Livy session and send an RPC to the actual session to process that request. What else is there for the server side to do?

Unless I didn't really understand what you're trying to do here.

vanzin: BTW, one thing that can help, especially since this is a large-ish project with what I expect will be a few upcoming PRs, is to have a link to some repo / branches where you are keeping a version of what's coming later.

I did something like that with the history server revamp in Spark (which had more than 10 separate PRs), and it really helped with showing people what some of the early changes were for, since it wasn't necessarily clear from looking at just the change in isolation.

mgaido91: Ok, so let me start from the second point (which is easy :) ): the branch where I had developed this first is: https://github.com/mgaido91/incubator-livy/commits/livy_thrift. Please while looking at it disregard the last commit as I am doing some tests for proper keytab management, but I am still working on it. Actually there isn't much more than what is here, the main things which are missing here are a simple UI which I added for the thriftserver part and the actual code triggering the startup of the thriftserver from the livy server. So basically the core is all here.

> At some point you will have to send an RPC to the actual Livy session to actually parse and execute statements. And that's where I don't see where the Hive classes help on the server side. You'll need state on the session side, not the server side, which only needs to know to which session to dispatch incoming requests. That's basically a hash table lookup.

Yes, basically you perfectly described what is done in the SessionManager and the OperationManager implemented here in the Livy part. What is missing is the lifecycle handling of the session and the operation on the server side, which is a bit more complicated. Each operation, for instance, can timeout and this is handled here: https://github.com/mgaido91/incubator-livy/blob/d926dfa7c6b341341433a9024915f3880b0a2804/thriftserver/src/main/java/org/apache/hive/service/cli/session/SessionManager.java#L200. More in general, the `CLIService` class should be rewritten (which is not a big deal), but we need to make sure that we introduce no leakage in the handling and this is bit less trivial (so we have to make sure we handle properly the session/operation creation, cancelation, destroy). So it can be definitely done, but I think it can be considered as a next step.

vanzin: Let's think of the timeout case: do you really need to:

- fork Hive's session manager
- have to deal with a whole bunch of things Livy doesn't need (e.g. all the metric stuff in that session manager class)
- hack things so you disable code that Livy doesn't want to run
- extend it to plumb Livy concepts into the Hive code's internal API
- maybe other things I'm missing

Just so you can avoid writing code to have a list of operations and a thread that goes through them and see if they have timed out?

It just seems to me that it is going to be more painful to match Hive and Livy concepts just to re-use some code that shouldn't really be that complicated to write - or code that even, in some cases, can just be copy & pasted and adjusted for Livy's needs without the extra baggage.

I see the benefit of copying some of the code, as I've said, like the auth stuff and the code that sets up the listening sockets, since that's required for wire compatibility. But this session management stuff I just see causing more trouble than it's worth. You've already mentioned a bunch of hacks you have to add to this code just to make it stop doing things Livy doesn't want, like Hive code trying to connect to a metastore.

So, how about this. Instead of just forking all this stuff at once, start simple. Divide this into separate features:

- adding the new module to Livy
- bringing up the thrift server
- starting and tearing down sessions
- mapping sessions to operations, and other server-side operation management
- execution of operations on the remote session

And make each one a separate PR. At that point, you can investigate what Hive code does it make sense to import. But doing it like this, bulk-importing a bunch of things, feels to me like painting ourselves into a corner - because the code is there it feels like it has to be used, instead of having a proper investigation of whether it's really beneficial to use that code.

server/src/main/scala/org/apache/livy/LivyConf.scala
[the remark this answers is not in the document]
jerryshao: We can leave this change to other PRs for better understanding.

thriftserver/pom.xml
mgaido91: Ok, I can try and rewrite the session management without relying on Hive's one. I just considered it as a next step, so anyway it is not a big deal. It may make the PRs cleaner.

I am not sure I get what you mean when separating the features, may I ask you to elaborate a bit more what you mean exactly for each of them?
> - adding the new module to Livy

here you mean just adding the new module with the dependencies, but without any code in it?
 
> - bringing up the thrift server

Including the basic code for the thrift API to be up, without providing any answer?

> - starting and tearing down sessions

this is clear I think

> - mapping sessions to operations, and other server-side operation management
> - execution of operations on the remote session

maybe we can merge the 2 above? otherwise I am not sure how to do operation management (eg. clearing things) if they have not been allocated since no execution is performed.

cc also @jerryshao for his opinion since he advised the current splitting.

Thanks you all for your review and your inputs again, much appreciated.

vanzin: I was aiming at making it clearer what parts of Hive we're borrowing, and for what reason, with those steps.

e.g., for the "bringing up the Thrift server" step, you'd bring in needed classes from Hive to make that easy (like the auth stuff), modified to work with Livy.

Similarly for the session part, if parts of Hive are needed for that part, they'd be brought in that PR.

I understand breaking things down like that would be harder than just copying a bunch of code and hacking away. But I really do feel like forking too much stuff is the wrong approach, and we should avoid that if possible. For an example of problems that approach can bring, see Spark, where upgrading Hive is a huge hassle because of how much code it forked. This is not nearly in that same ballpark, but if we can avoid that kind of hassle, I think some extra work is worth it.

I you look at it and see it's really a lot of work, I could be convinced to be ok with this fork if there's a large cleanup later to remove all of the stuff that is not needed - especially extra Hive artifacts that are being brought it an dependencies just because of the forked code.

mgaido91: I think that the first step to remove the extra dependencies is to remove the `HiveSession` from the code. For doing this, we need to re-implement the classes `CLIService`, `SessionManager` and `OperationManager`, merging the last two in the corresponding Livy classes extending them. Moreover, if you check all the methods of `CLIService`, you'll see that most of them deal with a different `Operation`. All these operations extend from `Operation` (which has a reference to the `HiveSession`) and their lifecycle is controlled by the `HiveSessionImpl` class.

So the reason why I say that it is a considerable amount of work is that we need to:
 - migrate (kind of copy with some small modification) the code from `SessionManager` (for the background pool, timeout, etc.), `OperationManager` and `CLIService` (this is not the biggest part of the work);
 - migrate somewhere the thread management, synchronization (if actually needed), and the other things which are taken care by the `HiveSessionImpl` in order to proper run, cancel an `Operation` (this, including the testing, is probably the most time consuming part);
 - either (clean solution) reimplement something analogous to all the classes under the `operation` package or (faster solution) hack them a bit in order to avoid the reference to the `HiveSession` (if we go for the clean option, this is quite time consuming too given the amount of things).

I cannot ensure that no other issue may come out in this process, but these are the required changes IMO. I am not sure how much time they could require. Anyway, I started working on it, since sooner or later, this has to be done, since we all agree that the less dependencies we bring and the less code we borrow from Hive, the better is for maintaining the solution and avoiding the many problems you correctly underlined for STS.

Hope this is a more detailed and clearer answer than the previous ones. Let me know if there is something I need to elaborate better. Thanks.

mgaido91: Of course I have not finished with this task, but I think that I can provide some starting answers. The dependencies on Hive we cannot get rid of are (at least) the modules:

 - `hive-exec`: it contains the `HiveConf` class, the returned thrift Types, the `ColumnBuffer` which is the class for returning results, ...;
 - `hive-service`: it contains the authentication methods (but we can remove the classes from here, ie. we can avoid copying some of them which we don't need to modify);
 - `hive-jdbc`: needed for testing as it contains the Hive JDBC client and for beeline to work;
 - `hive-beeline`:  needed for beeline to work...

You can find a WIP branch for this at: https://github.com/mgaido91/incubator-livy/tree/wip_no_hive_session. In this WIP I also removed the Hive classes which we don't really need to copy here as they needed no change.

So basically in this way we can remove the `hive-cli`, which brings out also `hive-common`, `hive-shims` and `hive-serde`.

Do you think this is acceptable?

mgaido91: I will try and check further but I do think getting rid of `hive-exec` is a huge effort. So, if you agree, I'd not target that for this PR. We would also need to rewrite the authentication part as it relies on HiveConf. So basically we would get rid of `hive-service` as well, but I'd really set this as put of scope for the current PR as it is gonna be a very huge effort.

The beeline command client is not needed by the server of course, but it is useful for connecting to it...in the bin folder I added the beeline script...otherwise to test connecting to the server you have to either download a hive beeline or another client... I think it is nice to have it.

vanzin: >  rewrite the authentication part as it relies on HiveConf

I think rewrite is a little too pessimistic. It shouldn't be hard to replace it with `LivyConf` or a simple wrapper around it. But ok to postpone.

> The beeline command client is not needed by the server ...  it is nice to have it.

How about adding it in a different module that the Livy server does not depend on? Then you could still even package it with Livy if you want (with the possibility of Hive folks frowning at you), but the server wouldn't have it in its classpath.

mgaido91: > I think rewrite is a little too pessimistic.

Let's call it copy and paste with some modification.  Anyway, we have to rewrite completely `ThriftBinaryCLIService`, `ThriftHttpCLIService`, the `ICLIService` interface, the `RowSet` abstraction to return the results as well as the table schema representation. They all heavily rely on classes from `hive-exec`.

> How about adding it in a different module that the Livy server does not depend on?

Yes, this is reasonable. I am doing it. Maybe we can avoid shipping it, but for dev/testing I think it is great to have it.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/SessionStates.scala
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.livy.thriftserver
+
+object SessionStates extends Enumeration {
```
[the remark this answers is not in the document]
mgaido91: I am not sure that is a good idea. The `SessionState` contains so many states, while here we have only these 3. I don't feel as a big inefficiency adding an enum...

thriftserver/server/pom.xml
```diff
@@ -0,0 +1,234 @@
...
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <executions>
+          <execution>
+            <phase>package</phase>
+            <goals>
+              <goal>copy-dependencies</goal>
```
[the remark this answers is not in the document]
mgaido91: Yes, I think you are missing a detail: this module `livy-thtiftserver` depends on the `livy-server` one, as it uses many classes which are defined there (eg. `InteractiveSession`, `SessionManger`, ...). So adding it as a dependency for `livy-server` would create a cyclic dependency in maven.

So the solution I had to avoid it was to not have this module in the `livy-server` dependencies. The thriftserver can be started using reflection from the `livy-server` module (https://github.com/apache/incubator-livy/compare/master...mgaido91:wip_no_hive_session#diff-42679418f67f742a4a87255b51b9f5bc) and the jars here are added in the assembly module as well (https://github.com/apache/incubator-livy/compare/master...mgaido91:wip_no_hive_session#diff-08fed5944cd7042953facaa12f95d2e9).

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/rpc/RpcClient.scala
```diff
@@ -0,0 +1,285 @@
...
+  def executeSql(
+      sessionHandle: SessionHandle,
+      statementId: String,
+      statement: String): JobHandle[_] = {
+    info(s"RSC client is executing SQL query: $statement, statementId = $statementId, session = " +
+      sessionHandle)
+    require(null != statementId, s"Invalid statementId specified. StatementId = $statementId")
+    require(null != statement, s"Invalid statement specified. StatementId = $statement")
+    livySession.recordActivity()
+    rscClient.submit(executeSqlJob(sessionId(sessionHandle),
```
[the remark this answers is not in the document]
mgaido91: Sorry, I am not sure I got everything about your comment. My knowledge of Livy is pretty limited at the moment, so I may miss many things, but I don't understand:

> Also note that the repl stuff does not use Jobs 

Then I don't understand how you can get customized results (I mean, not represented as string). Returning them as strings here would be very inefficient since we may have to transfer a lot of data. But as I said, I may miss something. In case, please correct me.

> Need to make sure it's ok to instantiate them when Spark jars are not available, even though when actually run on the remote side they require Spark jars to be around.

I don't have Spark classes in the Livy classpath when running the thriftserver (please notice also the tests added). So I am not sure about what you mean here.

thriftserver/server/pom.xml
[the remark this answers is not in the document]
vanzin: I looked at the assembly module again and you're right, this is needed to copy the jars into the final place. Just looks a little weird after you forget the context of how all this stuff is packaged. My bad.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/rpc/RpcClient.scala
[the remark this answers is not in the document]
vanzin: The main thing I'm saying is that the code that implements the jobs that are run in the Spark session should be in a different module, not here. You shouldn't have to add the thrift server jar to the Spark session to get this to work, especially to avoid any conflicts with Spark's Hive classes.

Mentioning the REPL was an attempt to given an example of how this was done previously, but in hindsight not a good one since that is a custom RPC. But there are examples (like "PingJob" which is in the RSC module and is used by the server module to make sure sessions are up, or the test jobs used by the integration tests).

> I don't have Spark classes in the Livy classpath....

Good, so that means that the server is able to instantiate those classes even if they reference Spark classes internally. I've seen weird problems with late binding (which make things work like in your case) vs. early binding (which would cause the instantiation to throw an error because classes are missing) and generally try to avoid relying on those semantics for things to work.

But if it works, it works.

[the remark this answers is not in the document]
mgaido91: > The main thing I'm saying is that the code that implements the jobs that are run in the Spark session should be in a different module, not here.

Yes, sure, I get this and I agree. But I don't think it can cause conflicts with the Hive classes as the classes included here are used only in the thriftserver and I don't think that in an interactive session a Spark thriftserver can be started. Anyway I agree that it is better to do this.

I won't include it here as you also mentioned that we can postpone this, but I can already create a JIRA for tracking this. Is this ok?

mgaido91:
@vanzin I updated the description adding some comments on what was done in the Hive classes. Let me know if something else is needed. Thanks.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
```diff
@@ -0,0 +1,609 @@
...
+import org.apache.livy.Logging
+import org.apache.livy.server.interactive.{CreateInteractiveRequest, InteractiveSession}
+import org.apache.livy.sessions.Spark
+import org.apache.livy.thriftserver.SessionStates._
+import org.apache.livy.thriftserver.rpc.RpcClient
+import org.apache.livy.utils.LivySparkUtils
+
+class LivyThriftSessionManager(val server: LivyThriftServer, hiveConf: HiveConf)
+  extends CompositeService(classOf[LivyThriftSessionManager].getName) with Logging {
+  private[thriftserver] val operationManager = new LivyOperationManager(this)
```
vanzin: nit: add empty line to separate from class declaration

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
```diff
@@ -0,0 +1,609 @@
...
+    hiveConf.getIntVar(ConfVars.HIVE_SERVER2_LIMIT_CONNECTIONS_PER_USER_IPADDRESS)
+  private val checkInterval = HiveConf.getTimeVar(
+    hiveConf, ConfVars.HIVE_SERVER2_SESSION_CHECK_INTERVAL, TimeUnit.MILLISECONDS)
+  private val sessionTimeout = HiveConf.getTimeVar(
+    hiveConf, ConfVars.HIVE_SERVER2_IDLE_SESSION_TIMEOUT, TimeUnit.MILLISECONDS)
+  private val checkOperation = HiveConf.getBoolVar(
+    hiveConf, ConfVars.HIVE_SERVER2_IDLE_SESSION_CHECK_OPERATION)
+
+  private var backgroundOperationPool: ThreadPoolExecutor = _
+
```
vanzin: nit: too many empty lines

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
```diff
@@ -0,0 +1,609 @@
...
+  def livySessionState(sessionHandle: SessionHandle): SessionStates = {
+    sessionHandleToLivySession.get(sessionHandle).value match {
+      case Some(Success(_)) => CREATION_SUCCESS
+      case Some(Failure(_)) => CREATION_FAILED
+      case None => CREATION_IN_PROGRESS
+    }
+  }
+
+  def numberOfActiveUsers(livySessionId: Int): Int = synchronized[Int] {
```
vanzin: Can probably drop `[Int]` from the synchronized part.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
```diff
@@ -0,0 +1,609 @@
...
+    managedLivySessionActiveUsers.getOrElse(livySessionId, 0)
+  }
+
+  def onLivySessionOpened(livySession: InteractiveSession): Unit = {
+    server.livySessionManager.register(livySession)
+  }
+
+  private def incrementManagedSessionActiveUsers(livySessionId: Int): Unit = synchronized {
```
vanzin: There's only one caller, so inline?

Another option is to make this more generic and also call it from `onUserSessionClosed`.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
```diff
@@ -0,0 +1,609 @@
...
+      createLivySession: () => InteractiveSession): InteractiveSession = {
+    sessionId match {
+      case Some(id) =>
+        server.livySessionManager.get(id) match {
+          case None =>
+            warn(s"Session id $id doesn't exist, so we will ignore it.")
+            createLivySession()
+          case Some(session) if !server.isAllowedToUse(username, session) =>
+            warn(s"Session id $id doesn't belong to $username, so we will ignore it.")
+            createLivySession()
```
vanzin: Seems like this should result in an error, not in a new session being created?

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
```diff
@@ -0,0 +1,609 @@
...
+          case Some(session) => if (session.state.isActive) {
```
vanzin: Move code to next line and indent the block.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
```diff
@@ -0,0 +1,609 @@
...
+   *  - register the new Thriftserver session in the Spark application;
+   *  - runs the initialization statements;
+   */
+  private def initSession(
+      sessionHandle: SessionHandle,
+      livySession: InteractiveSession,
+      initStatements: List[String]): Unit = {
+    // Add the thriftserver jar to Spark application as we need to deserialize there the classes
+    // which handle the job submission.
+    // Note: if this is an already existing session, adding the JARs multiple times is not a
```
vanzin: I'm not sure this is necessarily true in all versions of Spark.

But in any case, after we separate the thrift server from the jobs code, it may become easier to automatically package the needed code with the rest of the RSC jars, so this would become unnecessary.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
```diff
@@ -0,0 +1,609 @@
...
+    // problem as Spark ignores JARs which have already been added.
+    try {
+      livySession.addJar(LivyThriftSessionManager.thriftserverJarLocation(server.livyConf))
+    } catch {
+      case e: java.util.concurrent.ExecutionException
```
vanzin: I guess this is handling the Spark versions that don't do what the comment above says?

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
```diff
@@ -0,0 +1,609 @@
...
+  def openSession(
+      protocol: TProtocolVersion,
+      username: String,
+      password: String,
+      ipAddress: String,
+      sessionConf: JMap[String, String],
+      withImpersonation: Boolean,
+      delegationToken: String): SessionHandle = {
+    val sessionHandle = new SessionHandle(protocol)
+    incrementConnections(username, ipAddress, SessionInfo.getForwardedAddresses)
```
vanzin: Will this be undone of any of the things below break?

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
```diff
@@ -0,0 +1,609 @@
...
+
+  private val timeoutCheckerLock: Object = new Object
+  @volatile private var shutdown: Boolean = false
+
+  // Taken from Hive
+  private def startTimeoutChecker(): Unit = {
+    val interval: Long = Math.max(checkInterval, 3000L)
+    // minimum 3 seconds
+    val timeoutChecker: Runnable = new Runnable() {
+      override def run(): Unit = {
```
vanzin: This seems like mostly borrowed from Hive, but this is forever blocking one of the pool's threads, which kinda raises the question of why is this using the shared pool in the first case, instead of either having its own thread, or the pool allowing for scheduled tasks too.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
```diff
@@ -0,0 +1,609 @@
...
+    val res = Try(value.toInt)
+    if (res.isFailure) {
+      warn(s"Ignoring $key = $value as it is not a valid integer")
+      None
+    } else {
+      Some(res.get)
+    }
+  }
+
+  private def processSessionConf(
```
vanzin: This is ok for an initial version, but using the connection string for this is a little messy.

Hive does this generally by using "SET" commands. In the case of MR jobs, since they're started on demand, that's easy. For Hive-on-Spark, they check special options and re-start the underlying Spark session with the changed configuration.

We should think about supporting something like that at some point. No need to do anything here.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/SessionInfo.scala
```diff
@@ -0,0 +1,90 @@
...
+import org.apache.livy.Logging
+
+
+case class SessionInfo(username: String,
+    ipAddress: String,
+    forwardedAddresses: util.List[String],
+    protocolVersion: TProtocolVersion) {
+  val creationTime: Long = System.currentTimeMillis()
+}
+/**
```
vanzin: Add empty line.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/SessionInfo.scala
```diff
@@ -0,0 +1,90 @@
...
+
+package org.apache.livy.thriftserver
+
+import java.util
+
+import org.apache.hive.service.rpc.thrift.TProtocolVersion
+
+import org.apache.livy.Logging
+
+
```
vanzin: Drop empty line.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/SessionInfo.scala
```diff
@@ -0,0 +1,90 @@
...
+case class SessionInfo(username: String,
+    ipAddress: String,
+    forwardedAddresses: util.List[String],
+    protocolVersion: TProtocolVersion) {
+  val creationTime: Long = System.currentTimeMillis()
+}
+/**
+ * Mirrors Hive behavior which stores thread local information in its session manager.
+ */
+object SessionInfo extends Logging {
```
vanzin: Where is this stuff being used?

I looked for "SessionInfo.blah",  and imports that might reference this object but couldn't find anything.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/rpc/RpcClient.scala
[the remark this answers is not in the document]
vanzin: > but I can already create a JIRA for tracking this.

Sounds good. If you could create another one for cleaning up Hive dependencies that would be good too.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/serde/ColumnBuffer.scala
```diff
@@ -0,0 +1,184 @@
...
+          binaryVars.add(ByteBuffer.wrap(field.asInstanceOf[Array[Byte]]))
+        case _ =>
+          stringVars.add(DataTypeUtils.toHiveString(field, dataType))
+      }
+    }
+
+    currentSize += 1
+  }
+
+  private def checkBoolVarsSize(): Unit = if (boolVars.length == currentSize) {
```
vanzin: Could you change these to "ensureBlahVarsCapacity"? That makes is clearer that the methods are actually modifying stuff, not just checking things.

The code also uses "size" and "currentSize" interchangeably, it would be good to be consistent even if both are the same thing under the hood.

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerBaseTest.scala
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  def jdbcUri(defaultDb: String, sessionConf: String*): String = if (mode == ServerMode.http) {
+    s"jdbc:hive2://localhost:$port/$defaultDb?hive.server2.transport.mode=http;" +
+      s"hive.server2.thrift.http.path=cliservice;${sessionConf.mkString(";")}"
+  } else {
+    s"jdbc:hive2://localhost:$port/$defaultDb?${sessionConf.mkString(";")}"
+  }
+
+  override def beforeAll() {
```
vanzin: `: Unit = {`; same in other methods.

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerBaseTest.scala
```diff
@@ -0,0 +1,124 @@
...
+    val sessionManager = new InteractiveSessionManager(livyConf, ss)
+    val accessManager = new AccessManager(livyConf)
+    LivyThriftServer.start(livyConf, sessionManager, ss, accessManager)
+    LivyThriftServer.thriftServerThread.join(THRIFT_SERVER_STARTUP_TIMEOUT)
+    assert(LivyThriftServer.getInstance.isDefined)
+    assert(LivyThriftServer.getInstance.get.getServiceState == STATE.STARTED)
+  }
+
+  override def afterAll() {
+    LivyThriftServer.stopServer
```
vanzin: `stopServer()`

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyCLIService.scala
```diff
@@ -0,0 +1,416 @@
...
+    if (UserGroupInformation.isSecurityEnabled) {
+      try {
+        serviceUGI = Utils.getUGI
+      } catch {
+        case e: IOException =>
+          throw new ServiceException("Unable to login to kerberos with given principal/keytab", e)
+        case e: LoginException =>
+          throw new ServiceException("Unable to login to kerberos with given principal/keytab", e)
+      }
+      // Also try creating a UGI object for the SPNego principal
```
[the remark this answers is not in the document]
jerryshao: Is this spnego used for the auth of thrift http protocol or simply UI?

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyCLIService.scala
```diff
@@ -0,0 +1,416 @@
...
+      case GetInfoType.CLI_MAX_TABLE_NAME_LEN => new GetInfoValue(128)
+      case GetInfoType.CLI_ODBC_KEYWORDS =>
+        new GetInfoValue(ParseUtils.getKeywords(LivyCLIService.ODBC_KEYWORDS))
+      case _ => throw new HiveSQLException(s"Unrecognized GetInfoType value: $getInfoType")
+    }
+  }
+
+
+  @throws[HiveSQLException]
+  def openSession(protocol: TProtocolVersion,
```
jerryshao: nit: move the parameter to the next line (here and below).

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyCLIService.scala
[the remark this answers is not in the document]
mgaido91: @jerryshao 's answer is the right one: this is used for auth when the thrift transport mode configured is http instead of binary (see `hive.server2.transport.mode`)

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
```diff
@@ -0,0 +1,206 @@
...
+import org.apache.hadoop.hive.shims.Utils
+import org.apache.hive.service.cli._
+import org.apache.hive.service.cli.operation.Operation
+
+import org.apache.livy.Logging
+import org.apache.livy.thriftserver.SessionStates._
+import org.apache.livy.thriftserver.rpc.RpcClient
+import org.apache.livy.thriftserver.types.DataTypeUtils._
+
+class LivyExecuteStatementOperation(
```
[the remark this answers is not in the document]
mgaido91: I don't think this would need to be stored for session recovery or HA. Also in Hive, for instance, if a server goes down the operation running there is not recovered (ie. the query it was running on it is lost, you have to rerun it on the new server). We may do something better here: checking whether we have the result for the operation in the Spark session during the recovery and in that case we can return the results for it, but I don't think we need to store anything (well, we may want to store the number of records returned in case). Anyway I agree that if there are ideas how to improve this management in Livy itself, then it's great to have it.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
```diff
@@ -0,0 +1,206 @@
...
+              } catch {
+                case e: HiveSQLException =>
+                  setOperationException(e)
+                  error("Error running hive query: ", e)
+              }
+            }
+          }
+
+          try {
+            livyServiceUGI.doAs(doAsAction)
```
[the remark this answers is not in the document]
mgaido91: the reason is that this is performed in a different thread.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
```diff
@@ -0,0 +1,206 @@
...
+          } catch {
+            case e: Exception =>
+              setOperationException(new HiveSQLException(e))
+              error("Error running hive query as user : " +
+                livyServiceUGI.getShortUserName, e)
+          }
+        }
+      }
+      try {
+        // This submit blocks if no background threads are available to run this operation
```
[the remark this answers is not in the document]
mgaido91: I took it from Hive, anyway it would happen where there are so many operations running that we run out of thread in the background thread pool.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyOperationManager.scala
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  /**
+   * Cancel the running operation unless it is already in a terminal state
+   */
+  @throws[HiveSQLException]
+  def cancelOperation(opHandle: OperationHandle, errMsg: String): Unit = {
+    val operation = getOperation(opHandle)
+    val opState = operation.getStatus.getState
+    if (opState.isTerminal) { // Cancel should be a no-op in either cases
```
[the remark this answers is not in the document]
mgaido91: Sorry, this is an outdated comment (when it was written in Hive there were only 2 terminal states), a left-over from Hive code. I am fixing it, thanks.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
mgaido91: Not really, I observed that if the thriftserver location is a local path Livy tries to copy it. If this was already done one time, then this exception happens.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
mgaido91: if the things below break, the `Future` is a failure. When the session is closed this is removed. Basically, if anything of the things below break, the user gets an error for any query he/she submits. When he/she closes the connection, this is undone.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
mgaido91: As you said, this was borrowed from Hive, so I just did what was done there. I can change this if you think it is needed.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
mgaido91: well, most of the Spark SQL option are already supported through SET.  As the query is forwarded to Spark, it is like running the SET on the SparkSession, so they would take effect. I am not sure if there are other things you can set you have in mind here, though.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/SessionInfo.scala
mgaido91: it is used in `ThriftHttpServlet` and `ThriftCLIService` mainly. Previously they were in `HiveSession` so I had to move them here to get rid of it (and the `hive-cli` dependency with it).

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/rpc/RpcClient.scala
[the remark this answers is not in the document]
mgaido91: thanks, done

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
```diff
@@ -0,0 +1,206 @@
...
+    sessionHandle: SessionHandle,
+    statement: String,
+    confOverlay: JMap[String, String],
+    runInBackground: Boolean = true,
+    sessionManager: LivyThriftSessionManager)
+  extends Operation(sessionHandle, confOverlay, OperationType.EXECUTE_STATEMENT)
+    with Logging {
+
+  /**
+   * Contains the messages which have to be sent to the client.
```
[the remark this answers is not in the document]
mgaido91: Yes that is intentional (at least at the moment).

These messages are logs which are sent from the server to the client. Once the client submits a operation, then it starts polling the server for operation logs until the operation completes (then the results of the operation are submitted).

I think this is an useful feature, for instance, in the way it is used here: when you submit a query, if the spark session is not yet started, it may take a very long time to answer. Letting the user know that the reason is that the spark session is starting is a useful thing IMHO: he knows that he has to wait a bit and it is not just stuck.

.travis.yml
```diff
@@ -31,13 +31,14 @@ env:
...
+    - env: MVN_FLAG='-Pthriftserver -Pspark-2.2-it -DskipTests'
       jdk: oraclejdk8
-    - env: MVN_FLAG='-Pspark-2.2 -DskipITs'
+    - env: MVN_FLAG='-Pthriftserver -Pspark-2.2 -DskipITs'
       jdk: oraclejdk8
-    - env: MVN_FLAG='-Pspark-2.3-it -DskipTests'
+    - env: MVN_FLAG='-Pthriftserver -Pspark-2.3-it -DskipTests'
       jdk: oraclejdk8
-    - env: MVN_FLAG='-Pspark-2.3 -DskipITs'
+    - env: MVN_FLAG='-Pthriftserver -Pspark-2.3 -DskipITs'
```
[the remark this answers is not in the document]
mgaido91: > Does thriftserver module depends on specific Spark module?

No it doesn't. It depends on Java 8 only.

> I would suggest to separate thriftserver mode to the new test bundle.

Sorry, I am not about what you mean by this. May you please elaborate more what is your suggestion here? Thanks.

server/src/main/scala/org/apache/livy/LivyConf.scala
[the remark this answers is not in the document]
mgaido91: Sorry, I forgot to answer here somehow. 
The point is that without adding the JAR this the tests don't pass. I can remove the configuration, as in the tests the location of the JAR can be inferred automatically, but I cannot remove the adding of the JAR at all. Is this OK?

.travis.yml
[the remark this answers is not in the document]
jerryshao: Travis config file can specify different tests in matrix, and travis could run all the tests in parallel, as you can see we separate different Spark version tests. What I'm saying is that we could add new test in this matrix for thriftserver, so that it can also be tested in parallel. You can check the travis doc to understand more about travis configuration.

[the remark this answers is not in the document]
mgaido91: I see, but then shall we run on the tests only on the thriftserver module there? I don't think that is feasible, because we would need to build the other Livy modules as well. So in that case we would just end up doing the same thing done here IIUC.

So I am not sure how you are thinking to split this. specifically.

thriftserver/server/src/main/java/org/apache/hive/service/cli/ColumnBasedSet.java
```diff
@@ -0,0 +1,205 @@
...
+import org.apache.thrift.protocol.TProtocol;
+import org.apache.thrift.transport.TIOStreamTransport;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * ColumnBasedSet.
+ */
+public class ColumnBasedSet implements RowSet {
```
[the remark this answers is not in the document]
mgaido91: Yes, you're right. That is guaranteed because the livy-thriftserver JAR is the first one being loaded from this module, so its classes are loaded first.

Anyway, I am going on in the work for cleaning up from Hive classes, despite I think it will take about 1-2 months to be completed.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyCLIService.scala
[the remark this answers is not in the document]
mgaido91: yes, it is part of the work for getting rid of the Hive dependencies...

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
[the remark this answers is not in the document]
mgaido91: To be honest, I am not sure, I haven't met this condition in my tests,  but it should block since we are using a BlockingQueue when creating the `backgroundOperationPool`.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
mgaido91: The some problem would happen if the user specifies a non-existing session id. Then we should throw an error in that case too. And the same when the session is not active anymore. What do you think?

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
mgaido91: I'll open a JIRA to fix it, so we can do it with more focus on the change and since it doesn't compromise the functionality seems OK to me to postpone this change. Thanks.

jerryshao:
Sorry I'm quite busy these days, will take a review before the end of this week.

thriftserver/server/src/main/java/org/apache/hive/service/cli/ColumnBasedSet.java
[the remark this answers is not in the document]
vanzin: >  That is guaranteed because the livy-thriftserver JAR is the first one being loaded from this module

But how are you guaranteeing that? If you just tell java to use "jars/*" as the classpath, it does not guarantee you the order in which those jars will be loaded. And that's the current way stuff is added to the Livy server's classpath.

> I think it will take about 1-2 months to be completed.

How much bandwidth will you have to spend on this? I'm just trying to gauge whether it's worth for me to try to implement things the way I suggested previously (incremental steps pulling Hive stuff as needed). I still think in the end that would result in better code, but if you're going to be working on this too then it's probably better to spend those resources on improving your code.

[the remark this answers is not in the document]
mgaido91: > But how are you guaranteeing that?

I am sorry I cannot provide a definitive answer to this as I am not a class-loader expert honestly. My understanding is that when the LivyServer starts, it starts the LivyThriftServer class (though reflection) so at this point the first JAR loaded from the module is livy-thriftserver and the classes there are loaded first. The same happens also on Spark. 

> How much bandwidth will you have to spend on this?

I am working on it about 50% of my time. But I am a bit stuck and any help on this would be super-appreciated. Actually I have a [WIP](https://github.com/mgaido91/incubator-livy/commit/cb00753acc15d49469aed82fd6911ae36bd09e98) (which is really a WIP - I mean, it doesn't even compile at the moment since I have not finished removing all the parts referencing hive-exec's classes) where I did most of the rewrite work (it needs improvement, I know, but my first goal was to solve all the dependency issues, then the cleanup can be done), but the most critical parts are still missing and I am not sure how to resolve them: the hard part (at least or me) is replacing the `HiveAuthFactory` functionalities. If you can help me with this, we can probably already include the dependency removal in this PR.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
bikassaha: +1 for erroring out in all cases where the user expected session is not available.

.travis.yml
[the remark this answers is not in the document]
jerryshao: Never mind, we can improve it later.

pom.xml
```diff
@@ -83,6 +83,7 @@
     <spark.scala-2.11.version>1.6.2</spark.scala-2.11.version>
     <spark.scala-2.10.version>1.6.2</spark.scala-2.10.version>
     <spark.version>${spark.scala-2.11.version}</spark.version>
+    <hive.version>3.0.0</hive.version>
```
jerryshao: Do we have potential issues using hive 3.0.0, especially during test? Spark itself uses Hive 1.2.1, it has many issues when both two versions of Hive existed in Spark's classpath.

Can we just parameterize this configuration, also which kind of Hive 3.0 features are we using here? Is it possible to make it hive version independent?

thriftserver/server/pom.xml
```diff
@@ -0,0 +1,234 @@
...
+        <exclusion>
+          <groupId>org.apache.hive</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.hive</groupId>
+      <artifactId>hive-exec</artifactId>
+      <version>${hive.version}</version>
```
jerryshao: We may also need to exclude Hadoop related dependencies here.

pom.xml
```diff
@@ -1077,6 +1080,39 @@
   </distributionManagement>
 
   <profiles>
+    <profile>
+      <id>thriftserver</id>
+      <modules>
+        <module>thriftserver/server</module>
+      </modules>
+      <properties>
+        <jetty.version>9.3.8.v20160314</jetty.version>
```
jerryshao: Why can't we upgrade jetty version there in the parent maven, rather than make a special case in thrift module?

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyCLIService.scala
```diff
@@ -0,0 +1,432 @@
...
+        case e: IOException =>
+          throw new ServiceException("Unable to login to kerberos with given principal/keytab", e)
+        case e: LoginException =>
+          throw new ServiceException("Unable to login to kerberos with given principal/keytab", e)
+      }
+      // Also try creating a UGI object for the SPNego principal
+      val principal = hiveConf.getVar(ConfVars.HIVE_SERVER2_SPNEGO_PRINCIPAL)
+      val keyTabFile = hiveConf.getVar(ConfVars.HIVE_SERVER2_SPNEGO_KEYTAB)
+      if (principal.isEmpty || keyTabFile.isEmpty) {
+        info(s"SPNego httpUGI not created, spNegoPrincipal: $principal, ketabFile: $keyTabFile")
```
jerryshao: nit: "SPNego". Also it seems that spnego ugi can be nullable, is it acceptable in the code?

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyCLIService.scala
```diff
@@ -0,0 +1,432 @@
...
+   */
+  @throws[HiveSQLException]
+  override def executeStatement(
+      sessionHandle: SessionHandle,
+      statement: String,
+      confOverlay: util.Map[String, String],
+      queryTimeout: Long): OperationHandle = {
+    val opHandle: OperationHandle = sessionManager.operationManager.executeStatement(
+      sessionHandle, statement, confOverlay, runAsync = false, queryTimeout)
+    debug(sessionHandle + ": executeStatement()")
```
jerryshao: Such debug log seems not so meaningful, would you please remove it or enrich it.

jerryshao:
Hi @mgaido91 , would you please provide us a diff about what you changed to the Hive code? Currently I don't know which part is changed in Hive code, and which part is just simply pulled from Hive.

I think we can merge as it is and play it locally to see how it is. but I agree with Marcelo that we should not ship or suggest user to use it until it is stable enough.

pom.xml
mgaido91: I don't think we can have issues. Here we are using the Thriftserver functionalities in Hive (the RPC layer as well as other classes related to the thriftserver). There is no conflict with Spark's Hive 1.2.1 because this classes are not put in Spark's applications classpath, they are only present in Livy's classpath.

I am not sure what you mean by parametrize the configuration. This is already a parameter, right? I don't think it can be Hive version independent because we need to specify which RPC interfaces we are using: different RPC layers may not be compatible, ie. if you use a Hive 1.2 beeline client to connect to this 3.0 server, probably it won't work.

thriftserver/server/pom.xml
mgaido91: Let me check whether it is feasible. Anyway, middle/long term with the refactor, we will get rid of this whole.

pom.xml
mgaido91: Here is the reason: https://github.com/apache/incubator-livy/pull/104#discussion_r211006082

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyCLIService.scala
mgaido91: yes, this is acceptable when we are in binary mode for instance. This is taken from the analogous class in Hive.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyCLIService.scala
mgaido91: This is the same in Hive. While doing some debug I think is kind of useful because this is the entry point for the client requests, so having this debug allows to know which are the calls performed by the client and make easier to understand what was going on in the server.

If you really think that we should remove this, I can do that, but I'd prefer to keep it. Thanks.

mgaido91:
>  I agree with Marcelo that we should not ship or suggest user to use it until it is stable enough.

I agree too. I think at very least, we should complete all the JIRAs under LIVY-489 to consider the first version of it to be ready. Then more testing is needed too in order to make it stable. Only then we can consider to ship/suggest this.

> would you please provide us a diff about what you changed to the Hive code? 

Here is the diff. Please notice that now this diff is quite large because of the removal of some Hive dependencies.

[hive-diff.txt](https://github.com/apache/incubator-livy/files/2360574/hive-diff.txt)

pom.xml
jerryshao: I was thinking that if we can use 1.2.1 for community Livy, maybe this could avoid some potential issues, also 3.0.0 seems too new to most of the users.

jerryshao:
One unrelated flaky issue. I'm going to merge this so that we could verify it locally. @mgaido91 you can continue the work based on this. Thanks!

mgaido91:
Thanks @jerryshao and @vanzin

PR #105, opened 2018-08-17
[LIVY-498][REPL] Fix Windows CRLF line ending issue in SparkR interpreter

jerryshao:

If the issued query contains CRLF EOL, it will be failed to execute on *nix machine. This happens when submitting queries from Windows machine and executing on Linux machine.

So here propose to convert statement to match system's EOL.

## How was this patch tested?

New UT added.

### DIFF
--- a/core/src/main/scala/org/apache/livy/EOLUtils.scala
+++ b/core/src/main/scala/org/apache/livy/EOLUtils.scala
@@ -1,0 +1,116 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


jerryshao:
@ajbozarth @zjffdu would you please help to review, thanks!

core/src/main/scala/org/apache/livy/EOLUtils.scala
```diff
@@ -0,0 +1,116 @@
...
+
+  def hasSystemDefaultEOL(s: String): Boolean = Mode.determineEOL(s) == Mode.SYSTEM_DEFAULT
+
+  def convertToUnixEOL(s: String): String = convertLineEndings(s, EOL_UNIX)
+
+  def convertToWindowsEOL(s: String): String = convertLineEndings(s, EOL_WINDOWS)
+
+  def convertToOldMacEOL(s: String): String = convertLineEndings(s, EOL_OLD_MAC)
+
+  def convertToSystemEOL(s: String): String = convertLineEndings(s, EOL_SYSTEM_DEFAULT)
```
zjffdu: Do we need other covert method ? I see only this method is used by livy

jerryshao: Yes, this is the one only required, I can remove others if you think they're not necessary.

ajbozarth: Was your intent by having all these options to reuse this file in other places? because if it's just for this one Livy use case then this file is a bit excessive.

ajbozarth:
The code LGTM, but I don't have easy access to a windows machine to test it out though (my team is all MacBooks and linux VMs)

jerryshao:
I'm going to merge this if there's no other comment, thanks!

jerryshao:
Merging to master and 0.5. Thanks for the review.

PR #106, opened 2018-09-08
fix bug -- looks like I can not submit a local jar

firemonk9:
## Fix the bug : It looks like I can not submit a local jar in batch job rest api. with error message, Local path ...cannot be added to user sessions.

https://groups.google.com/a/cloudera.org/forum/#!topic/livy-user/mm-XEhANDHU
Iinclude a link to the associated JIRA and make sure to add a link to this pr on the JIRA as well)

## The patch is tested manually.


vanzin:
Your code doesn't compile. Could you instead send an e-mail to the mailing list explaining your issue, or if you're really sure it's a bug, just file a bug in https://issues.apache.org/jira/browse/LIVY?

PR #107, opened 2018-09-10
[LIVY-494] Add thriftserver to Livy server

mgaido91:

The PR adds a configuration parameter in order to startup also the thriftserver when starting Livy server.

Apart from this trivial change, other 3 main things were needed and are present in this PR:

 - Add the thriftserver JARs to the assembly and the livy-server script;
 - A small refactor in order to enforce impersonation in the `*Session` classes, instead of in the `*Servlet` ones, so that it is picked up by the thriftserver module too (this change is not strictly needed, but I consider it a better option that duplicating this logic in the thriftserver module too);
 - Creating a UGI from the configured keytab. This is needed because the thriftserver requires the UGI to be created from a keytab in order to work properly and previously Livy was using a UGI generated from the cached TGT (created by the `kinit` command).

## How was this patch tested?

Manual test: starting the server and having it up for more than 9 days.

### DIFF
--- a/server/src/main/scala/org/apache/livy/server/LivyServer.scala
+++ b/server/src/main/scala/org/apache/livy/server/LivyServer.scala
@@ -117,11 +117,12 @@ class LivyServer extends Logging {
         error("Failed to run kinit, stopping the server.")
         sys.exit(1)
       }
-      // This is and should be the only place where a login() on the UGI is performed. Any
-      // other login should be avoided as it would change the current UGI leading to potential
-      // very bad conditions which may vary from not finding valid Kerberos credentials to other
-      // issues. If an other login in the codebase is strictly needed, a needLogin check should
-      // be added in order to avoid anyway that 2 login are performed.
+      // This is and should be the only place where a login() on the UGI is performed.
+      // If an other login in the codebase is strictly needed, a needLogin check should be added to
+      // avoid anyway that 2 logins are performed.
+      // This is needed because the thriftserver requires the UGI to be created from a keytab in
+      // order to work properly and previously Livy was using a UGI generated from the cached TGT
+      // (created by the kinit command).
       if (livyConf.getBoolean(LivyConf.THRIFT_SERVER_ENABLED)) {
         UserGroupInformation.loginUserFromKeytab(launch_principal, launch_keytab)
       }
@@ -144,14 +145,6 @@ class LivyServer extends Logging {
 
     server = new WebServer(livyConf, host, port)
     server.context.setResourceBase("src/main/org/apache/livy/server")
-
-    def invokeOnStaticThriftserver(method: String, args: (Class[_], Object)*): Object = {
-      val thriftserverObj = Class.forName("org.apache.livy.thriftserver.LivyThriftServer$")
-        .getField("MODULE$").get(null)
-      val (classes, values) = args.unzip
-      val m = thriftserverObj.getClass.getMethod(method, classes: _*)
-      m.invoke(thriftserverObj, values: _*)
-    }
 
     val livyVersionServlet = new JsonServlet {
       before() { contentType = "application/json" }
@@ -286,12 +279,8 @@ class LivyServer extends Logging {
     })
 
     if (livyConf.getBoolean(LivyConf.THRIFT_SERVER_ENABLED)) {
-      invokeOnStaticThriftserver(
-        "start",
-        (classOf[LivyConf], livyConf),
-        (classOf[InteractiveSessionManager], interactiveSessionManager),
-        (classOf[SessionStore], sessionStore),
-        (classOf[AccessManager], accessManager))
+      ThriftServerFactory.getInstance.start(
+        livyConf, interactiveSessionManager, sessionStore, accessManager)
     }
 
     _serverUrl = Some(s"${server.protocol}://${server.host}:${server.port}")
--- a/server/src/main/scala/org/apache/livy/server/LivyServer.scala
+++ b/server/src/main/scala/org/apache/livy/server/LivyServer.scala
@@ -56,6 +56,8 @@ class LivyServer extends Logging {
   private var executor: ScheduledExecutorService = _
   private var accessManager: AccessManager = _
 
+  private var ugi: UserGroupInformation = _
+
   def start(): Unit = {
     livyConf = new LivyConf().loadFromFile("livy.conf")
     accessManager = new AccessManager(livyConf)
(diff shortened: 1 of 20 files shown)


mgaido91:
cc @jerryshao @vanzin

mgaido91:
kindly ping @jerryshao @vanzin

server/src/main/scala/org/apache/livy/server/LivyServer.scala
```diff
@@ -115,6 +117,15 @@ class LivyServer extends Logging {
         error("Failed to run kinit, stopping the server.")
         sys.exit(1)
       }
+      // This is and should be the only place where a login() on the UGI is performed. Any
```
vanzin: The comment you have in the PR description, that the current Hive Thrift code needs this, is way better here, since otherwise it's unclear why you need this login at all.

The rest of the comment is basically a long way of explaining the `assert` you added later in this class.

server/src/main/scala/org/apache/livy/server/LivyServer.scala
```diff
@@ -134,6 +145,14 @@ class LivyServer extends Logging {
     server = new WebServer(livyConf, host, port)
     server.context.setResourceBase("src/main/org/apache/livy/server")
 
+    def invokeOnStaticThriftserver(method: String, args: (Class[_], Object)*): Object = {
```
vanzin: Instead of reflection hacks I'd use an interface to define a factory method to be called. You could use a `java.util.ServiceLoader` to load the class too, but with a single one, it's fine to use `Class.forName`.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -157,24 +158,12 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
...
-      target: Option[String],
-      req: HttpServletRequest): Option[String] = {
-    if (livyConf.getBoolean(LivyConf.IMPERSONATION_ENABLED)) {
-      if (!target.map(hasSuperAccess(_, req)).getOrElse(true)) {
-        halt(Forbidden(s"User '${remoteUser(req)}' not allowed to impersonate '$target'."))
-      }
-      target.orElse(Option(remoteUser(req)))
-    } else {
-      None
+  protected def withHaltOnForbiddenAction(f: => S): S = {
```
vanzin: Not a fan of the name; just `haltOnForbiddenAction` sounds better.

Also, it's been a long time since I looked at scalatra... but could you do this instead in the `error` handler of SessionServlet, and map the exception to the http error? Or is the "halt" required for some reason?

server/src/main/scala/org/apache/livy/sessions/Session.scala
```diff
@@ -133,6 +134,41 @@ object Session {
...
+   *         disabled.
+   */
+  def checkImpersonation(
+      target: Option[String],
+      requestUser: String,
+      livyConf: LivyConf,
+      accessManager: AccessManager): Option[String] = {
+    if (livyConf.getBoolean(LivyConf.IMPERSONATION_ENABLED)) {
+      if (!target.forall(hasSuperAccess(_, requestUser, accessManager))) {
+        throw new InvalidParameterException(
```
vanzin: `AccessControlException` makes more sense here.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
mgaido91: I think it is doable. I'll try and do that. Thanks.

server/src/main/scala/org/apache/livy/server/ThriftServerFactory.scala
```diff
@@ -0,0 +1,40 @@
...
+  def start(
+    livyConf: LivyConf,
+    livySessionManager: InteractiveSessionManager,
+    sessionStore: SessionStore,
+    accessManager: AccessManager): Unit
+}
+
+object ThriftServerFactory {
+  def getInstance: ThriftServerFactory = {
+    Class.forName("org.apache.livy.thriftserver.LivyThriftServer$").getField("MODULE$").get(null)
```
vanzin: Instead of this, have a `ThriftServerFactoryImpl` class that implements the `start` method, and instantiate it with `Class.forName(blah).newInstance()`. That's closer to what `ServiceLoader` does.

server/src/main/scala/org/apache/livy/sessions/Session.scala
```diff
@@ -133,6 +134,40 @@ object Session {
...
+
+  /**
+   * Checks that the requesting user can impersonate the target user.
+   * If the user does not have permission to impersonate, then throws an `AccessControlException`.
+   *
+   * @return The user that should be impersonated. That can be the target user if defined, the
+   *         request's user - which may not be defined - otherwise, or `None` if impersonation is
+   *         disabled.
+   */
+  def checkImpersonation(
```
vanzin: Since this method requires an `accessManager`, wouldn't it be better as a method in the `AccessManager` class itself? (Same for `hasSuperAccess`.)

server/src/main/scala/org/apache/livy/server/LivyServer.scala
```diff
@@ -115,6 +117,16 @@ class LivyServer extends Logging {
         error("Failed to run kinit, stopping the server.")
         sys.exit(1)
       }
+      // This is and should be the only place where a login() on the UGI is performed.
```
vanzin: This should be pretty easy to check with a scalastyle rule that disallows calls to `UserGroupInformation.login*` (and disabling the check around this call). Not foolproof, but should catch inadvertent use at least.

server/src/main/scala/org/apache/livy/sessions/Session.scala
mgaido91: sure, I agree. I'll move `hasViewAccess` and `hasModifyAccess` too for coherency. Seems also that the `AccessManager` is the right place for putting the logic of these methods IMHO. Thanks.

server/src/main/scala/org/apache/livy/server/LivyServer.scala
mgaido91: Actually this wouldn't really work as - for instance - `UserGroupInformation.loginUserFromKeytabAndReturnUGI` is ok to be called, as it doesn't perform a `UserGroupInformation.login()`. We can list all the APIs which we should not call, but I am not sure it is worth.

mgaido91:
Travis failure is intermittent, the branch build passed: https://travis-ci.org/mgaido91/incubator-livy/builds/432380763

vanzin:
Merging to master.

PR #108, opened 2018-09-10
[MINOR] Update version of ffi.

vanzin:
The other updates were made when building the docs, more exactly when
trying to fix "bundle install", since things wouldn't build with the
existing dependencies.

### DIFF
--- a/docs/Gemfile.lock
+++ b/docs/Gemfile.lock
@@ -19,7 +19,7 @@ GEM
     execjs (2.7.0)
     faraday (0.12.1)
       multipart-post (>= 1.2, < 3)
-    ffi (1.9.24)
+    ffi (1.9.25)
     forwardable-extended (2.6.0)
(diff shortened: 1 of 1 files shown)


vanzin:
@ajbozarth since you made these kinds of changes before; not sure whether the non-ffi changes are needed, but it's the only way they would build.

vanzin:
There's a more recent version, let me try to update to that one instead.

ajbozarth:
As long as you've tested building the docs I'll trust you it good

vanzin:
They build fine and clicking around looked ok. I just don't know what all this stuff is actually used for...

ajbozarth:
Crazy enough, everything in the `Gemfile.lock` is part of the dependency tree of the few actual dependancies in the `Gemfile`

vanzin:
Merging to master.

PR #111, opened 2018-09-17
[LIVY-510] Remove support for JDK7.

vanzin:
With this change Livy will only build and run with JDK8. While this
doesn't necessarily preclude using Java 7 for running Spark itself,
that is not encouraged due to possible issues with serialization.

There are a few small changes elsewhere that I made so tests would pass,
but aren't related to the java 8 change:

- a small tweak to the fallback code that initializes the python
  interpreter, which was failing with a weird reflection error.

- expose the user's home directory in the MiniYarnCluster, to allow
  the tests to see pip packages installed in the user's home.

- tweak the python API tests so that the pytest cache does not pollute
  the source directory.

- changed Travis to use Oracle's JDK, since the Open JDK path was
  mixing Java 8 and 9 and causing errors.

- updated jacoco to latest stable.

### DIFF
--- a/repl/src/main/scala/org/apache/livy/repl/PythonInterpreter.scala
+++ b/repl/src/main/scala/org/apache/livy/repl/PythonInterpreter.scala
@@ -201,9 +201,11 @@ object PythonInterpreter extends Logging {
           .newInstance(parts(0), gateway)
       } catch {
         case NonFatal(e) =>
-          classOf[PythonProxyHandler].getConstructor(classOf[String],
-            Class.forName("py4j.CallbackClient"), classOf[Gateway])
-            .newInstance(parts(0), gateway.getCallbackClient, gateway)
+          val cbClient = gateway.getClass().getMethod("getCallbackClient").invoke(gateway)
+          val cbClass = Class.forName("py4j.CallbackClient")
+          classOf[PythonProxyHandler]
+            .getConstructor(classOf[String], cbClass, classOf[Gateway])
+            .newInstance(parts(0), cbClient, gateway)
       }
 
       Proxy.newProxyInstance(Thread.currentThread.getContextClassLoader,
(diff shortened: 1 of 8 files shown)


vanzin:
The build errors seem to be because of some oddity in the travis configs... it seems to be compiling the sources with jdk9 and running tests with jdk8. I can reproduce locally with jdk 10 (compiled with jdk 10, running with jdk 8):

```
$ java b
Exception in thread "main" java.lang.NoSuchMethodError: java.nio.ByteBuffer.flip()Ljava/nio/ByteBuffer;
        at b.main(b.java:5)
```

I'll try to figure out what's going wrong...

vanzin:
Anyone feel like reviewing this?

jerryshao:
Sorry for late response. generally LGTM. I will take a look at the details.

BTW, are you going to remove Scala 2.10 support in the following PR?

repl/src/main/scala/org/apache/livy/repl/PythonInterpreter.scala
```diff
@@ -201,9 +201,11 @@ object PythonInterpreter extends Logging {
           .newInstance(parts(0), gateway)
       } catch {
         case NonFatal(e) =>
-          classOf[PythonProxyHandler].getConstructor(classOf[String],
-            Class.forName("py4j.CallbackClient"), classOf[Gateway])
-            .newInstance(parts(0), gateway.getCallbackClient, gateway)
```
jerryshao: what kind of issue happens here? I didn't meet this reflection issue before.

vanzin: The error was a "NoSuchMethodError" complaining about `getCallbackClient`. That generally happens if the method's return type changes, but I couldn't find a py4j change in that area...

Anyway, things worked with the updated code.

jerryshao:
Merging to master branch!

PR #112, opened 2018-09-21
[LIVY-511][LIVY-512] Remove support for old Spark, Scala versions.

vanzin:
This change restricts Livy support to Spark 2.2+ and Scala 2.11.  Both changes
are made together because by supporting Spark 2.2+ only, it becomes impossible
to test Scala 2.10.

As part of the change, a lot of code that used reflection to support different
versions of Spark could be cleaned up and directly call Spark APIs.

The Scala 2.10 parts of the builds also have been removed, but the actual
support for building and running with different Scala versions (and related
tests) have been left as is. This will allow us to support 2.12 in the future.

This change intentionally does not touch the public API (the "api/" module).
There are things that could be cleaned up now that Spark 1.x is not supported,
but that would mean an API breakage so I chose to leave those alone for now.

The test matrix and build profiles have also been simplified a lot.  There are
now two profiles to choose from (for Spark 2.2 and 2.3); integration tests can
be run against a different version of Spark by running just the integration
test module with the desired profile.

Tested with Spark 2.2 and 2.3, and also by building against 2.2 and running
integration tests against 2.3.


.travis.yml
```diff
@@ -19,24 +19,12 @@ sudo: required
...
   include:
-      # Spark 2.2+ will only be verified using JDK8
-      # Thriftserver requires JDK8
-    - env: MVN_FLAG='-Pthriftserver -Pspark-2.2-it -DskipTests'
-    - env: MVN_FLAG='-Pthriftserver -Pspark-2.2 -DskipITs'
-    - env: MVN_FLAG='-Pthriftserver -Pspark-2.3-it -DskipTests'
-    - env: MVN_FLAG='-Pthriftserver -Pspark-2.3 -DskipITs'
-
+  - name: "Spark 2.2 + ITs"
+    env: MVN_FLAG='-Pthriftserver'
```
mgaido91: I remember a comment by @jerryshao when he mentioned that there is a timeout for a single build in travis but we have not limit on the number of runs. If this is the case, what about restoring having one build with `-DskipTests` and one with `-DskipITs` in order to reduce the time of each of them?

README.md
```diff
@@ -57,12 +57,8 @@ Required python packages for building Livy:
 To run Livy, you will also need a Spark installation. You can get Spark releases at
 https://spark.apache.org/downloads.html.
 
-Livy requires at least Spark 1.6 and supports both Scala 2.10 and 2.11 builds of Spark, Livy
-will automatically pick repl dependencies through detecting the Scala version of Spark.
-
-Livy also supports Spark 2.0+ for both interactive and batch submission, you could seamlessly
-switch to different versions of Spark through ``SPARK_HOME`` configuration, without needing to
-rebuild Livy.
+Livy requires Spark 2.2 or 2.3. You can switch to a different version of Spark by setting the
```
mgaido91: Is there a specific reason not to state just Spark 2.2+? Do we have problems with 2.4?

.travis.yml
vanzin: The builds seem to be taking a little bit over 40m. If that's close to the timeout we can break them down.

README.md
vanzin: The code explicitly leaves 2.4 out. In fact, even if it didn't, 2.4 wouldn't work because the REPL API has changed.

mgaido91: I see, thanks for the explaination.

vanzin:
Any more comments? I'd like this to get in to unblock my changes for LIVY-503 (which right now depend on Spark 2.2).

.travis.yml
mgaido91: Seems like the limit is 50m, so we are pretty close to it: https://docs.travis-ci.com/user/customizing-the-build/#build-timeouts.

Can we have separated builds? Thanks.

mgaido91:
@vanzin thanks, I have no more comments apart from https://github.com/apache/incubator-livy/pull/112#discussion_r220647323. I have not reviewed it carefully, but I am still quite new to Livy codebase, so it would take a lot for me to do a throughout review of this PR, which is pretty huge.

So once you address that comment I am fine with this PR, thanks.

mgaido91:
thanks @vanzin

vanzin:
ok, merging to master.

PR #113, opened 2018-09-25
[LIVY-500] Add beeline client for dev

mgaido91:

The PR adds a new `thriftserver/client` module and a `beeline` script which are useful to have a client for the thriftserver during local testing. Both things are intended for dev usage only and are not shipped with the distribution.

## How was this patch tested?

manual test

### DIFF
--- a/thriftserver/client/pom.xml
+++ b/thriftserver/client/pom.xml
@@ -22,8 +22,10 @@
     <artifactId>livy-main</artifactId>
     <groupId>org.apache.livy</groupId>
     <version>0.6.0-incubating-SNAPSHOT</version>
+    <relativePath>../../pom.xml</relativePath>
   </parent>
   <modelVersion>4.0.0</modelVersion>
+  <packaging>pom</packaging>
 
   <artifactId>livy-beeline</artifactId>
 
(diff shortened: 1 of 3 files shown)


mgaido91:
cc @vanzin

thriftserver/client/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>livy-main</artifactId>
+    <groupId>org.apache.livy</groupId>
+    <version>0.6.0-incubating-SNAPSHOT</version>
```
vanzin: missing `relativePath`

thriftserver/client/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+            <goals>
+              <goal>copy-dependencies</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
```
vanzin: There's no code here, right? Shouldn't the pom's packaging be just "pom" instead of the default ("jar")? Or does that mess up the dependency plugin?

thriftserver/client/pom.xml
mgaido91: added,thanks.

vanzin:
Merging to master.

PR #116, opened 2018-09-27
[LIVY-520] Add log when session is GC-ed

mgaido91:

When a session is inactive for a long time, it is GC-ed, ie. it is closed. Currently this operation is not logged, so it is not easy to understand why a session was closed.

The PR proposes to add a log in order to explain what happened.

## How was this patch tested?

manual tests


vanzin:
Merging to master.

PR #117, opened 2018-09-28
[LIVY-502] Remove dependency on hive-exec

mgaido91:

This PR removes the dependency on `hive-exec`. Only modules of Hive which are used after this PR are `hive-service-rpc` and `hive-service`. This drastically reduces the amount of JARs needed by the thriftserver module.

The PR takes the Hive classes which we were using and adapts them when necessary (or simpify when we don't need something) in order to work in the Livy thriftserver.

Most of the classes are just migrated replacing all occurrences of `HiveConf`, `HiveSession` and other Hive specific classes. Only one class has a quite different logic than Hive's and it is `AuthFactory`, as we are using a different UGI handling from Hive (as we are not running the TS in a standalone JVM).

The functionalities we are taking from Hive are:
 - the thrift protocol endpoints (the classes in the `cli` package), ie. the classes handling lower level details about the communication with the client;
 - the authentication layer (classes in the `auth` package, which are used in the ones in `cli`), despite this PR leaves LDAP and PAM as out of scope as they are not trivial to be ported and not needed for a working solution. We can add them later;
  - the classes in the `operation` package which are used to answer to metadata queries by the JDBC driver.

## How was this patch tested?

existing UTs + manual tests


mgaido91:
cc @jerryshao  @vanzin

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/auth/AuthenticationProvider.scala
```diff
@@ -0,0 +1,74 @@
...
+package org.apache.livy.thriftserver.auth
+
+import java.lang.reflect.InvocationTargetException
+import javax.security.sasl.AuthenticationException
+
+import org.apache.hive.service.auth.PasswdAuthenticationProvider
+
+import org.apache.livy.LivyConf
+
+object AuthenticationProvider {
```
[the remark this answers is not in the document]
mgaido91: Yes, I agree. Anyway, I don't think there is much difference as of now, ie. both APIs support Kerberos or none (this supports custom too at the moment, it i true, but I am not even sure how widespread its adoption is in Hive).

Anyway I think we can revisit this later. This is a very sensitive part and I think a PR focused only on this would be better.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/auth/LivyDelegationTokenSecretManager.scala
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.hadoop.io.Text
+import org.apache.hadoop.security.token.Token
+import org.apache.hadoop.security.token.delegation.{AbstractDelegationTokenIdentifier, AbstractDelegationTokenSecretManager}
+
+import org.apache.livy.LivyConf
+
+/**
+ * A secret manager. It is taken from analogous implementation in the MapReduce client.
+ */
+class LivyDelegationTokenSecretManager(val livyConf: LivyConf)
```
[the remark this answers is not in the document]
mgaido91: I don't think we can remove it. In particular, the `getTokenIdentifier` and `verifyDelegationToken` are used or the client authentication.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/cli/ThriftCLIService.scala
```diff
@@ -0,0 +1,745 @@
...
+    }
+  }
+
+  override def run(): Unit
+
+  /**
+   * If the proxy user name is provided then check privileges to substitute the user.
+   */
+  @throws[HiveSQLException]
+  private def getProxyUser(
```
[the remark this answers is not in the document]
mgaido91: I agree. As previously, I'd do this in a separate PR in order to focus carefully on how to change these parts since they are sensitive.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/cli/ThriftHttpCLIService.scala
```diff
@@ -0,0 +1,210 @@
...
+
+import org.apache.livy.LivyConf
+import org.apache.livy.thriftserver.LivyCLIService
+import org.apache.livy.thriftserver.auth.AuthFactory
+
+/**
+ * This class is ported from Hive. We cannot reuse Hive's one because we need to use the
+ * `LivyCLIService`, `LivyConf` and `AuthFacotry` instead of Hive's one.
+ */
+class ThriftHttpCLIService(
```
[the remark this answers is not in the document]
mgaido91: I am not sure about this. Here we have a dedicated thread-pool associated with this which is separate from Livy's HTTP server. I am not sure it is feasible. We'd need to try.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/cli/ThriftHttpServlet.scala
```diff
@@ -0,0 +1,500 @@
...
+      case e: HiveSQLException => throw new HttpAuthenticationException(e);
+    }
+  }
+
+  /**
+   * Do the GSS-API kerberos authentication. We already have a logged in subject in the form of
+   * serviceUGI, which GSS-API will extract information from.
+   * In case of a SPNego request we use the httpUGI, for the authenticating service tickets.
+   */
+  private def doKerberosAuth(request: HttpServletRequest): String = {
```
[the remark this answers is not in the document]
mgaido91: mmh...IIRC there was no such filter which was able to do exactly the same thing which is done here. We might have to adapt/extend one. But if we cannot just reuse it I am not sure it is worth switching to them.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/GetTypeInfoOperation.scala
```diff
@@ -0,0 +1,130 @@
...
+import org.apache.livy.thriftserver.types.{BasicDataType, Field, Schema}
+
+sealed case class TypeInfo(name: String, sqlType: Int, precision: Option[Int],
+  caseSensitive: Boolean, searchable: Short, unsignedAttribute: Boolean, numPrecRadix: Option[Int])
+
+/**
+ * GetTypeInfoOperation.
+ *
+ */
+class GetTypeInfoOperation(sessionHandle: SessionHandle)
```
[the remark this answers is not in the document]
mgaido91: not sure bout your question:
 - If the question is: what do we need them for? They are used when answering to metadata calls by the JDBC driver;
 - If the question is: can't we reuse Hive's one? No, because Hive's classes use `HiveSession` and `HiveConf` and we need to get rid of them.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/auth/AuthBridgeServer.scala
```diff
@@ -0,0 +1,296 @@
...
+      pc.setPassword(password)
+    }
+  }
+}
+
+/**
+ * Processor that pulls the SaslServer object out of the transport, and assumes the remote user's
+ * UGI before calling through to the original processor.
+ *
+ * This is used on the server side to set the UGI for each specific call.
```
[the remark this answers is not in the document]
mgaido91: Most of the operations here are needed as this is setting the `remoteUser` which is used in many places later. But the UGI impersonation may not be needed indeed.

mgaido91:
thanks for the review @vanzin. I have updated the PR description. Let me know if it still needs to be improved.

>  I think as a first step getting the Hive code in, with the minimal amount of changes necessary, is better.

I am not sure, as the only difference from that was moving from Java to Scala, which saves many conversion back and forth of collections and hence adds unneeded code only to handle this.

> overall take another look at whether the things you're forking really need to be forked

I am pretty sure that nothing which is there is unneeded.

mgaido91:
@jerryshao @vanzin anymore comments about this?

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyCLIService.scala
```diff
@@ -22,60 +22,57 @@ import java.util
...
-import org.apache.hadoop.hive.ql.parse.ParseUtils
-import org.apache.hadoop.hive.shims.Utils
-import org.apache.hadoop.security.UserGroupInformation
-import org.apache.hive.service.{CompositeService, ServiceException}
-import org.apache.hive.service.auth.HiveAuthFactory
-import org.apache.hive.service.cli._
-import org.apache.hive.service.cli.operation.Operation
+import org.apache.hadoop.security.{SecurityUtil, UserGroupInformation}
+import org.apache.hive.service.ServiceException
+import org.apache.hive.service.cli.{FetchOrientation, FetchType, GetInfoType, GetInfoValue, HiveSQLException, OperationHandle, SessionHandle}
```
vanzin: Just use a wildcard for long import lists...

server/src/main/scala/org/apache/livy/LivyConf.scala
```diff
@@ -98,10 +98,78 @@ object LivyConf {
...
 
+  // Thrift configurations
   val THRIFT_SERVER_ENABLED = Entry("livy.server.thrift.enabled", false)
   val THRIFT_INCR_COLLECT_ENABLED = Entry("livy.server.thrift.incrementalCollect", false)
   val THRIFT_SESSION_CREATION_TIMEOUT = Entry("livy.server.thrift.session.creationTimeout", "10m")
   val THRIFT_SERVER_JAR_LOCATION = Entry("livy.server.thrift.jarLocation", null)
+  // The following configs are the same present in Hive
+  val THRIFT_RESULTSET_DEFAULT_FETCH_SIZE =
+    Entry("livy.server.thrift.resultset.default.fetch.size", 1000)
+  val THRIFT_SPNEGO_PRINCIPAL = Entry("livy.server.thrift.authentication.spnego.principal", "")
```
vanzin: Won't this be exactly the same as `AUTH_KERBEROS_PRINCIPAL`?

The spnego principal must be `HTTP`, so it's not like they can be different.

Which also means the keytab config will have to match, too.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftServer.scala
```diff
@@ -114,24 +98,56 @@ object LivyThriftServer extends Logging {
...
+      }
+    }
+    val thriftCLIService = if (LivyThriftServer.isHTTPTransportMode(livyConf)) {
+      new ThriftHttpCLIService(cliService, oomHook)
+    } else {
+      new ThriftBinaryCLIService(cliService, oomHook)
+    }
+    addService(thriftCLIService)
+    super.init(livyConf)
+    Runtime.getRuntime.addShutdownHook(new Thread("LivyThriftServer Shutdown") {
```
vanzin: Is this necessary? Seems like if needed, it's something that the main server class should handle (and then shut down all loaded plugins).

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/auth/AuthBridgeServer.scala
```diff
@@ -0,0 +1,301 @@
...
+import org.apache.thrift.protocol.TProtocol
+import org.apache.thrift.transport.{TSaslServerTransport, TSocket, TTransport, TTransportException, TTransportFactory}
+
+import org.apache.livy.Logging
+
+/**
+ * The class is taken from Hive's `HadoopThriftAuthBridge.Server`. It bridges Thrift's SASL
+ * transports to Hadoop's SASL callback handlers and authentication classes.
+ *
+ * This class is based on Hive's one.
```
vanzin: Redundant comment.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/auth/AuthFactory.scala
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  @throws[LoginException]
+  def getAuthTransFactory: TTransportFactory = {
+    val isAuthKerberos = authTypeStr.equalsIgnoreCase(AuthTypes.KERBEROS.getAuthName)
+    val isAuthNoSASL = authTypeStr.equalsIgnoreCase(AuthTypes.NOSASL.getAuthName)
+    // TODO: add LDAP and PAM when supported
+    val isAuthOther = authTypeStr.equalsIgnoreCase(AuthTypes.NONE.getAuthName) ||
+      authTypeStr.equalsIgnoreCase(AuthTypes.CUSTOM.getAuthName)
+
```
vanzin: too many empty lines

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/auth/PlainSaslServer.scala
```diff
@@ -0,0 +1,184 @@
...
+import org.apache.thrift.transport.TSaslServerTransport
+
+import org.apache.livy.LivyConf
+
+
+/**
+ * Sun JDK only provides a PLAIN client and no server. This class implements the Plain SASL server
+ * conforming to RFC #4616 (http://www.ietf.org/rfc/rfc4616.txt).
+ */
+class PlainSaslServer private[auth] (
```
vanzin: I'm almost suggesting keeping a Livy-specific class in the `org.apache.hive.service.auth` package, extending the Hive one and just exposing a public constructor.

(Similar for other classes you're copying for seemingly similar reasons.)

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/cli/ThriftHttpServlet.scala
```diff
@@ -0,0 +1,500 @@
...
+import org.apache.thrift.TProcessor
+import org.apache.thrift.protocol.TProtocolFactory
+import org.apache.thrift.server.TServlet
+import org.ietf.jgss.{GSSContext, GSSCredential, GSSException, GSSManager, Oid}
+
+import org.apache.livy.{LivyConf, Logging}
+import org.apache.livy.thriftserver.SessionInfo
+import org.apache.livy.thriftserver.auth.{AuthenticationProvider, AuthFactory}
+
+class ThriftHttpServlet(
```
vanzin: Comment about where this comes from? (Even if it's similar to the other classes above.)

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/GetCatalogsOperation.scala
```diff
@@ -0,0 +1,62 @@
...
+package org.apache.livy.thriftserver.operation
+
+import org.apache.hive.service.cli._
+
+import org.apache.livy.Logging
+import org.apache.livy.thriftserver.serde.ResultSet
+import org.apache.livy.thriftserver.types.{BasicDataType, Field, Schema}
+
+/**
+ * GetCatalogsOperation.
```
vanzin: Not a super useful doc.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/GetTableTypesOperation.scala
```diff
@@ -0,0 +1,63 @@
...
+package org.apache.livy.thriftserver.operation
+
+import org.apache.hive.service.cli._
+
+import org.apache.livy.Logging
+import org.apache.livy.thriftserver.serde.ResultSet
+import org.apache.livy.thriftserver.types.{BasicDataType, Field, Schema}
+
+/**
+ * GetTableTypesOperation.
```
vanzin: Same.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/GetTypeInfoOperation.scala
```diff
@@ -0,0 +1,130 @@
...
+
+import org.apache.livy.Logging
+import org.apache.livy.thriftserver.serde.ResultSet
+import org.apache.livy.thriftserver.types.{BasicDataType, Field, Schema}
+
+sealed case class TypeInfo(name: String, sqlType: Int, precision: Option[Int],
+  caseSensitive: Boolean, searchable: Short, unsignedAttribute: Boolean, numPrecRadix: Option[Int])
+
+/**
+ * GetTypeInfoOperation.
```
vanzin: Same.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/GetTypeInfoOperation.scala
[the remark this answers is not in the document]
vanzin: I missed the `HiveConf` dependency when looking at the Hive code. Well, sigh.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/auth/PlainSaslServer.scala
mgaido91: I think you answered yourself to this comment in https://github.com/apache/incubator-livy/pull/117#discussion_r237319439. In this case, the Hive classes implemented here (see PlainSaslHelper) depend on `ThriftCLIService` which depends on `HiveConf`. Moreover some of them are final, so it is not easy to override them.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftServer.scala
mgaido91: yes, I moved it in the main server class, thanks.

server/src/main/scala/org/apache/livy/LivyConf.scala
mgaido91: yes, I thought about that too but I was not sure, so I kept this. Let me remove it and replace it with `AUTH_KERBEROS_PRINCIPAL` then. Thanks.

mgaido91:
Thanks for the review @vanzin! I addressed your comments.

> I still think there's stuff to clean up here

Not sure what you mean here exactly, but I think that after this patch we may start working with the Hive community in order to make some things easier to be re-used. Eg. we cannot reuse all the auth classes because they depend on the `HiveConf` we may propose to the Hive community to have them accepting in the constructor the parameters they need instead of the `HiveConf`, so we can reuse them. This may help reducing significantly the code size here.

vanzin:
Merging to master.

PR #124, opened 2018-10-25
[LIVY-531][BUILD] Update spark-2.3 default version to 2.3.2

LantaoJin:

Due to spark 2.3.2 is highly recommended version to upgrade for all 2.3.x users which contains many stability bug fixes.

## How was this patch tested?

Existing UTs.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1031,13 +1031,13 @@
     <profile>
       <id>spark-2.3</id>
       <properties>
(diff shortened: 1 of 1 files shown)


LantaoJin:
cc @jerryshao

vanzin:
Livy doesn't ship with Spark, so in the end this doesn't make a whole lot of difference. But sure, let's test against the latest one at least.

Merging to master.

PR #127, opened 2018-10-30
LIVY-525 SessionServlet: Convert response body for error messages to json

IngoSchuster:
SessionServlet is supposed to deliver json responses. This is not the case in some error cases today. This pull requests converts the response bodies for NotFound, BadRequest and Forbidden to json
https://issues.apache.org/jira/browse/LIVY-525

## How was this patch tested?
manual tests since this is a trivial fix.


vanzin:
Could you put the bug in the PR title like generally done in other PRs? You could also make the title shorter by saying "body for error messages" instead of listing all error messages...

I'd prefer to define a proper type for this instead of hardcoding the `Map` construction everywhere. Could be as simple as a `case class Error(message: String)`.

IngoSchuster:
Ok, I'll submit an update - hopefully later this week.

vanzin:
Merging to master. I'll fix the title during merge.

PR #128, opened 2018-11-13
[LIVY-533] Use setJobGroup/cancelJobGroup to cancel bypass jobs

bjoernlohrmann:

This PR fixes [LIVY-533](https://issues.apache.org/jira/browse/LIVY-533).

The proposed fix invokes setJobGroup() in the worker thread before calling job code, and cancelJobGroup() when another thread tries to cancel job execution. synchronized blocks are employed to guard against race conditions.

Also, this PR removes some unused code from the JobWrapper class.

## How was this patch tested?

Manual as well as Livy test suite. No new tests were added, since the code is covered by existing tests.

### DIFF
--- a/rsc/src/main/java/org/apache/livy/rsc/driver/BypassJobWrapper.java
+++ b/rsc/src/main/java/org/apache/livy/rsc/driver/BypassJobWrapper.java
@@ -38,28 +38,49 @@ public class BypassJobWrapper extends JobWrapper<byte[]> {
 
   @Override
   public Void call() throws Exception {
-    state = JobHandle.State.STARTED;
+    // we ignore the return value here, because super.call() will detect cancellation
+    // and still needs to perform some cleanup
+    tryTransitionToState(JobHandle.State.STARTED);
+
     return super.call();
   }
 
   @Override
   protected synchronized void finished(byte[] result, Throwable error) {
-    if (error == null) {
+    if (error == null && tryTransitionToState(JobHandle.State.SUCCEEDED)) {
       this.result = result;
-      this.state = JobHandle.State.SUCCEEDED;
-    } else {
+    } else if (error != null && tryTransitionToState(JobHandle.State.FAILED)) {
       this.error = error;
-      this.state = JobHandle.State.FAILED;
     }
   }
 
   @Override
-  boolean cancel() {
-    if (super.cancel()) {
-      this.state = JobHandle.State.CANCELLED;
-      return true;
+  synchronized boolean cancel() {
+    return tryTransitionToState(JobHandle.State.CANCELLED) && super.cancel();
+  }
+
+  private synchronized boolean tryTransitionToState(JobHandle.State newState) {
+    boolean success = false;
+
+    switch (this.state) {
+      case QUEUED:
+        if (newState == JobHandle.State.STARTED || newState == JobHandle.State.CANCELLED) {
+          this.state = newState;
+          success = true;
+        }
+        break;
+      case STARTED:
+        if (newState == JobHandle.State.CANCELLED || newState == JobHandle.State.SUCCEEDED  ||
+            newState == JobHandle.State.FAILED) {
+          this.state = newState;
+          success = true;
+        }
+        break;
+      default:
+        break;
     }
-    return false;
+
+    return success;
   }
 
   @Override
@@ -71,5 +92,4 @@ public class BypassJobWrapper extends JobWrapper<byte[]> {
     String stackTrace = error != null ? Utils.stackTraceAsString(error) : null;
     return new BypassJobStatus(state, result, stackTrace);
   }
-
 }
--- a/rsc/src/main/java/org/apache/livy/rsc/driver/JobWrapper.java
+++ b/rsc/src/main/java/org/apache/livy/rsc/driver/JobWrapper.java
@@ -17,15 +17,12 @@
 
 package org.apache.livy.rsc.driver;
 
-import java.util.ArrayList;
-import java.util.List;
 import java.util.concurrent.Callable;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
-import java.util.concurrent.atomic.AtomicInteger;
 
-import org.apache.spark.api.java.JavaFutureAction;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -38,8 +35,10 @@ public class JobWrapper<T> implements Callable<Void> {
   public final String jobId;
 
   private final RSCDriver driver;
+
   private final Job<T> job;
-  private final AtomicInteger completed;
+
+  private boolean isCancelled = false;
 
   private Future<?> future;
 
@@ -47,12 +46,20 @@ public class JobWrapper<T> implements Callable<Void> {
     this.driver = driver;
     this.jobId = jobId;
     this.job = job;
-    this.completed = new AtomicInteger();
   }
 
   @Override
   public Void call() throws Exception {
     try {
+      // this is synchronized to avoid races with cancel()
+      synchronized (this) {
+        if (isCancelled) {
+          throw new CancellationException("Job has been cancelled");
+        } else {
+          driver.jobContext().sc().setJobGroup(jobId, "", true);
+        }
+      }
+
       jobStarted();
       T result = job.call(driver.jobContext());
       finished(result, null);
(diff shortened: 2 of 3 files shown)


rsc/src/main/java/org/apache/livy/rsc/driver/JobWrapper.java
```diff
@@ -39,20 +39,27 @@
...
     this.driver = driver;
     this.jobId = jobId;
     this.job = job;
-    this.completed = new AtomicInteger();
   }
 
   @Override
   public Void call() throws Exception {
     try {
+      // this is synchronized to avoid races with cancel()
```
vanzin: If this race is possible, then there's another race you're missing: if `cancel()` is called before `call()` runs, then you'll start the job after it's been cancelled (since `future` will be `null`).

bjoernlohrmann: `future` will only be null for synchronous bypass jobs, but those cannot be cancelled unless I am mistaken?

For async bypass jobs, I think `future` should always be non-null during `call()/cancel()`, because it is assigned in `submit()` which also has the synchronized modifier and always runs before `call()` or `cancel()`.

I can add a boolean cancelled flag if you prefer, but this feels like duplicating information that is already available in the future.

vanzin: `future` will be null before `submit` is called. What if `cancel` is called before `submit` is called?

That's my question.

bjoernlohrmann: Good point. I just checked in `RSCDriver.handle(ChannelHandlerContext, BypassJobRequest)` and it looks like this can actually happen. If it does, cancellation won't work properly (as you said).

I updated the PR with a boolean flag `isCancelled` that addresses the issue.

rsc/src/main/java/org/apache/livy/rsc/driver/JobWrapper.java
```diff
@@ -38,21 +34,31 @@
...
+      synchronized (this) {
+        if (isCancelled) {
+          throw new Exception("Job isCancelled");
```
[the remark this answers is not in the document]
bjoernlohrmann: Done.

Sorry that this took so long, but I also added two tests for bypass job cancellation, which revealed problems in the state handling of BypassJobWrapper. The state would end up as FAILED after cancellation (due to CancellationException being thrown). Also, state transitions like ANCELLED -> STARTED -> FAILED were possible. The latest update to this PR should fix that.

It would be great if this fix could still make it into the 0.6 release.

rsc/src/main/java/org/apache/livy/rsc/driver/BypassJobWrapper.java
```diff
@@ -38,28 +38,50 @@ public BypassJobWrapper(RSCDriver driver, String jobId, Job<byte[]> serializedJo
 
   @Override
   public Void call() throws Exception {
-    state = JobHandle.State.STARTED;
+    // we ignore the return value here, because super.call() will also detect cancellation
+    // and needs to perform some cleanup
+    tryTransitionToState(JobHandle.State.STARTED);
```
vanzin: There's a race in between this call and the call to `super.call` though. You may transition to started and the job is cancelled before `super.call()` runs. Although as far as I can see that won't cause problems.

rsc/src/main/java/org/apache/livy/rsc/driver/BypassJobWrapper.java
```diff
@@ -38,28 +38,50 @@ public BypassJobWrapper(RSCDriver driver, String jobId, Job<byte[]> serializedJo
...
+    switch (this.state) {
+      case QUEUED:
+        if (newState == JobHandle.State.STARTED || newState == JobHandle.State.CANCELLED) {
+          this.state = newState;
+          success = true;
+        }
+        break;
+      case STARTED:
+        if (newState == JobHandle.State.CANCELLED
+                || newState == JobHandle.State.SUCCEEDED
```
vanzin: nit: indented too far

rsc/src/test/java/org/apache/livy/rsc/TestSparkClient.java
```diff
@@ -490,20 +487,72 @@ public void call(LivyClient client) throws Exception {
...
-          assertTrue(ee.getCause() instanceof RpcException);
-          assertTrue(ee.getCause().getMessage().contains(
-            "java.util.NoSuchElementException: " + jobId));
+        assertJobIdUntracked(lclient, jobId);
+      }
+    });
+  }
+
+  private void assertJobIdUntracked(RSCClient lclient, String jobId) throws InterruptedException,
+          TimeoutException {
```
vanzin: nit: indented too far. I also slightly prefer to keep all the thrown exceptions in the same line.

rsc/src/test/java/org/apache/livy/rsc/TestSparkClient.java
```diff
@@ -490,20 +487,72 @@ public void call(LivyClient client) throws Exception {
...
+        BypassJobStatus currStatus = lclient.getBypassJobStatus(jobId).get(TIMEOUT,
+                TimeUnit.SECONDS);
+        assertEquals(JobHandle.State.CANCELLED, currStatus.state);
+        assertJobIdUntracked(lclient, jobId);
       }
     });
   }
 
+  private BypassJobStatus assertNotCancelledOrFailed(RSCClient lclient, String jobId)
+          throws InterruptedException, ExecutionException, TimeoutException {
```
vanzin: nit: indented too far

rsc/src/test/java/org/apache/livy/rsc/TestSparkClient.java
```diff
@@ -490,20 +487,72 @@ public void call(LivyClient client) throws Exception {
...
+
+    BypassJobStatus currStatus = lclient.getBypassJobStatus(jobId).get(TIMEOUT,
+            TimeUnit.SECONDS);
```
vanzin: nit: too far. basically all your continuation lines are indented too far.

rsc/src/test/java/org/apache/livy/rsc/TestSparkClient.java
```diff
@@ -490,20 +487,72 @@ public void call(LivyClient client) throws Exception {
...
+  public void testBypassWithImmediateCancellation() throws Exception {
+    runBypassCancellationTest(0, 1000);
+  }
+
+  @Test
+  public void testBypassWithCancellation() throws Exception {
+    runBypassCancellationTest(200, 1000);
+  }
+
+  public void runBypassCancellationTest(long waitBeforeCancel, long jobDuration) throws Exception {
```
vanzin: I'm generally a bit wary of tests that rely on timing, especially when it's a little tight. Overloaded build machines may make this test flaky.

Can the job duration be much larger than the wait, or even basically "infinite"? The cancel or the shutting down of the client should interrupt the sleep, right?

rsc/src/main/java/org/apache/livy/rsc/driver/BypassJobWrapper.java
bjoernlohrmann: That is actually the standard case "job is cancelled while running" and should work as intended.

rsc/src/test/java/org/apache/livy/rsc/TestSparkClient.java
bjoernlohrmann: Fixed.

I was using IntelliJ IDEA defaults. Is there an "official" code style one can import?

rsc/src/test/java/org/apache/livy/rsc/TestSparkClient.java
bjoernlohrmann: Yes, cancel() interrupts the sleeper job.

I upped the job duration to one minute which I hope is enough. I am hesitant to make the job duration infinite. If cancellation fails due to some bug, the test will hang forever (or until some timeout kills the build) -- which is would be very annoying.

rsc/src/test/java/org/apache/livy/rsc/TestSparkClient.java
vanzin: Livy follows Spark's guide (http://spark.apache.org/contributing.html), maybe there's something there. But I don't use IntelliJ so can't help you there.

bjoernlohrmann: Thanks, good to know.

bjoernlohrmann:
I have updated some comment in the PR to force a rebuild. The last build I think failed due to an unrelated flaky testcase in the batch integration tests.

Is this otherwise good enough to be merged?

vanzin:
Merging to master.

I can see this not handling 100% of cases (e.g. people setting their own job group in their jobs), but at least it's a best effort.

PR #129, opened 2018-11-22
[LIVY-535] Fix non-atomic session creation

aromanenko-dev:

All steps, that include check for too many sessions and creating new
session, should be atomic operation to avoid concurrent access of
several threads to different part of this code in the same time.

https://issues.apache.org/jira/browse/LIVY-535

## How was this patch tested?
If was tested manually on local virtual cluster with this different values of config option `livy.server.session.max-creation`: 1, 2 and 3

Then I run the following command to make sure that only allowed number of sessions were created in the same time and others were rejected:
`for i in {1..5}; do curl -X POST --data '{"file": "/tmp/spark-examples-2.jar", "className": "org.apache.spark.examples.SparkPi"}' -H "Content-Type: application/json" http://localhost:8999/batches; done`

### DIFF
--- a/server/src/main/scala/org/apache/livy/server/SessionServlet.scala
+++ b/server/src/main/scala/org/apache/livy/server/SessionServlet.scala
@@ -127,16 +127,18 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
   }
 
   post("/") {
-    if (tooManySessions) {
-      BadRequest(ResponseMessage("Rejected, too many sessions are being created!"))
-    } else {
-      val session = sessionManager.register(createSession(request))
-      // Because it may take some time to establish the session, update the last activity
-      // time before returning the session info to the client.
-      session.recordActivity()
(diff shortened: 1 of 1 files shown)


vanzin:
Took a look at the surrounding code and this seems fine. I was worried that `createSession` might be blocking, but it's not.

Merging to master.

vanzin:
The code shouldn't be spending that much time in the synchronized block. It does do quite a lot of things, including spawning a child process, but it doesn't wait for that process to finish or anything like that. So it should be reasonably ok.

Otherwise this fix would be way more complicated. The accounting for how many things are starting is all over the place. If you want to spend time fixing that instead...

aromanenko-dev:
@Jassy1994 I tend to agree with @vanzin - it was a simple fix for a bug in current implementation. Also, I kept in mind that this code won't take too much time, so the waiting time should be quite acceptable. Please, let me know if it's not so in your case.

PR #131, opened 2018-12-03
[LIVY-538] Add integration tests for thriftserver

mgaido91:

Added a new module containing ITs for the thriftserver. The current tests contain only simple queries which ensure that the thriftserver is working correctly in the distributed environment. Specific checks of functionalities are already tested in the related UTs

## How was this patch tested?

added tests


mgaido91:
cc @vanzin

thriftserver/integration-test/src/test/scala/org/apache/livy/test/JdbcIT.scala
```diff
@@ -0,0 +1,66 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.livy.test
+
+import java.sql.Date
+
+import org.apache.livy.test.framework.BaseThriftIntegrationTestSuite
+
+class JdbcIT extends BaseThriftIntegrationTestSuite {
```
vanzin: Hmm... I think instead of having a separate module for this, we could put this test in a sub-directory of `integration-test/src/`, and then use `org.codehaus.mojo:build-helper-maven-plugin` to add that directory to the build when the thrift profile is enabled.

Similarly you could set a config to enable the thrift server in the cluster when the profile is enabled.

That should make this change a lot smaller.

mgaido91: thanks, I did what you suggested.

integration-test/src/thriftserver/main/scala/org/apache/livy/test/framework/ThriftJdbcClient.scala
```diff
@@ -0,0 +1,51 @@
...
+
+package org.apache.livy.test.framework
+
+import java.sql.{Connection, DriverManager, ResultSet}
+
+class ThriftJdbcClient(val jdbcUri: String) {
+
+  private def newConnection: Connection = {
+    val user = System.getProperty("user.name")
+    DriverManager.getConnection(jdbcUri, user, "")
```
vanzin: Do you need user + passwd or could you just call `getConnection(uri)`?

integration-test/src/thriftserver/main/scala/org/apache/livy/test/framework/ThriftJdbcClient.scala
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.livy.test.framework
+
+import java.sql.{Connection, DriverManager, ResultSet}
+
+class ThriftJdbcClient(val jdbcUri: String) {
+
+  private def newConnection: Connection = {
```
vanzin: Prefer to use `()` for methods that do things.

integration-test/src/thriftserver/main/scala/org/apache/livy/test/framework/ThriftJdbcClient.scala
```diff
@@ -0,0 +1,51 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.livy.test.framework
+
+import java.sql.{Connection, DriverManager, ResultSet}
+
+class ThriftJdbcClient(val jdbcUri: String) {
```
vanzin: Does this need to be a separate class? It seems that if you add these helper methods to `BaseThriftIntegrationTestSuite`, the tests would become a little bit cleaner by not having this indirection. There's no state here, so that should be fine.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftServer.scala
```diff
@@ -138,6 +139,16 @@ class LivyThriftServer(
     session.owner == user || accessManager.checkModifyPermissions(user)
   }
 
+  def getJdbcUrl: String = {
+    val additionalUrlParams = if (thriftCLIService.isInstanceOf[ThriftHttpCLIService]) {
```
vanzin: It seems there's a whole lot of plumbing just to get this URL to the test code.

Maybe the test code could figure this out by looking at the configuration directly?

vanzin: Can't you control the bind host via `THRIFT_BIND_HOST`?

It just seems like a lot of code for very little gain.

mgaido91: ok, let me try and do that.

server/src/main/scala/org/apache/livy/server/LivyServer.scala
```diff
@@ -355,6 +355,21 @@ class LivyServer extends Logging {
...
 
+  /** For ITs only */
+  def getJdbcUrl: Option[String] = {
+    _thriftServerFactory.map { _ =>
+      val additionalUrlParams = if (livyConf.get(THRIFT_TRANSPORT_MODE) == "http") {
+        "?hive.server2.transport.mode=http;hive.server2.thrift.http.path=cliservice"
+      } else {
+        ""
+      }
+      val host = Option(livyConf.get(THRIFT_BIND_HOST)).getOrElse(
```
vanzin: When you mentioned that you were worried about which host the server was binding too, I was expecting you to explicitly set this config to "localhost" in the tests, but you don't seem to be doing that.

Also, instead of this method in the main server code, you could have all this in `MiniLivyMain.baseLivyConf`, where you configure the LTS options. Use the same options you explicitly set in the configuration to build this url. No code necessary outside of test classes.

mgaido91: Yes, if none is set, it binds on all addresses, so it is not needed to restrict to localhost IMHO. I chose to do this, because it reflects what was done for the (REST) server URL. I think it is more clear this way.

vanzin: Ok, since the other URL is also in this class, looks fine.

vanzin:
Merging to master.

PR #133, opened 2018-12-14
Try to process statements of single session concurrentlly (especially for sql kind)

WeiWenda:
1.  interpreterExecutor of repl.Session change to newFixedThreadPool, size configurable
2. Try to cleanup _statements of Session by judge expired or not, rather than drop Eldest statement casually.
3. Reject statement request when buffer is full and none is expired
- The new Statechart Diagram of Statement looks like this: (brown box means inner sate, user unknowable)
livy statement2
- so that user can configure these properties to control a statement's lifecycle

| Property Name                           | Default | Meaning                                                      |
| --------------------------------------- | ------- | ------------------------------------------------------------ |
| livy.rsc.server.statements.parallel.num | 1      | Size of ExecutorThreadPool |
| livy.rsc.retained-statements            | 100     | Buffer size of statement |
| livy.rsc.result-discard.timeout         | 10m     | **Expired Timeout after read**，When the user retrieves the query result, the state change to Readed from Available, and automatically becomes Expired after 10 minutes. Expired means when the buffer queue space is insufficient, this statement may be moved out of the buffer queue. |
| livy.rsc.result-retained.timeout        | 1h      | **Expired timeout before read**，After statement became Available, If user never get the query result within 1 hour , StatementState will change to Expired automatically. |

## How was this patch tested?

we run a jmeter test of 20 threads each post statement 1/60s, set "spark.scheduler.mode": "FAIR", print debug message in AM stdout, and examine the buffer state.

### DIFF
--- a/repl/src/main/scala/org/apache/livy/repl/Session.scala
+++ b/repl/src/main/scala/org/apache/livy/repl/Session.scala
@@ -17,12 +17,10 @@
 
 package org.apache.livy.repl
 
-import java.util.{LinkedHashMap => JLinkedHashMap}
-import java.util.Map.Entry
-import java.util.concurrent.Executors
+import java.util.Date
+import java.util.concurrent.{Executors, TimeUnit}
 import java.util.concurrent.atomic.AtomicInteger
 
-import scala.collection.JavaConverters._
 import scala.collection.mutable
 import scala.concurrent.{ExecutionContext, Future}
 import scala.concurrent.duration._
@@ -70,11 +68,15 @@ class Session(
   // Number of statements kept in driver's memory
   private val numRetainedStatements = livyConf.getInt(RSCConf.Entry.RETAINED_STATEMENTS)
 
-  private val _statements = new JLinkedHashMap[Int, Statement] {
-    protected override def removeEldestEntry(eldest: Entry[Int, Statement]): Boolean = {
-      size() > numRetainedStatements
-    }
-  }.asScala
+  private val resultRetainedTimeout = livyConf.getTimeAsMs(RSCConf.Entry.STAETMENT_RESULT_RETAINED_TIMEOUT)
+
+  private val resultDiscardTimeout = livyConf.getTimeAsMs(RSCConf.Entry.STATEMENT_RESULT_DISCRAD_TIMEOUT)
+
+  private val _statements = mutable.HashMap[Int, Statement]()
+
+  private var _waitingRemove = mutable.HashMap[Int, Long]()
+
+  private var LAT: (Int, Long) = (0, 0)
 
   private val newStatementId = new AtomicInteger(0)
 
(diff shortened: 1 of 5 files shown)


repl/src/main/scala/org/apache/livy/repl/Session.scala
```diff
@@ -356,4 +373,96 @@ class Session(
...
+
+  private def snapshot(id: Int, msg: String): Unit = {
+    debug(s"statement No.${id} ${msg}")
+    val (running, o) = _statements.values.partition(_.state.get() == StatementState.Running)
+    debug(s"Buffer Size: ${numRetainedStatements}\tUsed Size: ${_statements.size}\t" +
+      s"Finished: ${_waitingRemove.size} Running: ${running.size} " +
+      s"Waiting: ${o.size-_waitingRemove.size}")
+  }
+
+  private def isOverload(proposer: Int): Boolean = _statements.synchronized {
```
vanzin: What does `isOverload` check? Could you use some more descriptive names, and add comments explaining these methods?

repl/src/main/scala/org/apache/livy/repl/Session.scala
```diff
@@ -356,4 +373,96 @@ class Session(
...
+      }
+      false
+    } else {
+      snapshot(proposer, "is rejected")
+      true
+    }
+  }
+
+  // checkExpired should have _statements's lock
+  private def checkExpired: Boolean = {
```
vanzin: Methods that do things should have `()`. And this method is not just checking that something (what?) expired.

repl/src/main/scala/org/apache/livy/repl/Session.scala
```diff
@@ -356,4 +373,96 @@ class Session(
...
+      false
+    } else if (LAT._2 == 0) {
+      // only compute LAT when _waitingRemove is nonEmpty
+      cleanUpExpired
+      _statements.size < numRetainedStatements
+    } else {
+      new Date().getTime > LAT._2
+    }
+  }
+  // Attention: do not use _statements.synchronized to prevent deadlock
```
vanzin: This is scary, because this method is actually modifying `_statements`. This comment tells me the logic you have is not completely right.

WeiWenda:
@vanzin thanks for your review, I have corrected them one by one in recent commit. But continuous-integration failed because of Spark 2.4 ITs. I have no idea how to fix it, have you met this before?

> org.apache.maven.lifecycle.LifecycleExecutionException:
Failed to execute goal org.jacoco:jacoco-maven-plugin:0.8.2:report-aggregate (jacoco-report) on project livy-coverage-report: An error has occurred in JaCoCo Aggregate report generation.

WeiWenda:
oh, I just recognize what is “file a bug”. It’s too late today, I will do these job morning. thank you.

PR #134, opened 2018-12-15
[LIVY-543] Clean thriftserver dependencies from Hive's jetty

mgaido91:

Hive 3.0 uses Jetty 9.3.8. So there are some jars coming from hive dependency which are conflicting with our Jetty dependencies which were upgraded in LIVY-526. Hence we should ignore the dependencies Hive has on Jetty and use our own version. After the refactor which has been done in the thriftserver, this should be easy.

The PR removes also an extra-dependency which was left in the thriftserver-server module which is not needed anymore after the refactor which created the thriftserver-session module.

## How was this patch tested?

manual tests, checking the generated JARs

### DIFF
--- a/thriftserver/server/pom.xml
+++ b/thriftserver/server/pom.xml
@@ -50,6 +50,10 @@
           <groupId>org.apache.hive</groupId>
           <artifactId>*</artifactId>
         </exclusion>
+        <exclusion>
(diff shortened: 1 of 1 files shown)


mgaido91:
cc @arunmahadevan @jerryshao @vanzin

PR #135, opened 2018-12-21
[LIVY-544] Allow interpreterExecutor run in ThreadPool

WeiWenda:
make interpreterExecutor run with newFixedThreadPool  rather than newSingleThreadExecutor
https://issues.apache.org/jira/browse/LIVY-544

## How was this patch tested?
we run a jmeter test of 20 threads each post statement 1/60s, set "spark.scheduler.mode": "FAIR"
we find setJobGroup is not thread-safe causing some reflect Error, so wrap it by Session.synchronized
> this.synchronized{ setJobGroup(tpe, statementId) }


Tagar:
@WeiWenda thanks for changing to `concurrentSQL` .

It seems all tests fail with 

> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Is this a known issue? cc @vanzin

WeiWenda:
> @WeiWenda thanks for changing to `concurrentSQL` .
> 
> It seems all tests fail with
> 
> > No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
> 
> Is this a known issue? cc @vanzin

I have fixed all errors. Can you merge this pr. Thank you.

Tagar:
@WeiWenda I am not a committer. 

Thanks for this great improvement. It would be awesome to have concurrentSQL
option in Livy, like we  have in bare Zeppelin. 

cc @mgaido91 @jerryshao @vanzin @ajbozarth @zjffdu - please help to review. 

Thank you.

ajbozarth:
I see no issues with the code, but since I don't have any experience with the SQL side of Livy I will leave final review and merging to a more knowledgable committer

core/src/main/scala/org/apache/livy/sessions/Kind.scala
```diff
@@ -35,6 +35,8 @@ object Shared extends Kind("shared")
 
 object SQL extends Kind("sql")
 
+object ConcurrentSQL extends Kind("concurrentSQL")
```
mgaido91: I don't think that it is a good idea to add a new session kind. I think we should rather have a configuration for enabling FAIR vs FIFO and I am wondering if this shouldn't be done cross kind, ie. for all the kinds: why was it done only for SQL?

Tagar: @mgaido91 good point. There was some discussion on this at https://issues.apache.org/jira/browse/LIVY-544
For code paragraphs chances are that there is a dependency between code items, and they can't run in parallel. For SQL sometimes there is no dependency often when it's just a SELECT.. but we let end-users decide if there is truly no dependency and let queries execute in parallel only when it was requested explicitly. Also this was modeled after Zeppelin to some degree - Zeppelin only allows parallel execution for SQL cells and not code cells like pyspark.

WeiWenda: Run code block in parallel is very confusing. Spark‘s FAIR scheduler mode meaning that all runable stages can receive executors right away no matter when it was submitted. So that concurrentSQL’s stages need run into FAIR pool. But for scala or python or java, code block can't  make stages in most time, it make no sense to let them run in FAIR pool.

WeiWenda: I review 'SparkSqlInterperter' in zeppelin, it do the same thing as I have done.
### NewSparkInterpreter
```
if (entry.getKey().toString().equals("zeppelin.spark.concurrentSQL")
            && entry.getValue().toString().equals("true")) {
          conf.set("spark.scheduler.mode", "FAIR");
        }
```
### SparkSqlInterpreter
```
public InterpreterResult interpret(String st, InterpreterContext context)
      throws InterpreterException {
   ...
    sc.setLocalProperty("spark.scheduler.pool", context.getLocalProperties().get("pool"));
```
#### InterpreterContext's localProperties come from RemoteInterpreterContext's localProperties, which is RemoteInterpreter's InterpreterContext, created by Paragrah's getInterpreterContext, configured by %spark.sql(pool=poolname,key=value)
```
  private static Pattern REPL_PATTERN =
      Pattern.compile("(\\s*)%([\\w\\.]+)(\\(.*?\\))?.*", Pattern.DOTALL);

```

WeiWenda: But different from zeppelin, I don't think user can provide spark's fairscheduler.xml in proper way especially in cluster deploy, and can keep in mind which sql should run into which pool. This need a great understand of spark scheduler. So I do all things in livy to release user from configuring above items. Of course we can modify to 
```
{"code":"", "type":"sql", "pool":"xx"}
//or
{"code":"set pool=xx, query content", "type":"sql"}
//or
{"code":"", "type":"sql?pool=xx"}
```
I have to say that expansibility of livy is not as excellent as zeppelin, there should be a properties part in statemnet boby.
image

Tagar:
Would somebody else from committers be available to review this?
cc @jerryshao @vanzin @zjffdu 

my 2 cents -
It's one of the most exciting Livy improvements among currently open PRs.
Since it's backward compatible and disabled by default, it seems fairly safe to add this in.

Thanks everyone.

jcdauchy:
Hello,

With this feature, would it mean that we can run several statement at the same time on one Livy session ? Regarding of course, that each statement is independant from the other.

Thanks.

WeiWenda:
> Hello,
> 
> With this feature, would it mean that we can run several statement at the same time on one Livy session ? Regarding of course, that each statement is independant from the other.
> 
> Thanks.

Yes, you can think so.

jcdauchy:
Hope to find some time to test. Is this the way you use this feature ?

mtNachiket:
@WeiWenda When will this be merged as its been almost year and half old ? As I have similar use case and this would solve my problem.

mgaido91:
this is not my main area of expertise, @vanzin and @jerryshao might be better reviewers

PR #136, opened 2018-12-21
[LIVY-545] Try to clear buffer by judge expired or not, rather than drop Eldest …

WeiWenda:
https://issues.apache.org/jira/browse/LIVY-545
1. Try to cleanup _statements of Session by judge expired or not, rather than drop Eldest statement casually.
2. Reject statement request when buffer is full and none is expired
- The new Statechart Diagram of Statement looks like this: (brown box means inner sate, user unknowable)
livy statement2
- so that user can configure these properties to control a statement's lifecycle

| Property Name                           | Default | Meaning                                                      |
| --------------------------------------- | ------- | ------------------------------------------------------------ |
| livy.rsc.retained-statements            | 100     | Buffer size of statement |
| livy.rsc.result-discard.timeout         | 10m     | **Expired Timeout after read**，When the user retrieves the query result, the state change to Readed from Available, and automatically becomes Expired after 10 minutes. Expired means when the buffer queue space is insufficient, this statement may be moved out of the buffer queue. |
| livy.rsc.result-retained.timeout        | 1h      | **Expired timeout before read**，After statement became Available, If user never get the query result within 1 hour , StatementState will change to Expired automatically. |
## How was this patch tested?
we run a jmeter test of 20 threads each post statement 1/60s, set "spark.scheduler.mode": "FAIR", print debug message in AM stdout, and examine the buffer state.

we modify the "should remove expired statements when reaching threshold" of SessionSpec to fit this new character.

### DIFF
--- a/repl/src/main/scala/org/apache/livy/repl/ReplDriver.scala
+++ b/repl/src/main/scala/org/apache/livy/repl/ReplDriver.scala
@@ -75,7 +75,9 @@ class ReplDriver(conf: SparkConf, livyConf: RSCConf)
     } else {
       assert(msg.from != null)
       assert(msg.size != null)
-      if (msg.size == 1) {
+      if (msg.from == -1) {
+        Array(new Statement(-1, session.getBufferState(), StatementState.Rejected, null))
+      } else if (msg.size == 1) {
         session.statements.get(msg.from).toArray
       } else {
         val until = msg.from + msg.size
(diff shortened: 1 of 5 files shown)


alex-the-man:
Would you mind describing how you are using interactive sessions?

WeiWenda:
> Would you mind describing how you are using interactive sessions?

we use livy to provide spark sql restful server as a substitute of hive-thriftserver. Such problem only happen when too many user connect to the same livy session. But as now, we find it maybe unrealistic, as spark FAIR schedule mode's performance is so pool, can only deal with 20 around query statement in the same time.

mgaido91:
@WeiWenda we just introduced a thrift endpoint in Livy (see https://issues.apache.org/jira/browse/LIVY-489). Do you think that would help your use case?

WeiWenda:
> @WeiWenda we just introduced a thrift endpoint in Livy (see https://issues.apache.org/jira/browse/LIVY-489). Do you think that would help your use case?

Thank you! It's much better than raw STS, I'm glad to use this feature of livy.

PR #138, opened 2019-01-14
[LIVY-547]do not check expire when session in state busy

Jassy1994:
add a logic in check expired function to ensure interactiveSession will not expired in state busy。 

## How was this patch tested?
set livy.server.session.timeout to a short interval and submit a job by interactiveSession
this session will not be recycled in state busy。

### DIFF
--- a/server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
+++ b/server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
@@ -142,6 +142,8 @@ class SessionManager[S <: Session, R <: RecoveryMetadata : ClassTag](
             false
           } else if (session.isInstanceOf[BatchSession]) {
             false
+          } else if (SessionState.Busy.equals(session.state)) {
+            false
           } else {
             val currentTime = System.nanoTime()
             currentTime - session.lastActivity > sessionTimeout


meisam:
Please make this a configurable option.

Jassy1994:
> Please make this a configurable option.

Hi meisam
    I don't think an interactiveSession in state busy should be expired.so I think its no need to set this into a configuration. or may you share me some counter-cases, Thanks a lot.

meisam:
> or may you share me some counter-cases
If the number of jobs running are small and can be monitored manually, it makes sense to keep busy jobs.
But in a system with hundreds of users and thousands of jobs running each day, it is impractical to check the status of each job manually. In many cases users submit jobs on a daily/hourly basis though an enterprise scheduler. In this case, if a job takes too long to run, it is better to fail it and figure out why it took so long.
Also, users have the option to reset the last activity time for a job though `sessions/:id/connect` endpoint.

mgaido91:
This would be a behavior change, so at very least I think we should add a config option and documentation for it. Anyway, it depends on what you consider as "Inactive".  In the current behavior, "Inactive" refers to the session abstraction in Livy server. So if a user doesn't access it, that object is inactive. If you consider the Spark job, instead, the right behavior is the one described in this PR.

I agree the current behavior is confusing (I got confused myself at the beginning) and despite the workaround is easy (there is a REST endpoint for avoiding the timeout...), it would be good to clarify and document it better. So, to summarize, I think we should first document properly the current status rather than doing a behavior change as proposed in this PR.
cc @vanzin too for his opinion on this.

vanzin:
I don't remember exactly how the clients monitor interactive sessions, but doesn't a client poll for the result of a statement? Doesn't that count as updating the session for activity?

Although I guess that would still not account for the "fire a job and go out for a few days" use case.

It sounds like a separate config is the only way to cover all user cases (e.g. a "busy timeout", which by default maintains the current behavior, and can be set to stop the session if it's been busy for a certain period).

alex-the-man:
I agree the current GC behaviour is not well documented. As a first step I will document them in LivyConf.scala.

@Jassy1994 Would you mind sharing how are you using interactive sessions? Are you using it directly without a notebook? Thanks :).

Jassy1994:
> I agree the current GC behaviour is not well documented. As a first step I will document them in LivyConf.scala.
> 
> @Jassy1994 Would you mind sharing how are you using interactive sessions? Are you using it directly without a notebook? Thanks :).

this issue both happened in jupyter notebook and livy rest api when user submit a long time spent job.

shanyu:
My apologies, I created another pull request (https://github.com/apache/incubator-livy/pull/146) and I didn't realize there is this pull request. I'd like to revive this discussion.

I think the primary use case for Livy interactive session is for notebooks (Jupyter/Zeppelin). Data Scientists type code in notebook and view results. Scheduled app submission should use Livy batches instead of sessions.

Livy server garbage collect sessions without "activity" for a period specified by "livy.server.session.timeout", where "activity" is defined as: a) submit statement; b) cancel statement; c) complete statement; d) access REST endpoint sessions/:id/connect;

The strange thing about config "livy.server.session.timeout" is that it simultaneously controls two different things: 1) timeout for idle sessions (timeout since last statement completion time); and 2) max run time for a statement (timeout since last statement submission);

It will eliminate confusion if we only let "livy.server.session.timeout" control timeout for idle sessions, as its name properly suggests. Whether or not we want to provide "livy.server.session.statement.maxruntime" is a different discussion. Note that I'm not aware of a configuration to limit batch job runtime.

@meisam for the scenario you described (thousands of jobs each day by enterprise scheduler), should you use Livy batch job instead of interactive sessions?

jerryshao:
This is fixed by #190 , thanks a lot for the contribution.

PR #140, opened 2019-01-18
[LIVY-550] quick failure when error happens in spark-submit process

Jassy1994:
support batch session a quick failure when error happens in spark-submit process

## How was this patch tested?
submit a batch session with wrong username or queuename，observe session state in livy ui


Jassy1994:
> Without this fix, the app state should still transit to FAILED after 2 minutes with default configuration. Is this fix intended to surface the error faster?

I am sorry to have a late reply, yes, app state will transit to FAILED in livy.server.yarn.app-leakage.check-timeout, in my situation, some users submit wrong params but they dont realize, and in livy ui, their session is starting so we can not correct in time.so I think a quick failure is more friendly to users.

PR #141, opened 2019-01-18
[LIVY-551] Add "doAs" impersonation support

risdenk:

Currently `proxyuser` is used in batches and sessions for impersonation. This should be extended to support impersonation for all endpoints. Adding the query parameter `doAs` matches others in the Hadoop ecosystem and will work across endpoints.

https://issues.apache.org/jira/browse/LIVY-551

## How was this patch tested?

Added unit and integration tests for this feature. Manually tested to ensure that `proxyuser` still works. Also checked that if both `proxyuser` and the new `doAs` parameter are specified that `doAs` takes precedence.

### DIFF
--- a/server/src/main/scala/org/apache/livy/server/AccessManager.scala
+++ b/server/src/main/scala/org/apache/livy/server/AccessManager.scala
@@ -101,13 +101,13 @@ private[livy] class AccessManager(conf: LivyConf) extends Logging {
   /**
    * Checks that the request user can impersonate the target user.
    * If impersonation is enabled and the user does not have permission to impersonate
-   * then throws an `AccessControlException`. If impersonation is disabled returns false
+   * then throws an `AccessControlException`. If impersonation is disabled returns false.
    */
   def checkImpersonation(
       requestUser: String,
       impersonatedUser: String): Boolean = {
     if (conf.getBoolean(LivyConf.IMPERSONATION_ENABLED)) {
-      if (hasSuperAccess(requestUser, impersonatedUser)) {
+      if (hasSuperAccess(impersonatedUser, requestUser)) {
         true
       } else {
         throw new AccessControlException(
@@ -119,10 +119,10 @@ private[livy] class AccessManager(conf: LivyConf) extends Logging {
   }
 
   /**
-   * Check that the request user has is able to impersonate the given user.
+   * Check that the requesting user has admin access to resources owned by the given target user.
    */
-  def hasSuperAccess(requestUser: String, impersonatedUser: String): Boolean = {
-    requestUser == impersonatedUser || checkSuperUser(requestUser)
+  def hasSuperAccess(target: String, requestUser: String): Boolean = {
+    requestUser == target || checkSuperUser(requestUser)
   }
 
   /**
@@ -133,7 +133,7 @@ private[livy] class AccessManager(conf: LivyConf) extends Logging {
   }
 
   /**
-   * Check that the request user has view access to the given session
+   * Check that the request user has view access to the given session.
    */
   def hasViewAccess(session: Session, effectiveUser: String): Boolean = {
     session.owner == effectiveUser || checkViewPermissions(effectiveUser)
--- a/server/src/main/scala/org/apache/livy/server/AccessManager.scala
+++ b/server/src/main/scala/org/apache/livy/server/AccessManager.scala
@@ -119,6 +119,18 @@ private[livy] class AccessManager(conf: LivyConf) extends Logging {
   }
 
   /**
+   * Returns the impersonated user or None after checking if impersonation is allowed
+   */
+  def enforceProxyUser(owner: String,
+      proxyUser: Option[String]): Option[String] = {
+    if(conf.getBoolean(LivyConf.IMPERSONATION_ENABLED)) {
+      proxyUser.filter(checkImpersonation(owner, _)).orElse(Option(owner))
+    } else {
+      None
+    }
+  }
+
+  /**
    * Check that the requesting user has admin access to resources owned by the given target user.
    */
   def hasSuperAccess(target: String, requestUser: String): Boolean = {
--- a/server/src/main/scala/org/apache/livy/server/AccessManager.scala
+++ b/server/src/main/scala/org/apache/livy/server/AccessManager.scala
@@ -20,7 +20,6 @@ package org.apache.livy.server
 import java.security.AccessControlException
 
 import org.apache.livy.{LivyConf, Logging}
-import org.apache.livy.sessions.Session
 
 private[livy] class AccessManager(conf: LivyConf) extends Logging {
   private val aclsOn = conf.getBoolean(LivyConf.ACCESS_CONTROL_ENABLED)
@@ -99,32 +98,22 @@ private[livy] class AccessManager(conf: LivyConf) extends Logging {
   def isAccessControlOn: Boolean = aclsOn
 
   /**
-   * Checks that the request user can impersonate the target user.
-   * If impersonation is enabled and the user does not have permission to impersonate
-   * then throws an `AccessControlException`. If impersonation is disabled returns false.
+   * Checks that the requesting user can impersonate the target user.
+   * If the user does not have permission to impersonate, then throws an `AccessControlException`.
+   *
+   * @return The user that should be impersonated. That can be the target user if defined, the
+   *         request's user - which may not be defined - otherwise, or `None` if impersonation is
+   *         disabled.
    */
   def checkImpersonation(
-      requestUser: String,
-      impersonatedUser: String): Boolean = {
+      target: Option[String],
+      requestUser: String): Option[String] = {
     if (conf.getBoolean(LivyConf.IMPERSONATION_ENABLED)) {
-      if (hasSuperAccess(impersonatedUser, requestUser)) {
-        true
-      } else {
+      if (!target.forall(hasSuperAccess(_, requestUser))) {
         throw new AccessControlException(
-          s"User '$requestUser' not allowed to impersonate '$impersonatedUser'.")
+          s"User '$requestUser' not allowed to impersonate '$target'.")
       }
-    } else {
-      false
-    }
-  }
-
-  /**
-   * Returns the impersonated user or None after checking if impersonation is allowed
-   */
-  def enforceProxyUser(owner: String,
-      proxyUser: Option[String]): Option[String] = {
-    if (conf.getBoolean(LivyConf.IMPERSONATION_ENABLED)) {
-      proxyUser.filter(checkImpersonation(owner, _)).orElse(Option(owner))
+      target.orElse(Option(requestUser))
     } else {
       None
     }
@@ -138,16 +127,16 @@ private[livy] class AccessManager(conf: LivyConf) extends Logging {
   }
 
   /**
-   * Check that the request user has modify access to the given session.
+   * Check that the request's user has modify access to resources owned by the given target user.
    */
-  def hasModifyAccess(session: Session, effectiveUser: String): Boolean = {
-    session.owner == effectiveUser || checkModifyPermissions(effectiveUser)
+  def hasModifyAccess(target: String, requestUser: String): Boolean = {
+    requestUser == target || checkModifyPermissions(requestUser)
   }
 
   /**
-   * Check that the request user has view access to the given session.
+   * Check that the request's user has view access to resources owned by the given target user.
    */
-  def hasViewAccess(session: Session, effectiveUser: String): Boolean = {
-    session.owner == effectiveUser || checkViewPermissions(effectiveUser)
+  def hasViewAccess(target: String, requestUser: String): Boolean = {
+    requestUser == target || checkViewPermissions(requestUser)
   }
 }
--- a/server/src/main/scala/org/apache/livy/server/AccessManager.scala
+++ b/server/src/main/scala/org/apache/livy/server/AccessManager.scala
@@ -20,6 +20,7 @@ package org.apache.livy.server
 import java.security.AccessControlException
 
 import org.apache.livy.{LivyConf, Logging}
+import org.apache.livy.sessions.Session
 
 private[livy] class AccessManager(conf: LivyConf) extends Logging {
   private val aclsOn = conf.getBoolean(LivyConf.ACCESS_CONTROL_ENABLED)
@@ -98,46 +99,43 @@ private[livy] class AccessManager(conf: LivyConf) extends Logging {
   def isAccessControlOn: Boolean = aclsOn
 
   /**
-   * Checks that the requesting user can impersonate the target user.
-   * If the user does not have permission to impersonate, then throws an `AccessControlException`.
-   *
-   * @return The user that should be impersonated. That can be the target user if defined, the
-   *         request's user - which may not be defined - otherwise, or `None` if impersonation is
-   *         disabled.
+   * Checks that the request user can impersonate the target user.
+   * If impersonation is enabled and the user does not have permission to impersonate
+   * then throws an `AccessControlException`. If impersonation is disabled returns false
    */
   def checkImpersonation(
-      target: Option[String],
       requestUser: String,
-      livyConf: LivyConf): Option[String] = {
-    if (livyConf.getBoolean(LivyConf.IMPERSONATION_ENABLED)) {
-      if (!target.forall(hasSuperAccess(_, requestUser))) {
+      impersonatedUser: String): Boolean = {
+    if (conf.getBoolean(LivyConf.IMPERSONATION_ENABLED)) {
+      if (hasSuperAccess(requestUser, impersonatedUser)) {
+        true
+      } else {
         throw new AccessControlException(
-          s"User '$requestUser' not allowed to impersonate '$target'.")
+          s"User '$requestUser' not allowed to impersonate '$impersonatedUser'.")
       }
-      target.orElse(Option(requestUser))
     } else {
-      None
+      false
     }
   }
 
   /**
-   * Check that the requesting user has admin access to resources owned by the given target user.
+   * Check that the request user has is able to impersonate the given user.
    */
-  def hasSuperAccess(target: String, requestUser: String): Boolean = {
-    requestUser == target || checkSuperUser(requestUser)
+  def hasSuperAccess(requestUser: String, impersonatedUser: String): Boolean = {
+    requestUser == impersonatedUser || checkSuperUser(requestUser)
   }
 
   /**
-   * Check that the request's user has modify access to resources owned by the given target user.
+   * Check that the request user has modify access to the given session.
    */
-  def hasModifyAccess(target: String, requestUser: String): Boolean = {
-    requestUser == target || checkModifyPermissions(requestUser)
+  def hasModifyAccess(session: Session, effectiveUser: String): Boolean = {
+    session.owner == effectiveUser || checkModifyPermissions(effectiveUser)
   }
 
   /**
-   * Check that the request's user has view access to resources owned by the given target user.
+   * Check that the request user has view access to the given session
    */
-  def hasViewAccess(target: String, requestUser: String): Boolean = {
-    requestUser == target || checkViewPermissions(requestUser)
+  def hasViewAccess(session: Session, effectiveUser: String): Boolean = {
+    session.owner == effectiveUser || checkViewPermissions(effectiveUser)
   }
 }
--- a/server/src/main/scala/org/apache/livy/server/AccessManager.scala
+++ b/server/src/main/scala/org/apache/livy/server/AccessManager.scala
@@ -123,7 +123,7 @@ private[livy] class AccessManager(conf: LivyConf) extends Logging {
    */
   def enforceProxyUser(owner: String,
       proxyUser: Option[String]): Option[String] = {
-    if(conf.getBoolean(LivyConf.IMPERSONATION_ENABLED)) {
+    if (conf.getBoolean(LivyConf.IMPERSONATION_ENABLED)) {
       proxyUser.filter(checkImpersonation(owner, _)).orElse(Option(owner))
     } else {
       None
--- a/server/src/main/scala/org/apache/livy/server/SessionServlet.scala
+++ b/server/src/main/scala/org/apache/livy/server/SessionServlet.scala
@@ -164,8 +164,7 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
    * Returns the impersonated user as given by "doAs" as a request parameter.
    */
   protected def impersonatedUser(request: HttpServletRequest): Option[String] = {
-    val impersonatedUser = Option(request.getParameter("doAs"))
-    impersonatedUser.filter(accessManager.checkImpersonation(remoteUser(request), _))
+    Option(request.getParameter("doAs"))
   }
 
   /**
@@ -173,22 +172,16 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
    */
   protected def proxyUser(request: HttpServletRequest,
       createRequestProxyUser: Option[String]): Option[String] = {
-    impersonatedUser(request).orElse(legacyProxyUser(remoteUser(request), createRequestProxyUser))
-  }
-
-  /**
-   * Returns the proxy user as determined by the json request body or owner if available.
-   * This is necessary to preserve backwards compatibility prior to "doAs" impersonation.
-   */
-  protected def legacyProxyUser(owner: String, bodyProxyUser: Option[String]): Option[String] = {
-    bodyProxyUser.filter(accessManager.checkImpersonation(owner, _)).orElse(Option(owner))
+    impersonatedUser(request).orElse(createRequestProxyUser).orElse(Option(remoteUser(request)))
   }
 
   /**
    * Gets the request user or impersonated user to determine the effective user.
    */
   protected def effectiveUser(request: HttpServletRequest): String = {
-    impersonatedUser(request).getOrElse(remoteUser(request))
+    val requestUser = remoteUser(request)
+    impersonatedUser(request).orElse(Option(requestUser))
+      .filter(accessManager.checkImpersonation(requestUser, _)).orNull
   }
 
   /**
--- a/server/src/main/scala/org/apache/livy/server/SessionServlet.scala
+++ b/server/src/main/scala/org/apache/livy/server/SessionServlet.scala
@@ -156,26 +156,37 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
   }
 
   /**
-    * Returns the remote user for the given request. Separate method so that tests can override it.
-    */
-  protected def remoteUser(req: HttpServletRequest): String = req.getRemoteUser
+   * Returns the remote user for the given request. Separate method so that tests can override it.
+   */
+  protected def remoteUser(req: HttpServletRequest): String = req.getRemoteUser()
 
+  /**
+   * Returns the impersonated user as given by "doAs" as a request parameter.
+   */
   protected def impersonatedUser(request: HttpServletRequest): Option[String] = {
     val impersonatedUser = Option(request.getParameter("doAs"))
     impersonatedUser.filter(accessManager.checkImpersonation(remoteUser(request), _))
   }
 
   /**
-    * Returns the proxy user as determined by the json request body or owner if available.
-    * This is necessary to preserve backwards compatibility prior to "doAs" impersonation.
-    */
+   * Returns the proxyUser for the given request.
+   */
+  protected def proxyUser(request: HttpServletRequest,
+      createRequestProxyUser: Option[String]): Option[String] = {
+    impersonatedUser(request).orElse(legacyProxyUser(remoteUser(request), createRequestProxyUser))
+  }
+
+  /**
+   * Returns the proxy user as determined by the json request body or owner if available.
+   * This is necessary to preserve backwards compatibility prior to "doAs" impersonation.
+   */
   protected def legacyProxyUser(owner: String, bodyProxyUser: Option[String]): Option[String] = {
     bodyProxyUser.filter(accessManager.checkImpersonation(owner, _)).orElse(Option(owner))
   }
 
   /**
-    * Gets the request user or impersonated user to determine the effective user
-    */
+   * Gets the request user or impersonated user to determine the effective user.
+   */
   protected def effectiveUser(request: HttpServletRequest): String = {
     impersonatedUser(request).getOrElse(remoteUser(request))
   }
--- a/server/src/main/scala/org/apache/livy/server/SessionServlet.scala
+++ b/server/src/main/scala/org/apache/livy/server/SessionServlet.scala
@@ -180,7 +180,8 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
    * Gets the request user or impersonated user to determine the effective user.
    */
   protected def effectiveUser(request: HttpServletRequest): String = {
-    accessManager.checkImpersonation(impersonatedUser(request), remoteUser(request)).orNull
+    val requestUser = remoteUser(request)
+    accessManager.checkImpersonation(impersonatedUser(request), requestUser).getOrElse(requestUser)
   }
 
   /**
--- a/server/src/main/scala/org/apache/livy/server/SessionServlet.scala
+++ b/server/src/main/scala/org/apache/livy/server/SessionServlet.scala
@@ -172,7 +172,7 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
    */
   protected def proxyUser(request: HttpServletRequest,
       createRequestProxyUser: Option[String]): Option[String] = {
-    impersonatedUser(request).orElse(createRequestProxyUser).orElse(Option(remoteUser(request)))
+    impersonatedUser(request).orElse(createRequestProxyUser)
   }
 
   /**
--- a/server/src/main/scala/org/apache/livy/server/SessionServlet.scala
+++ b/server/src/main/scala/org/apache/livy/server/SessionServlet.scala
@@ -180,8 +180,8 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
    */
   protected def effectiveUser(request: HttpServletRequest): String = {
     val requestUser = remoteUser(request)
-    impersonatedUser(request).orElse(Option(requestUser))
-      .filter(accessManager.checkImpersonation(requestUser, _)).orNull
+    impersonatedUser(request).filter(accessManager.checkImpersonation(requestUser, _))
+      .getOrElse(requestUser)
   }
 
   /**
--- a/server/src/main/scala/org/apache/livy/server/SessionServlet.scala
+++ b/server/src/main/scala/org/apache/livy/server/SessionServlet.scala
@@ -170,7 +170,8 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
   /**
    * Returns the proxyUser for the given request.
    */
-  protected def proxyUser(request: HttpServletRequest,
+  protected def proxyUser(
+      request: HttpServletRequest,
       createRequestProxyUser: Option[String]): Option[String] = {
     impersonatedUser(request).orElse(createRequestProxyUser)
   }
@@ -179,9 +180,7 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
    * Gets the request user or impersonated user to determine the effective user.
    */
   protected def effectiveUser(request: HttpServletRequest): String = {
-    val requestUser = remoteUser(request)
-    impersonatedUser(request).filter(accessManager.checkImpersonation(requestUser, _))
-      .getOrElse(requestUser)
+    accessManager.checkImpersonation(impersonatedUser(request), remoteUser(request)).orNull
   }
 
   /**
@@ -207,11 +206,11 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
 
   private def doWithSession(fn: (S => Any),
       allowAll: Boolean,
-      checkFn: Option[(Session, String) => Boolean]): Any = {
+      checkFn: Option[(String, String) => Boolean]): Any = {
     val sessionId = params("id").toInt
     sessionManager.get(sessionId) match {
       case Some(session) =>
-        if (allowAll || checkFn.map(_(session, effectiveUser(request))).getOrElse(false)) {
+        if (allowAll || checkFn.map(_(session.owner, effectiveUser(request))).getOrElse(false)) {
           fn(session)
         } else {
           Forbidden()
--- a/server/src/main/scala/org/apache/livy/server/SessionServlet.scala
+++ b/server/src/main/scala/org/apache/livy/server/SessionServlet.scala
@@ -156,9 +156,29 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
   }
 
   /**
-   * Returns the remote user for the given request. Separate method so that tests can override it.
-   */
-  protected def remoteUser(req: HttpServletRequest): String = req.getRemoteUser()
+    * Returns the remote user for the given request. Separate method so that tests can override it.
+    */
+  protected def remoteUser(req: HttpServletRequest): String = req.getRemoteUser
+
+  protected def impersonatedUser(request: HttpServletRequest): Option[String] = {
+    val impersonatedUser = Option(request.getParameter("doAs"))
+    impersonatedUser.filter(accessManager.checkImpersonation(remoteUser(request), _))
+  }
+
+  /**
+    * Returns the proxy user as determined by the json request body or owner if available.
+    * This is necessary to preserve backwards compatibility prior to "doAs" impersonation.
+    */
+  protected def legacyProxyUser(owner: String, bodyProxyUser: Option[String]): Option[String] = {
+    bodyProxyUser.filter(accessManager.checkImpersonation(owner, _)).orElse(Option(owner))
+  }
+
+  /**
+    * Gets the request user or impersonated user to determine the effective user
+    */
+  protected def effectiveUser(request: HttpServletRequest): String = {
+    impersonatedUser(request).getOrElse(remoteUser(request))
+  }
 
   /**
    * Performs an operation on the session, without checking for ownership. Operations executed
@@ -183,11 +203,11 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
 
   private def doWithSession(fn: (S => Any),
       allowAll: Boolean,
-      checkFn: Option[(String, String) => Boolean]): Any = {
+      checkFn: Option[(Session, String) => Boolean]): Any = {
     val sessionId = params("id").toInt
     sessionManager.get(sessionId) match {
       case Some(session) =>
-        if (allowAll || checkFn.map(_(session.owner, remoteUser(request))).getOrElse(false)) {
+        if (allowAll || checkFn.map(_(session, effectiveUser(request))).getOrElse(false)) {
           fn(session)
         } else {
           Forbidden()
--- a/server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
+++ b/server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
@@ -57,10 +57,12 @@ object BatchSession extends Logging {
       livyConf: LivyConf,
       accessManager: AccessManager,
       owner: String,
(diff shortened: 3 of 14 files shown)


risdenk:
Most of the changes are to pass the Session and HttpServletRequest around to allow for checking different parts of the request with the access manager.

risdenk:
The latest build succeeded except for "Spark 2.3 Unit Tests"

```
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running org.apache.livy.rsc.TestJobHandle
Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.634 sec - in org.apache.livy.rsc.TestJobHandle
Running org.apache.livy.rsc.TestSparkClient
Tests run: 19, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 57.489 sec <<< FAILURE! - in org.apache.livy.rsc.TestSparkClient
testRemoteClient(org.apache.livy.rsc.TestSparkClient)  Time elapsed: 4.546 sec  <<< ERROR!
java.util.concurrent.ExecutionException: java.io.IOException: RSCClient instance stopped.
	at org.apache.livy.rsc.TestSparkClient.runTest(TestSparkClient.java:588)
	at org.apache.livy.rsc.TestSparkClient.runTest(TestSparkClient.java:575)
	at org.apache.livy.rsc.TestSparkClient.testRemoteClient(TestSparkClient.java:213)
Caused by: java.io.IOException: RSCClient instance stopped.

Running org.apache.livy.rsc.rpc.TestKryoMessageCodec
Tests run: 9, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.042 sec - in org.apache.livy.rsc.rpc.TestKryoMessageCodec
Running org.apache.livy.rsc.rpc.TestRpc
Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.407 sec - in org.apache.livy.rsc.rpc.TestRpc

Results :

Tests in error: 
  TestSparkClient.testRemoteClient:213->runTest:575->runTest:588 Â» Execution jav...
```

None of the other Spark versions failed. I can't rekick off the single job since it looks to be a transient failure.

server/src/main/scala/org/apache/livy/server/AccessManager.scala
```diff
@@ -97,47 +99,71 @@ private[livy] class AccessManager(conf: LivyConf) extends Logging {
...
+      throw new AccessControlException(
+        s"User '$requestUser' not allowed to impersonate '$impersonatedUser'.")
     }
+    false
+  }
+
+  def checkProxyUser(request: HttpServletRequest, impersonatedUser: String): Boolean = {
+    val proxyUser = getRequestUser(request)
+    val remoteHost = request.getRemoteHost
+    val allowedHosts = conf.hadoopConf.get("hadoop.proxyuser." + proxyUser + ".hosts")
```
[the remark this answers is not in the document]
risdenk: The idea was to model after https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/Superusers.html#Configurations and how a bunch of other Hadoop ecosystem components work.

Superusers can impersonate anyone yes. There should be a limit as to who can impersonate who. There should also be a limit from where superuser requests can come from.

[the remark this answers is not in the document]
vanzin: If you want to add this functionality, you shouldn't use the Hadoop configs. Every different service has its own configs for this.

It should also be a separate change, since it's not really related to what you're implementing here.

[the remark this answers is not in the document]
risdenk: Yea thats a valid point that it should be a separate change.

risdenk:
Pushed updated changes that:
* removes the Hadoop proxyuser configs 
* fixes formatting for parameters in multi-line declarations
* Renamed `checkBodyProxyUser` method to explain better what it does
* Added/improved comments to new/changed methods

server/src/test/scala/org/apache/livy/server/AccessManagerSuite.scala
```diff
@@ -116,4 +121,60 @@ class AccessManagerSuite extends FunSuite with Matchers with LivyBaseUnitTestSui
...
+    val conf = new LivyConf()
+      .set(LivyConf.IMPERSONATION_ENABLED, true)
+    val accessManager = new AccessManager(conf)
+
+    val req = mock[HttpServletRequest]
+    when(req.getRemoteUser).thenReturn("abc")
+    when(req.getParameter("doAs")).thenReturn("def")
+
+    accessManager.getRequestUser(req) should be("abc")
+    try {
```
vanzin:
```
val e = intercept[AccessControlException] {
  // something that throws
}
```

server/src/test/scala/org/apache/livy/server/AccessManagerSuite.scala
```diff
@@ -116,4 +121,60 @@ class AccessManagerSuite extends FunSuite with Matchers with LivyBaseUnitTestSui
...
+      accessManager.getImpersonatedUser(req) should be(Some("def"))
```
vanzin: the "should be" doesn't seem necessary since you expect an exception here. Also in the next test.

risdenk:
* Added tests to `SessionServletSpec` and `AclsEnabledSessionServletSpec`
* Cleaned up test error handled to use `intercept[...]`
* removed unnecessary `should be` usage

server/src/main/scala/org/apache/livy/server/batch/BatchSessionServlet.scala
```diff
@@ -42,20 +42,24 @@ class BatchSessionServlet(
 
   override protected def createSession(req: HttpServletRequest): BatchSession = {
     val createRequest = bodyAs[CreateBatchRequest](req)
+    val owner = accessManager.getRequestUser(req)
+    val legacyProxyUser = getLegacyProxyUser(owner, createRequest.proxyUser, req)
+    val impersonatedUser = getImpersonatedUser(req).orElse(legacyProxyUser)
```
mgaido91: we are doing the parsing for `legacyProxyUser ` also when it is not needed, we can do:
```
orElse(getLegacyProxyUser(....))
```

risdenk: Addressed in latest commit.

server/src/main/scala/org/apache/livy/server/AccessManager.scala
```diff
@@ -98,46 +100,79 @@ private[livy] class AccessManager(conf: LivyConf) extends Logging {
...
   /**
-   * Checks that the requesting user can impersonate the target user.
-   * If the user does not have permission to impersonate, then throws an `AccessControlException`.
-   *
-   * @return The user that should be impersonated. That can be the target user if defined, the
-   *         request's user - which may not be defined - otherwise, or `None` if impersonation is
-   *         disabled.
+   * Gets the remote user from the given request
+   */
+  def getRequestUser(request: HttpServletRequest): String = {
```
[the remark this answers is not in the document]
risdenk: Logic wasn't bad to move out of AccessManager so took that approach in latest commit.

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
```diff
@@ -55,12 +54,11 @@ object BatchSession extends Logging {
       id: Int,
       request: CreateBatchRequest,
       livyConf: LivyConf,
-      accessManager: AccessManager,
       owner: String,
+      impersonatedUser: Option[String],
       sessionStore: SessionStore,
       mockApp: Option[SparkApp] = None): BatchSession = {
     val appTag = s"livy-batch-$id-${Random.alphanumeric.take(8).mkString}"
-    val impersonatedUser = accessManager.checkImpersonation(request.proxyUser, owner, livyConf)
```
mgaido91: I am not sure about this. I feel it is a bit risky to make this check outside here, because we do not guarantee anymore that we create the session only if the impersonated user has the right to do so. Can't we leave it here?

server/src/main/scala/org/apache/livy/server/AccessManager.scala
```diff
@@ -98,46 +99,51 @@ private[livy] class AccessManager(conf: LivyConf) extends Logging {
...
-   *         request's user - which may not be defined - otherwise, or `None` if impersonation is
-   *         disabled.
+   * Checks that the request user can impersonate the target user.
+   * If impersonation is enabled and the user does not have permission to impersonate
+   * then throws an `AccessControlException`. If impersonation is disabled returns false
    */
   def checkImpersonation(
-      target: Option[String],
       requestUser: String,
-      livyConf: LivyConf): Option[String] = {
```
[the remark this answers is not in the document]
vanzin: I don't really see a use case for the conf instances being different. At the call sites, it's just a matter of sub-optimal APIs (e.g. `BatchSession.create` being a static method and thus not having any state associated, like a configuration object).

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
vanzin: It's just moving the check to the servlet, which happens before this is called. Are you worried that another call site might be added and we forget about the access check?

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -156,9 +156,39 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
   }
 
   /**
-   * Returns the remote user for the given request. Separate method so that tests can override it.
-   */
-  protected def remoteUser(req: HttpServletRequest): String = req.getRemoteUser()
+    * Gets the remote user from the given request
+    */
+  protected def getRemoteUser(request: HttpServletRequest): String = {
```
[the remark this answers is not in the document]
vanzin: Also I prefer the old naming style, without the java bean-esque "get", for the new methods below too.

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
mgaido91: yes, exactly, that's my worry.

server/src/main/scala/org/apache/livy/server/AccessManager.scala
```diff
@@ -98,46 +99,51 @@ private[livy] class AccessManager(conf: LivyConf) extends Logging {
...
   /**
-   * Check that the request's user has modify access to resources owned by the given target user.
+   * Check that the request user has modify access to the given session.
    */
-  def hasModifyAccess(target: String, requestUser: String): Boolean = {
-    requestUser == target || checkModifyPermissions(requestUser)
+  def hasModifyAccess(
+      session: Session,
+      effectiveUser: String): Boolean = {
+    session.owner == effectiveUser  ||
```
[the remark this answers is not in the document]
risdenk: If you use a superuser to create a session, the owner is set to the superuser and proxyUser is the user running the job. If you try to access Livy directly as the user running the job, you won't be able to see details. This is weird since you should be able to see your own job. This just ass the check to see if the incoming user is either the owner or the proxyUser of the session.

server/src/main/scala/org/apache/livy/server/AccessManager.scala
```diff
@@ -98,46 +99,51 @@ private[livy] class AccessManager(conf: LivyConf) extends Logging {
...
   /**
-   * Check that the request's user has view access to resources owned by the given target user.
+   * Check that the request user has view access to the given session
    */
-  def hasViewAccess(target: String, requestUser: String): Boolean = {
-    requestUser == target || checkViewPermissions(requestUser)
+  def hasViewAccess(
+      session: Session,
+      effectiveUser: String): Boolean = {
+    session.owner == effectiveUser ||
```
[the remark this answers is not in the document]
risdenk: If you use a superuser to create a session, the owner is set to the superuser and proxyUser is the user running the job. If you try to access Livy directly as the user running the job, you won't be able to see details. This is weird since you should be able to see your own job. This just ass the check to see if the incoming user is either the owner or the proxyUser of the session.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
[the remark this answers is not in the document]
risdenk: Yea I can change it back. This happened from the refactor from accessManager back to SessionServlet.

server/src/main/scala/org/apache/livy/server/AccessManager.scala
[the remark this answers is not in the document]
mgaido91: Yes, but IIUC this is a behavior change. We can argue whether it is right or not the current behavior and eventually change it, but we should definitely not do that in this PR, as it is unrelated with it. If I am not missing something, please revert these changes and create a separate ticket in order to discuss them separately.

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
risdenk: Ok I put the accessManager back

risdenk:
Ok I think I addressed all the review comments:
* removed get from method names
* added accessManager back to session create() methods
* Removed checking of proxyuser in accessmanager for permissions
* fixed minor formatting issues

server/src/main/scala/org/apache/livy/server/AccessManager.scala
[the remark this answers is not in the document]
risdenk: Changes reverted in latest push

server/src/main/scala/org/apache/livy/server/AccessManager.scala
```diff
@@ -98,46 +99,43 @@ private[livy] class AccessManager(conf: LivyConf) extends Logging {
...
   /**
-   * Checks that the requesting user can impersonate the target user.
-   * If the user does not have permission to impersonate, then throws an `AccessControlException`.
-   *
-   * @return The user that should be impersonated. That can be the target user if defined, the
-   *         request's user - which may not be defined - otherwise, or `None` if impersonation is
-   *         disabled.
+   * Checks that the request user can impersonate the target user.
+   * If impersonation is enabled and the user does not have permission to impersonate
+   * then throws an `AccessControlException`. If impersonation is disabled returns false
```
mgaido91: nit: missing dot

server/src/main/scala/org/apache/livy/server/AccessManager.scala
```diff
@@ -98,46 +99,43 @@ private[livy] class AccessManager(conf: LivyConf) extends Logging {
...
-      target.orElse(Option(requestUser))
     } else {
-      None
+      false
     }
   }
 
   /**
-   * Check that the requesting user has admin access to resources owned by the given target user.
+   * Check that the request user has is able to impersonate the given user.
```
mgaido91: the new description is wrong IMHO, and these are unneeded changes (just renaming variables). Please avoid them

server/src/main/scala/org/apache/livy/server/AccessManager.scala
```diff
@@ -98,46 +99,43 @@ private[livy] class AccessManager(conf: LivyConf) extends Logging {
...
    */
-  def hasModifyAccess(target: String, requestUser: String): Boolean = {
-    requestUser == target || checkModifyPermissions(requestUser)
+  def hasModifyAccess(session: Session, effectiveUser: String): Boolean = {
+    session.owner == effectiveUser || checkModifyPermissions(effectiveUser)
   }
 
   /**
-   * Check that the request's user has view access to resources owned by the given target user.
+   * Check that the request user has view access to the given session
```
mgaido91: nit: missing dot

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -156,9 +156,29 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
   }
 
   /**
-   * Returns the remote user for the given request. Separate method so that tests can override it.
-   */
-  protected def remoteUser(req: HttpServletRequest): String = req.getRemoteUser()
+    * Returns the remote user for the given request. Separate method so that tests can override it.
```
mgaido91: bad indentation and unneeded changes, please avoid them

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -156,9 +156,29 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
...
+    */
+  protected def remoteUser(req: HttpServletRequest): String = req.getRemoteUser
+
+  protected def impersonatedUser(request: HttpServletRequest): Option[String] = {
+    val impersonatedUser = Option(request.getParameter("doAs"))
+    impersonatedUser.filter(accessManager.checkImpersonation(remoteUser(request), _))
+  }
+
+  /**
+    * Returns the proxy user as determined by the json request body or owner if available.
```
mgaido91: nit: bad indent

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -156,9 +156,29 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
...
+    * This is necessary to preserve backwards compatibility prior to "doAs" impersonation.
+    */
+  protected def legacyProxyUser(owner: String, bodyProxyUser: Option[String]): Option[String] = {
+    bodyProxyUser.filter(accessManager.checkImpersonation(owner, _)).orElse(Option(owner))
+  }
+
+  /**
+    * Gets the request user or impersonated user to determine the effective user
```
mgaido91: ditto

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -183,11 +203,11 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
 
   private def doWithSession(fn: (S => Any),
       allowAll: Boolean,
-      checkFn: Option[(String, String) => Boolean]): Any = {
+      checkFn: Option[(Session, String) => Boolean]): Any = {
     val sessionId = params("id").toInt
     sessionManager.get(sessionId) match {
       case Some(session) =>
-        if (allowAll || checkFn.map(_(session.owner, remoteUser(request))).getOrElse(false)) {
+        if (allowAll || checkFn.map(_(session, effectiveUser(request))).getOrElse(false)) {
```
mgaido91: again, this is a behavior change. Probably it is fine to leave it here, but at least we need to document this behavior change. cc @vanzin

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
```diff
@@ -57,10 +57,12 @@ object BatchSession extends Logging {
       livyConf: LivyConf,
       accessManager: AccessManager,
       owner: String,
+      impersonatedUser: Option[String],
       sessionStore: SessionStore,
       mockApp: Option[SparkApp] = None): BatchSession = {
     val appTag = s"livy-batch-$id-${Random.alphanumeric.take(8).mkString}"
-    val impersonatedUser = accessManager.checkImpersonation(request.proxyUser, owner, livyConf)
+
+    impersonatedUser.filter(accessManager.checkImpersonation(owner, _))
```
mgaido91: the result of the filter is never used, so it is useless as it is now... Moreover it is weird that tests are passing nonetheless... may you please add a UT for this?

server/src/main/scala/org/apache/livy/server/batch/BatchSessionServlet.scala
```diff
@@ -42,20 +42,25 @@ class BatchSessionServlet(
 
   override protected def createSession(req: HttpServletRequest): BatchSession = {
     val createRequest = bodyAs[CreateBatchRequest](req)
+    val owner = remoteUser(req)
+    val proxyUser = impersonatedUser(req)
```
mgaido91: this code is duplicated, what about creating a method `proxyUser`?

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
```diff
@@ -67,14 +67,16 @@ object InteractiveSession extends Logging {
...
       livyConf: LivyConf,
       accessManager: AccessManager,
       request: CreateInteractiveRequest,
       sessionStore: SessionStore,
       mockApp: Option[SparkApp] = None,
       mockClient: Option[RSCClient] = None): InteractiveSession = {
     val appTag = s"livy-session-$id-${Random.alphanumeric.take(8).mkString}"
-    val impersonatedUser = accessManager.checkImpersonation(request.proxyUser, owner, livyConf)
+
+    impersonatedUser.filter(accessManager.checkImpersonation(owner, _))
```
mgaido91: ditto

server/src/main/scala/org/apache/livy/sessions/Session.scala
```diff
@@ -204,7 +204,7 @@ abstract class Session(val id: Int, val owner: String, val livyConf: LivyConf)
 
   protected def stopSession(): Unit
 
-  protected val proxyUser: Option[String]
+  val proxyUser: Option[String]
```
mgaido91: please add a comment:
```
// Visible for testing.
```

server/src/test/scala/org/apache/livy/server/BaseSessionServletSpec.scala
```diff
@@ -77,7 +77,6 @@ abstract class BaseSessionServletSpec[S <: Session, R <: RecoveryMetadata]
   addServlet(servlet, "/*")
 
   protected def toJson(msg: AnyRef): Array[Byte] = mapper.writeValueAsBytes(msg)
-
```
mgaido91: in general we best avoid doing unneeded changes like these, makes easier checking the history of the changes.. so please undo the changes.

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
risdenk: Tests are passing because BatchSession/InteractiveSession create are never called directly. This was defensively put in based on your other comments. I reverted the change to remove it based on your request.

Having a test that fails on a PR is pointless. I don't understand what this comment of yours is even after.

Finally the result of the call doesn't actually matter since there will be an exception thrown by checkImpersonation if there is a problem. Otherwise security is turned off. There is no reason to filter the impersonated user. It is just cleaner than doing an isDefined check and a .get.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
risdenk: If you only ever look at the remoteUser you can't authorize any other types of requests. That is the whole point of this PR that you can authorize not just POST create session/batch requests. This is described in the description of LIVY-551. This change is required.

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
mgaido91: well, then `foreach` is better than `filter` if we return nothing. Anyway, it actually works because you are doing the same check in `def impersonatedUser` and `def legacyProxyUser`, so here you are relying on them, rather than on this check .What I meant was doing the check here as it was earlier and in those methods just check which user needs to be impersonated. I see that you need that for the `effectiveUser` call, but I think we can move the filter there and use this one here.

risdenk: I think the latest changes in 7829cd9 are what you were after.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -160,6 +160,30 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
...
+    impersonatedUser(request).orElse(createRequestProxyUser).orElse(Option(remoteUser(request)))
+  }
+
+  /**
+   * Gets the request user or impersonated user to determine the effective user.
+   */
+  protected def effectiveUser(request: HttpServletRequest): String = {
+    val requestUser = remoteUser(request)
+    impersonatedUser(request).orElse(Option(requestUser))
+      .filter(accessManager.checkImpersonation(requestUser, _)).orNull
```
mgaido91: I think this should be:
```
impersonatedUser(request).filter(accessManager.checkImpersonation(requestUser, _)).getOrElse(requestUser)
```
Am I wrong? Why are you checking the impersonation also when there is no impersonated user?

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
```diff
@@ -57,10 +57,12 @@ object BatchSession extends Logging {
       livyConf: LivyConf,
       accessManager: AccessManager,
       owner: String,
+      impersonatedUser: Option[String],
       sessionStore: SessionStore,
       mockApp: Option[SparkApp] = None): BatchSession = {
     val appTag = s"livy-batch-$id-${Random.alphanumeric.take(8).mkString}"
-    val impersonatedUser = accessManager.checkImpersonation(request.proxyUser, owner, livyConf)
+
+    impersonatedUser.foreach(accessManager.checkImpersonation(owner, _))
```
mgaido91: IIUC, as the code is now, when impersonation is disabled and we are using the old proxyUser feature, we would allow the impersonation, right? Because having the foreach here doesn't throw any exception and the returned false is just ignored.

Anyway, can we add a UT for this case, in order to ensure the behavior is fine?

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -160,6 +160,30 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
...
+  protected def impersonatedUser(request: HttpServletRequest): Option[String] = {
+    Option(request.getParameter("doAs"))
+  }
+
+  /**
+   * Returns the proxyUser for the given request.
+   */
+  protected def proxyUser(request: HttpServletRequest,
+      createRequestProxyUser: Option[String]): Option[String] = {
+    impersonatedUser(request).orElse(createRequestProxyUser).orElse(Option(remoteUser(request)))
```
mgaido91: why `orElse(Option(remoteUser(request)))`? In that case we should return `None` here

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
risdenk: I think you are right here.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
risdenk: So this caused a failure in one of the existing test cases if you don't have the remoteUser set as the proxyUser. 

https://github.com/apache/incubator-livy/blob/master/server/src/test/scala/org/apache/livy/server/interactive/JobApiSpec.scala#L132

I don't know why that is the case but didn't change that existing behavior.

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
risdenk: Yea so there are no test suites that previously test the no impersonation code path. `BatchServletSpec` and `InteractiveServletSpec` both mock the Session `create()` method so there is no tests there. `JobApiSpec` is the closest to a real test but that one uses impersonation enabled always with no easy way to override that config. Let me look into it a bit.

server/src/test/scala/org/apache/livy/server/interactive/JobApiSpec.scala
```diff
@@ -227,3 +226,10 @@ class JobApiSpec extends BaseInteractiveServletSpec {
   }
 
 }
+
+class JobApiSpecNoImpersonation extends JobApiSpec {
```
risdenk: Currently this shows that the behavior is the same without and without impersonation which is most likely not correct. I need to go back to the master branch and see what the behavior is.

FYI @mgaido91

risdenk:
Force pushed only the last commit to add updated tests from master that are now failing on this branch. I need to see how they affect the authorization being pushed around.

server/src/test/scala/org/apache/livy/server/interactive/JobApiSpec.scala
risdenk: Updated to now have tests that pass on master and fail on this branch. Working on fixing them.

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
risdenk: ebf3514 fixes the test failures from the added UT for this behavior. The old behavior is now restored and confirmed via tests.

server/src/test/scala/org/apache/livy/server/interactive/JobApiSpec.scala
risdenk: ebf3514 fixes these tests.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
mgaido91: Because previously the check was in `checkImpersonation`. Since you moved it out from there, you should have been put it somewhere else. The point is: this is not the right place where to put it, since this is not the `proxyUser` which this method should return and to match previous behavior this should be done only when impersonation in enabled (otherwise we'd need None at the end).

risdenk: hmmm yea I see your point. Let me take another look.

risdenk:
Latest commit tries to address proxyUser enforcement separate from determining the proxyUser from the request.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
risdenk: Commit 3772534 should address this concern.

server/src/main/scala/org/apache/livy/server/AccessManager.scala
```diff
@@ -98,23 +99,32 @@ private[livy] class AccessManager(conf: LivyConf) extends Logging {
...
+   * If impersonation is enabled and the user does not have permission to impersonate
+   * then throws an `AccessControlException`. If impersonation is disabled returns false.
    */
   def checkImpersonation(
-      target: Option[String],
       requestUser: String,
-      livyConf: LivyConf): Option[String] = {
-    if (livyConf.getBoolean(LivyConf.IMPERSONATION_ENABLED)) {
-      if (!target.forall(hasSuperAccess(_, requestUser))) {
+      impersonatedUser: String): Boolean = {
```
mgaido91: I am wondering why you are changing this to Boolean and introducing a new method for doing what this was doing earlier. Seems you are just complicating things.

server/src/main/scala/org/apache/livy/server/AccessManager.scala
```diff
@@ -128,16 +138,16 @@ private[livy] class AccessManager(conf: LivyConf) extends Logging {
   }
 
   /**
-   * Check that the request's user has modify access to resources owned by the given target user.
+   * Check that the request user has modify access to the given session.
    */
-  def hasModifyAccess(target: String, requestUser: String): Boolean = {
-    requestUser == target || checkModifyPermissions(requestUser)
+  def hasModifyAccess(session: Session, effectiveUser: String): Boolean = {
```
mgaido91: as mentioned, these changes are useless...

server/src/test/scala/org/apache/livy/server/SessionServletSpec.scala
```diff
@@ -62,27 +59,27 @@ object SessionServletSpec {
     new SessionServlet(sessionManager, conf, accessManager) with RemoteUserOverride {
       override protected def createSession(req: HttpServletRequest): Session = {
         val params = bodyAs[Map[String, String]](req)
-        accessManager.checkImpersonation(params.get(PROXY_USER), remoteUser(req), livyConf)
-        new MockSession(sessionManager.nextId(), remoteUser(req), conf)
+        val bodyImpersonatedUser = params.get(PROXY_USER)
+          .filter(accessManager.checkImpersonation(remoteUser(req), _))
```
mgaido91: I think this should be executed after the orElse of the following line...

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -183,11 +207,11 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
 
   private def doWithSession(fn: (S => Any),
       allowAll: Boolean,
-      checkFn: Option[(String, String) => Boolean]): Any = {
+      checkFn: Option[(Session, String) => Boolean]): Any = {
```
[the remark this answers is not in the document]
risdenk: Change left over from checking the session proxy user. Will create separate JIRA for that change.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -160,6 +160,29 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
...
+      request: HttpServletRequest,
+      createRequestProxyUser: Option[String]): Option[String] = {
+    impersonatedUser(request).orElse(createRequestProxyUser)
+  }
+
+  /**
+   * Gets the request user or impersonated user to determine the effective user.
+   */
+  protected def effectiveUser(request: HttpServletRequest): String = {
+    accessManager.checkImpersonation(impersonatedUser(request), remoteUser(request)).orNull
```
mgaido91: why orNull? Shouldn't be `remoteUser(request)` in that case? I'd expect some NullPointerException to happen with `orNull` or anyway, some problems when checking access permissions (since we have null instead of the requesting user). May you please also add a UT for this?

server/src/test/scala/org/apache/livy/server/interactive/JobApiSpec.scala
```diff
@@ -227,3 +230,48 @@ class JobApiSpec extends BaseInteractiveServletSpec {
...
+  override protected def createConf(): LivyConf = synchronized {
+    super.createConf()
+      .set(LivyConf.IMPERSONATION_ENABLED, false)
```
mgaido91: nit: one the previous line

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
risdenk: Well so I had a similar question looking at the `checkImpersonation` method. We shouldn't even need to have any `orNull` or anything. We should be able to return a `String` instead of `Option[String]` from `checkImpersonation`. Since we either are impersonating and that succeeds or we are returning `None` from `checkImpersonation` today. 

I have to think about it a bit more and probably won't be until Monday.

risdenk: Ok thought about this a bit more. `remoteUser(request)` I think is correct but there is no guarantee that `remoteUser(request)` will not be null (in the case of no Kerberos security for example). I checked the access checks and null is handled just fine.

I can try to look at adding a unit test for this.

server/src/test/scala/org/apache/livy/server/interactive/JobApiSpec.scala
risdenk: Hm not sure why that was on 2 to start with. Fixed now.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
risdenk: JobApiSpecNoImpersonation should handle this case actually. I added some more tests here as well around sending in no user (ie: no authentication enabled where request.getRemoteUser would be null.

risdenk:
There is a `docs/rest-api.md` file? Maybe that is the right place for docs. I could add a section there on the added ability for `doAs` parameter.

risdenk:
Last commit made effectiveUser return the remoteUser as a fall back, added tests, and cleaned up style based on review.

mgaido91:
@risdenk yes, that is the right place for sure, but since this parameter is valid for all the endpoints, I am not sure whether to create a new section for common parameters having this single one for the moment or do something else. So I'd like to hear if someone has a good/strong opinion on this.

The code itself now looks good to me. The only point I think still need to be addressed is the documentation.

ajbozarth:
I agree that this should be added to `rest-api.md` in some way, but this is the first param shared across all calls so I'm unsure the best way to add it. Given it's optional perhaps giving it it's own section towards the bottom with a quick explanation would be best. How about your give documenting it your best shot and I can give better feedback based on what you write.

server/src/test/scala/org/apache/livy/server/SessionServletSpec.scala
```diff
@@ -112,17 +109,60 @@ class SessionServletSpec extends BaseSessionServletSpec[Session, RecoveryMetadat
...
+          assert(res.logs === IndexedSeq("log"))
+        }
+        delete(res.id, aliceHeaders, SC_OK)
+      }
+
+      jpost[MockSessionView]("/?doAs=alice", Map(), headers = adminHeaders) { res =>
+        jget[MockSessionView](s"/${res.id}", headers = bobHeaders) { res =>
+          assert(res.owner === ADMIN)
+          assert(res.proxyUser === Some("alice"))
           assert(res.logs === IndexedSeq("log"))
```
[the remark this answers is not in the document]
risdenk: Let me double check and I can add a check for the scenario you described as well if it doesn't exist.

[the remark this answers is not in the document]
risdenk: So the reason this works on line ~165 is because ACLs aren't enabled. In the same `SessionServletSpec.scala` file on line ~245 in the class `AclsEnabledSessionServletSpec` you can see the same test with logs is null. This is because ACLs are enabled.

I don't see the exact scenario you outlined but added a test for admin with `doAs=bob` showing that permissions are adhered to.

risdenk:
Added test that @vanzin asked about. Added docs in a new section to document `doAs`. Merged master branch to fix merge conflicts after LIVY-41 was committed to master.

server/src/test/scala/org/apache/livy/server/interactive/JobApiSpec.scala
```diff
@@ -227,3 +230,59 @@ class JobApiSpec extends BaseInteractiveServletSpec {
...
+        data.proxyUser should be (null)
+        val user = runJob(data.id, new GetCurrentUser(), headers = headers)
+        user should be (UserGroupInformation.getCurrentUser.getUserName)
+      } finally {
+        deleteSession(data.id)
+      }
+    }
+  }
+
+  it("should not honor impersonation requests") {
```
vanzin: I'm not sure I follow these tests. What does it mean to not honor impersonation requests?

I'd expect that if impersonation is disabled and you ask for it, you'd get an error. That makes more sense to me than falling back to some other mode.

server/src/test/scala/org/apache/livy/server/SessionServletSpec.scala
```diff
@@ -114,17 +111,66 @@ class SessionServletSpec extends BaseSessionServletSpec[Session, RecoveryMetadat
...
         delete(res.id, aliceHeaders, SC_OK)
       }
     }
 
     it("should allow other users to see non-sensitive information") {
+      jpost[MockSessionView]("/", Map()) { res =>
+        jget[MockSessionView](s"/${res.id}", headers = bobHeaders) { res =>
+          assert(res.owner === null)
+          assert(res.proxyUser === None)
+          assert(res.logs === IndexedSeq("log"))
```
vanzin: Sorry, but this is really bothering me.

The test says "should allow other users to see non-sensitive information", and logs are sensitive information. So if "bob" is not the owner, and is not a super user, then either this test is wrong or the test name is wrong.

Seems like no security is enabled in these tests, so basically everyone has access to everything, so the test name should reflect that.

And if it's true that no security is enabled here, I'm not sure what's the point of adding impersonation-related tests.

server/src/test/scala/org/apache/livy/server/interactive/JobApiSpec.scala
risdenk: This matches the behavior from before this change. I added this change on master and made sure the tests pass before pulling to this branch. @mgaido91 was very clear that there should be no behavior changes.

server/src/test/scala/org/apache/livy/server/SessionServletSpec.scala
risdenk: This test was here before and the behavior is the same as before. The test name has been wrong before this change then.

@mgaido91 asked what happens if you don't send in a username and that is what was added. I also added the `doAs` tests to make sure behavior was the same if you added doAs or not. 

If there are no ACLs then users can do whatever they want. This is true before this doAs was added too.

I can update the test name but I want to be clear this was like this before `doAs`.

server/src/test/scala/org/apache/livy/server/interactive/JobApiSpec.scala
risdenk: To be honest I do agree with you that if impersonation is disabled and you ask for you you should get an error but that is not how Livy was setup previously. You will get an error if impersonation is enabled and you aren't a superuser, but if you have impersonation disabled, it falls back to the request user.

@mgaido91 made it clear not to change the behavior of `AccessManager` `checkImpersonation`

server/src/main/scala/org/apache/livy/server/AccessManager.scala
```diff
@@ -107,9 +107,8 @@ private[livy] class AccessManager(conf: LivyConf) extends Logging {
    */
   def checkImpersonation(
```
risdenk: @vanzin This is the method that is causing most of your concerns I think. It didn't change with `doAs`. Basically if impersonation is disabled, then it just falls through with no error.

server/src/test/scala/org/apache/livy/server/SessionServletSpec.scala
mgaido91: My point is: the focus of this PR is adding the parameter `doAs` in order to allow user impersonation. The authorization logic is orthogonal to this change. Adding tests for this situation I think is helpful in order to determine and ensure the current behavior, since they were missing.

If we decide that the logic/behavior we have currently for the authorization, I think we can change it adding some parameters in order to ensure users who upgrade can have no behavior change. But anyway, I think this should not be done in this PR, since the scope of this PR is to add the `doAs` parameter, not to change the impersonation and access control logic.

server/src/test/scala/org/apache/livy/server/interactive/JobApiSpec.scala
mgaido91: As I said before, I think that if we want to introduce a change in the behavior we currently have for user impersonation, we should do it in another JIRA/PR.

server/src/test/scala/org/apache/livy/server/SessionServletSpec.scala
vanzin: I'm fine with the "behavior changes in another PR" argument. But the test name here is so obviously wrong, that if someone else ends up touching this, they'll be scratching their heads.

So at the very least, the test name must be fixed.

risdenk: Updated the test name.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
```diff
@@ -230,6 +230,7 @@ class LivyThriftSessionManager(val server: LivyThriftServer, val livyConf: LivyC
         server.livySessionManager.nextId(),
         None,
         username,
+        None,
```
vanzin: Hmm, looks like the thrift side doesn't yet handle impersonation, so this is fine. Would probably be good to have a bug for that.

vanzin:
Merging to master.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
mgaido91: it does handle it but internally, as Hive server does, not using Livy impersonation (please see https://github.com/apache/incubator-livy/blob/47d3ee6b6555a63d0b871788a53aab022fff518a/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/cli/ThriftCLIService.scala#L265)

vanzin: Interesting; but at least now, it doesn't seem like that information is being reflected in Livy's state. Shouldn't matter much, but it's something to keep in mind.

PR #110, opened 2018-09-13
[LIVY-508][Server] Support custom auth filter for livy server

jerryshao:
## What changes were proposed in this pull request?

Users may have the requirement to add their own auth filter to the Livy server other than Kerberos, so here propose to add this support.

## How was this patch tested?

Verified locally in the cluster.

PR #142, opened 2019-01-22
[LIVY-552][WIP] Add JWTFilter to validate incoming JWT tokens

z-york:
This PR adds JWT Authentication to Livy via a new filter, JWTFilter that validates the JWT and accepts or rejects the request.

https://issues.apache.org/jira/browse/LIVY-552

## How was this patch tested?
Unit Tests passed.
Tested manually as well.

### DIFF
--- a/server/src/main/scala/org/apache/livy/LivyConf.scala
+++ b/server/src/main/scala/org/apache/livy/LivyConf.scala
@@ -168,6 +168,10 @@ object LivyConf {
   val THRIFT_DELEGATION_TOKEN_RENEW_INTERVAL =
     Entry("livy.server.thrift.delegation.token.renew-interval", "1d")
 
+  val JWT_FILTER_ENABLED = Entry("livy.jwt.filter.enabled", false)
+  val JWT_HEADER_NAME = Entry("livy.jwt.header.name", "Authentication")
+  val JWT_SIGNATURE_PUBLIC_KEY_PATH = Entry("livy.jwt.public-key.path", null)
+
   /**
    * Recovery mode of Livy. Possible values:
    * off: Default. Turn off recovery. Every time Livy shuts down, it stops and forgets all sessions.
--- a/pom.xml
+++ b/pom.xml
@@ -158,6 +158,8 @@
     <maven.compiler.source>${java.version}</maven.compiler.source>
     <maven.compiler.target>${java.version}</maven.compiler.target>
 
+    <nimbus-jwt.version>4.41.2</nimbus-jwt.version>
+
   </properties>
 
   <repositories>
--- a/server/pom.xml
+++ b/server/pom.xml
@@ -80,6 +80,12 @@
     </dependency>
 
     <dependency>
+      <groupId>com.nimbusds</groupId>
+      <artifactId>nimbus-jose-jwt</artifactId>
+      <version>${nimbus-jwt.version}</version>
+    </dependency>
+
+    <dependency>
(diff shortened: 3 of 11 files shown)


server/src/main/scala/org/apache/livy/LivyConf.scala
```diff
@@ -168,6 +168,10 @@ object LivyConf {
   val THRIFT_DELEGATION_TOKEN_RENEW_INTERVAL =
     Entry("livy.server.thrift.delegation.token.renew-interval", "1d")
 
+  val JWT_FILTER_ENABLED = Entry("livy.jwt.filter.enabled", false)
+  val JWT_HEADER_NAME = Entry("livy.jwt.header.name", "Authentication")
+  val JWT_SIGNATURE_PUBLIC_KEY_PATH = Entry("livy.jwt.public-key.path", null)
```
alex-the-man: Is there any chance we need to support multiple public keys for key rotation?

z-york: Do you mean allowing users to specify a list of public keys to workaround having to restart livy when a particular key is rotated? I think it's an interesting idea, but would require other changes to work (currently the key is read in once from the path and not refreshed, this would require some refresh mechanism). Anyways, I think this can be a separate feature request/JIRA after this one gets in.

alex-the-man:
LIVY-508 #110 will be merged soon and will provide a framework to add custom filters.
Would you mind rebasing your work on top of that? The only required changes should be all in LivyConf.

```
# If user wants to use custom authentication filter, configurations are:
# livy.server.auth.type = <custom>
# livy.server.auth.<custom>.class = <class of custom auth filter>
# livy.server.auth.<custom>.param.<foo1> = <bar1>
# livy.server.auth.<custom>.param.<foo2> = <bar2>
```

z-york:
Sure, let me try to rebase ontop of that PR

Tagar:
@z-york btw #110 is now merged to master.

z-york:
Thanks for the heads up @Tagar! @alex-the-man  I took a look at rebasing it ontop (and using) 110 and I think there are a few things that don't make sense with the current approach:

* It looks like only one auth_type can be specified which has a number of problems:
** Filters should be composable
** You should be able to use a custom filter/auth type with Kerberos

In addition to this, the JWTFilter in my PR requires LivyConf to be passed in when it is created which I *assume* won't work if it is being instantiated via the Filter Holder (not sure how this is being done, but I would assume reflection).

I think it should be fairly simple to change the config to a list of auth types which will solve the first problem, but I'm not sure about how to solve the 2nd issue. Any ideas?

alex-the-man:
@jerryshao What is your opinion on supporting multiple filters?

alex-the-man:
> I think it should be fairly simple to change the config to a list of auth types which will solve the first problem, but I'm not sure about how to solve the 2nd issue. Any ideas?

Those config will be passed to the function:
```
override def init(filterConfig: FilterConfig): Unit = {}
```
Could you change JWTFilter to pass those config to JWSVerifierProvider in that function call?

z-york:
My apologies for such a long lapse here... I will be picking this back up again.

naksh9619:
@z-york Are you planning to continue working on this? Otherwise I can pick up from where you left

z-york:
@naksh9619 Feel free to continue the work. I have since been pulled onto other work and didn't have the time to address follow-up comments.

PR #144, opened 2019-02-09
[LIVY-557] Make Travis-CI logs less verbose

meisam:
- Set the base travis image to ubuntu Xenial.
- Set the Scala version to `2.11.12` to match the version in pom.xml.
- Install `r-base` directly from xenial repos (no need to add r-packages).
- Combine the two `mvn` builds into one (`script` and `install` sections in `.travis.yml`).
- Make apt-get and pip are quiet.
- ~Set the SLF4J log level to warn for maven.~ 

## How was this patch tested?
Tested on Travis:
- Before this change, travis logs were about 6K+ lines long.
- After this change, the logs are less than 4K for unit tests and less than 3K for integration tests.
- I also verified that
  - Travis uses scala 2.11.12.
  - r-base is installed.
  - mvn runs without a problem after combining the two mvn builds.
  - Warn level messages and outputs of builds/tests are still available in the logs.
  - apt-get and pip run in quiet mode.

## Sample Travis outputs
**A build log that is  big to fit into Travis UI (size 4MB)**: https://travis-ci.org/apache/incubator-livy/jobs/487096239
**A build log BEFORE this change  (size 643KB)**: https://travis-ci.org/apache/incubator-livy/jobs/487164276
**A build log AFTER this change (size 327KB)**: https://travis-ci.org/apache/incubator-livy/jobs/492011690
**~A build log with mvn log level set to warn (323 KB)**: ~https://travis-ci.org/apache/incubator-livy/jobs/490937880~
This change is reverted because it was silencing useful information in mvn.

Task-url: https://issues.apache.org/jira/browse/LIVY-557

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,23 +50,23 @@ cache:
     - $HOME/.m2
 
 before_install:
-  - sudo apt-get -y install python3-pip python-dev
-  - sudo apt-get -y install libkrb5-dev
-  - sudo apt-get -y remove python-setuptools
-  - sudo pip2 install --upgrade "pip < 10.0.0" "setuptools < 36"
-  - sudo python3 -m pip install --upgrade "pip < 10.0.0" "setuptools < 36"
-  - sudo pip2 install codecov cloudpickle
-  - sudo python3 -m pip install cloudpickle
+  - sudo apt-get -q -y install python3-pip python-dev
+  - sudo apt-get -q -y install libkrb5-dev
+  - sudo apt-get -q -y remove python-setuptools
+  - sudo pip2 -q install --upgrade "pip < 10.0.0" "setuptools < 36"
+  - sudo python3 -m pip -q install --upgrade "pip < 10.0.0" "setuptools < 36"
+  - sudo pip2 -q install codecov cloudpickle
+  - sudo python3 -m pip -q install cloudpickle
 
 install:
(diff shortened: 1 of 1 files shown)


meisam:
Looks like the travis cache is stale.

vanzin:
To be frank I'm not sure I like the maven change. It removes the data about which module and phase are being executed from the output, which can be useful when investigating issues.

jerryshao:
Agree with @vanzin , this also hides lots of useful logs when debugging.

meisam:
I agree. Turning off maven info removes too much info, but I think we can make apt-get and pip quiet. Here is an example log: https://travis-ci.org/apache/incubator-livy/jobs/490913885

jerryshao:
I remembered that apt-get and pip log can be folded automatically, am I right?

meisam:
Yes. apt-get and pip logs auto-fold. My goal is to make the logs short enough so they fit into travis UI all the time. I think it is doable without silencing INFO logs in maven.

Here is an example of a log that does not fit (scroll all the way to the end): 
https://travis-ci.org/apache/incubator-livy/jobs/487096239
```
This log is too long to be displayed. Please reduce the verbosity of your build or download the raw log.
```

vanzin:
Merging to master.

PR #145, opened 2019-02-27
[LIVY-560] Added support for hostnames that resolve to an external address

smrtl:

In RSC context when the local hostname does not resolves to a loopback address, the canonical hostname is used and sent to other nodes.
This is a problem when running Livy on Kubernetes as another node won't be able to resolve the hostname. The proposed change is to return the IP address of the host when it's hostname resolves to an external IP.

Closes https://issues.apache.org/jira/browse/LIVY-560
Probably closes https://issues.apache.org/jira/browse/LIVY-372 too

## How was this patch tested?

This was tested by successfully running Livy with Spark 2.4.0 on Kubernetes 1.13.2

### DIFF
--- a/rsc/src/main/java/org/apache/livy/rsc/RSCConf.java
+++ b/rsc/src/main/java/org/apache/livy/rsc/RSCConf.java
@@ -141,6 +141,10 @@ public class RSCConf extends ClientConf<RSCConf> {
           }
         }
       }
+    } else {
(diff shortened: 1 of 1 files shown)


smrtl:
Closing this since it is covered by the great work done by @jahstreet, cf. https://github.com/apache/incubator-livy/pull/167, https://github.com/apache/incubator-livy/pull/167, https://github.com/apache/incubator-livy/pull/252

PR #146, opened 2019-03-01
LIVY-547: Livy kills session after livy.server.session.timeout even i…

shanyu:
…f the session is active

Signed-off-by: Shanyu Zhao <shzhao@microsoft.com>

To check for session expiration, if session state is busy, always return false.
https://issues.apache.org/jira/browse/LIVY-547

## How was this patch tested?

unit tests, integration tests and manual tests in a live cluster.

### DIFF
--- a/server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
+++ b/server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
@@ -150,6 +150,8 @@ class SessionManager[S <: Session, R <: RecoveryMetadata : ClassTag](
         case s: FinishedSessionState =>
           val currentTime = System.nanoTime()
           currentTime - s.time > sessionStateRetainedInSec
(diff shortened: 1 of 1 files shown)


server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
```diff
@@ -150,6 +150,8 @@ class SessionManager[S <: Session, R <: RecoveryMetadata : ClassTag](
         case s: FinishedSessionState =>
           val currentTime = System.nanoTime()
           currentTime - s.time > sessionStateRetainedInSec
+        case SessionState.Busy =>
```
yiheng: Add a unit test to cover this case？

mgaido91:
as mentioned in the other PR for this issue. I think we do need a new config for this: https://github.com/apache/incubator-livy/pull/138#issuecomment-457338491

PR #149, opened 2019-03-09
[LIVY-565] Fix the bug that sql query row count is less than the actual row count of the table at some condition

holmes-infra:
For example, there is a table with two columns ,which has two hundred rows ,and one of the column's data is all null .When the maxRows of sql query is 200 ,it will only return 100 rows.The cause is that the ColumnBuffer array ,only expands automatically when the inserted data size exceeds 100 , but does not expand if the one of column data is all null.

https://issues.apache.org/jira/browse/LIVY-565

## How was this patch tested?

### DIFF
--- a/thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ColumnBuffer.java
+++ b/thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ColumnBuffer.java
@@ -214,6 +214,8 @@ public class ColumnBuffer {
       // org.apache.hadoop.hive.serde2.thrift.ColumnBuffer expects a List<String>, so convert
       // when reading the value. The Hive/Thrift stack also dislikes nulls, and returning a
       // list with a different number of elements than expected.
+      strings=(strings.length != currentSize) ? Arrays.copyOfRange(strings, 0, currentSize)
+              : strings;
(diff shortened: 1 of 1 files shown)


thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ColumnBuffer.java
```diff
@@ -214,6 +214,8 @@ public Object getValues() {
       // org.apache.hadoop.hive.serde2.thrift.ColumnBuffer expects a List<String>, so convert
       // when reading the value. The Hive/Thrift stack also dislikes nulls, and returning a
       // list with a different number of elements than expected.
+      strings=(strings.length != currentSize) ? Arrays.copyOfRange(strings, 0, currentSize)
```
vanzin: nit: spaces around `=`.

I feel like this shouldn't change the internal state, like for the other cases in this switch. It's weird for a getter to change internal state.

This should also apply to the `BINARY` case above, right?

vanzin:
Hi, if you're not planning to address feedback, I'll close the PR and submit a new one. I think we should fix this before releasing 0.6.

vanzin:
I'm closing this for lack of response. I'll post my own fix.

PR #151, opened 2019-03-09
[LIVY-566] When cancel a sql job through livy, actually the sparkjob still running 

holmes-infra:
https://issues.apache.org/jira/browse/LIVY-566

## How was this patch tested?

### DIFF
--- a/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
+++ b/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
@@ -22,11 +22,9 @@ import java.util.concurrent.{ConcurrentLinkedQueue, RejectedExecutionException}
 
 import scala.collection.mutable
 import scala.util.control.NonFatal
-
 import org.apache.hadoop.security.UserGroupInformation
 import org.apache.hive.service.cli._
-
-import org.apache.livy.Logging
+import org.apache.livy.{JobHandle, Logging}
 import org.apache.livy.thriftserver.SessionStates._
 import org.apache.livy.thriftserver.operation.Operation
 import org.apache.livy.thriftserver.rpc.RpcClient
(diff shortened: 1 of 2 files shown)


thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/SqlJob.java
```diff
@@ -59,7 +59,7 @@ public SqlJob(
 
   @Override
   public Void call(JobContext ctx) throws Exception {
-    ctx.sc().setJobGroup(statementId, statement);
+    ctx.sc().setJobGroup(statementId, statement)
```
Tagar: Is setting job description as statement text related to LIVY-566? 
If yes, job description might need to be set to something else? as statement text can contain sensitive information. 
Thanks.

vanzin: The particular change you're commenting on is not part of things PR. In fact this PR does not compile, because this change is wrong...

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
```diff
@@ -60,6 +58,8 @@ class LivyExecuteStatementOperation(
 
   private def statementId: String = opHandle.getHandleIdentifier.toString
 
+  private var jobHandle: JobHandle[_]=_
```
vanzin: add spaces around `=`. Also later.

But really I think the correct fix would be in `ThriftSessionState.cleanupStatement`. Seems I missed to port some code from the old implementation when creating that class: https://github.com/apache/incubator-livy/commit/39fa887cfa482c5edaba2ad0c2af683d8e5c63c6#diff-712ea78474e0573e4dbe90b46f67ff05L160

vanzin:
Closing due to no reply. I'll post my own fix.

PR #152, opened 2019-03-12
[LIVY-567] fix SparkApp of InteractiveSession cannot be recovered

Jassy1994:

still create a SparkApp even driverProcess not exist, driverProcess only effects some logs, and we can get those log by visiting driverLogUrl.

## How was this patch tested?

recover livy server and observe in livy ui, validate whether sparkUiUrl can get from application. and  
whether livy session state changed when yarn state changed.

### DIFF
--- a/server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
+++ b/server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
@@ -383,7 +383,7 @@ class InteractiveSession(
   private val app = mockApp.orElse {
     val driverProcess = client.flatMap { c => Option(c.getDriverProcess) }
         .map(new LineBufferedProcess(_, livyConf.getInt(LivyConf.SPARK_LOGS_SIZE)))
-    driverProcess.map { _ => SparkApp.create(appTag, appId, driverProcess, livyConf, Some(this)) }
+    Option(SparkApp.create(appTag, appId, driverProcess, livyConf, Some(this)))
   }
 
   if (client.isEmpty) {


vanzin:
The unit test failure looks legit.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
```diff
@@ -392,8 +392,12 @@ class InteractiveSession(
     app = mockApp.orElse {
       val driverProcess = client.flatMap { c => Option(c.getDriverProcess) }
         .map(new LineBufferedProcess(_, livyConf.getInt(LivyConf.SPARK_LOGS_SIZE)))
-      driverProcess.map { _ => SparkApp.create(appTag, appId, driverProcess, livyConf, Some(this)) }
+    if (livyConf.isRunningOnYarn()) {
```
vanzin: The indentation here is wrong.

But this looks weird because in both cases, this is running on YARN, technically. You added this check because the unit test started failing when you tried the previous fix. Maybe the unit test needs fixing instead.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
```diff
@@ -392,8 +392,12 @@ class InteractiveSession(
...
+      Option(SparkApp.create(appTag, appId, driverProcess, livyConf, Some(this)))
+    } else {
+      driverProcess.map{ _ => SparkApp.create(appTag, appId, driverProcess, livyConf, Some(this)) }
```
vanzin: space after `map` (although I think this code can go away, see previous comment)

yiheng:
ping @Jassy1994 Are you still working on this PR?

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
Jassy1994: Thanks for this comment, I will modify UT instead

PR #155, opened 2019-03-14
[LIVY-565] Ensure column buffer for string and binary returns all data.

vanzin:
The code was failing to account for null values at the tail end of the buffer,
and could return a buffer that was too short for the number of rows being
returned.

Also added unit tests to verify the fix.

### DIFF
--- a/thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ColumnBuffer.java
+++ b/thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ColumnBuffer.java
@@ -18,9 +18,11 @@ package org.apache.livy.thriftserver.session;
 
 import java.math.BigDecimal;
 import java.nio.ByteBuffer;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.BitSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Spliterator;
 import java.util.Spliterators;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -41,8 +43,8 @@ import org.apache.spark.sql.types.StructField;
 public class ColumnBuffer {
 
   public static final int DEFAULT_SIZE = 100;
-  private static final String EMPTY_STRING = "";
-  private static final ByteBuffer EMPTY_BUFFER = ByteBuffer.wrap(new byte[0]);
+  public static final String EMPTY_STRING = "";
+  public static final ByteBuffer EMPTY_BUFFER = ByteBuffer.wrap(new byte[0]);
 
   private final DataType type;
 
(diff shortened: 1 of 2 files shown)


thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ColumnBuffer.java
```diff
@@ -241,6 +232,25 @@ private boolean isNull(int index) {
...
+   * org.apache.hadoop.hive.serde2.thrift.ColumnBuffer expects a List<String> or List<ByteBuffer>,
+   * depending on the column type. The Hive/Thrift stack also dislikes nulls, and returning a list
+   * with a different number of elements than expected.
+   */
+  private <T> List<T> toList(Stream<T> data, T defaultValue) {
+    final List<T> ret = new ArrayList<>(currentSize);
+    data.limit(currentSize).forEach(e -> {
+      ret.add(e != null ? e : defaultValue);
+    });
+    while (ret.size() < currentSize) {
```
mgaido91: Maybe we can rather do `ensureCapacity();` and `currentSize++;` before the null check in add?

vanzin: This uses less memory. e.g. if you have all nulls, the internal buffer would be empty, and you'd just allocate memory here, instead of allocating 2 `currentSize`-sized buffers.

mgaido91: Yes, makes sense.

thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ColumnBuffer.java
```diff
@@ -203,21 +205,10 @@ public Object getValues() {
...
         : doubles;
     case BINARY:
-      // org.apache.hadoop.hive.serde2.thrift.ColumnBuffer expects a List<ByteBuffer>, so convert
-      // when reading the value. The Hive/Thrift stack also dislikes nulls, and returning a
-      // list with a different number of elements than expected.
-      return Arrays.stream(buffers)
-        .limit(currentSize)
-        .map(b -> (b != null) ? ByteBuffer.wrap(b) : EMPTY_BUFFER)
-        .collect(Collectors.toList());
+      return toList(Arrays.stream(buffers).map(b -> (b != null) ? ByteBuffer.wrap(b) : null),
```
mgaido91: nit: we can maybe add a Function as argument about how to treat null, in order to avoid this map

vanzin: The map is not because of nulls, it's because the internal buffer is an array of `byte[]` and the Hive code expects `ByteBuffer`s, not byte arrays.

vanzin: I could move the null handling completely within the `map` if you prefer, instead of passing the default value as an argument to `toList`.

vanzin: err, actually I can't do that because I need to fill the missing items with the default value...

mgaido91: yes, sorry, I didn't explain well, I meant:
```
toList(buffers, b -> ByteBuffer.wrap(b), EMPTY_BUFFER)
```
here and on the other side
```
toList(strings, s -> s, EMPTY_STRING)
```
it seems a bit cleaner to be avoiding this map here, but just it is just a nit.

vanzin: Actually the `Stream<T>` argument was intentional, because you can't have a method that takes a "generic" array type...

mgaido91: I see, well then:
```
toList(Arrays.stream(buffer), b -> ByteBuffer.wrap(b), EMPTY_BUFFER)
```

vanzin: It may look a bit cleaner for the binary case, but it looks uglier for the string case.

Sorry, I'm not seeing much benefit in tweaking this internal API more.

vanzin:
Merging to master.

PR #161, opened 2019-03-21
[BUILD] Update release scripts with new features.

vanzin:
- The release build script now stages both the archives and maven
  artifacts with the same command, and includes some optimizations
  to avoid downloading too much stuff while doing that.

- A new update-version.sh script is added to update the Livy version
  when preparing an RC.


vanzin:
Merging to master.

PR #162, opened 2019-03-23
[LIVY-572] Avoid usage of spark classes in ColumnBuffer

mgaido91:

The `ColumnBuffers` can be created both inside spark jobs and in the Livy server. The latter case happens when operation logs are returned and  fails before this patch because we are using Spark classes in this code after the refactor in LIVY-503. Unfortunately, we do not have test coverage for operation logs retrieval and this is the reason why this wasn't spot out earlier.

Since operation logs are retrieved by beeline for each query, this means that every query run through beeline fails, unless `livy.server.thrift.logging.operation.enabled` is set to `false`.

## How was this patch tested?

manual tests using beeline

### DIFF
--- a/thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ResultSet.java
+++ b/thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ResultSet.java
@@ -39,21 +39,17 @@ public class ResultSet {
 
   private final String schema;
   private final ColumnBuffer[] columns;
-  private final boolean[] columnIsString;
 
   public ResultSet() {
     this.schema = null;
     this.columns = null;
-    this.columnIsString = null;
   }
 
   public ResultSet(DataType[] types, String schema) {
     this.schema = schema;
     this.columns = new ColumnBuffer[types.length];
-    this.columnIsString = new boolean[types.length];
     for (int i = 0; i < columns.length; i++) {
       columns[i] = new ColumnBuffer(types[i]);
-      columnIsString[i] = types[i].equals(DataType.STRING);
     }
   }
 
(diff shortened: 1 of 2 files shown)


thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ResultSet.java
```diff
@@ -43,7 +63,13 @@ public void addRow(Object[] fields) {
     }
 
     for (int i = 0; i < fields.length; i++) {
-      columns[i].add(fields[i]);
+      Object value;
+      if (columnIsString[i]) {
```
vanzin: Checking `columns[i].getType().equals(DataType.STRING)` here avoids the extra boolean array.

(It's also an enum so you could use `==`.)

mgaido91: Yes, the reason why I did so was to avoid doing the check for every field of each row and hence avoid introducing an overhead in a part of the code which is critical for performances. Anyway, if you feel strong about this I can change it: it is also true that the overhead is probably very small compared to all the operations which we need to perform, e.g. transfer the data over the network...

vanzin: Using `==` there's actually no more overhead at all. In fact, if you look at CPU caches and things like that it's probably faster since there's less risk of a cache miss...

mgaido91: ok, done, thanks.

vanzin:
Merging to master.

mgaido91:
Thanks @vanzin! What about backporting to 0.6?

vanzin:
If there's a need for another RC I'll do it, but we have never released a .z release before, I'd rather just do 0.7 instead.

PR #165, opened 2019-04-03
[LIVY-581] Fix edge-case where livy overrides user-provided spark properties instead of appending

Kim-Hammar:
This commit fixes a bug in livy that yarn properties are overridden when hive is enabled in the spark sessions.

Bugfix, append livy spark properties to the correct spark property (spark.yarn.X or spark.X. If the user have supplied spark config in spark.yarn.X and livy sets spark.X, spark will ignore spark.yarn.X which causes undesired behaviour since the user properties are then ignored. 
 
https://issues.apache.org/jira/browse/LIVY-581

## How was this patch tested?
Unit tests pass and this patch have been in production in our cluster since November last year

### DIFF
--- a/server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
+++ b/server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
@@ -290,8 +290,7 @@ object InteractiveSession extends Logging {
                           .map(_.split(",")).getOrElse(Array.empty[String])
       val sparkFiles = conf.get(LivyConf.SPARK_FILES)
                            .map(_.split(",")).getOrElse(Array.empty[String])
-      var files = Array.empty[String]
-      if (!sparkFiles.isEmpty || yarnFiles.isEmpty) files = sparkFiles else files = yarnFiles
+      val files = if (!sparkFiles.isEmpty || yarnFiles.isEmpty) sparkFiles else yarnFiles
       hiveSiteFile(files, livyConf) match {
         case (_, true) =>
           debug("Enable HiveContext because hive-site.xml is found in user request.")
--- a/server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
+++ b/server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
@@ -286,16 +286,33 @@ object InteractiveSession extends Logging {
     }
 
     def mergeHiveSiteAndHiveDeps(sparkMajorVersion: Int): Unit = {
-      val sparkFiles = conf.get("spark.files").map(_.split(",")).getOrElse(Array.empty[String])
-      hiveSiteFile(sparkFiles, livyConf) match {
+      val yarnFiles = conf.get(LivyConf.SPARK_YARN_DIST_FILES)
+                          .map(_.split(",")).getOrElse(Array.empty[String])
+      val sparkFiles = conf.get(LivyConf.SPARK_FILES)
+                           .map(_.split(",")).getOrElse(Array.empty[String])
+      var files = Array.empty[String]
+      if (!sparkFiles.isEmpty || yarnFiles.isEmpty) files = sparkFiles else files = yarnFiles
+      hiveSiteFile(files, livyConf) match {
         case (_, true) =>
           debug("Enable HiveContext because hive-site.xml is found in user request.")
-          mergeConfList(datanucleusJars(livyConf, sparkMajorVersion), LivyConf.SPARK_JARS)
+          if (conf.get(LivyConf.SPARK_JARS) != None || conf.get(LivyConf.SPARK_YARN_JARS) == None) {
+            mergeConfList(datanucleusJars(livyConf, sparkMajorVersion), LivyConf.SPARK_JARS)
+          } else {
+            mergeConfList(datanucleusJars(livyConf, sparkMajorVersion), LivyConf.SPARK_YARN_JARS)
+          }
         case (Some(file), false) =>
           debug("Enable HiveContext because hive-site.xml is found under classpath, "
             + file.getAbsolutePath)
-          mergeConfList(List(file.getAbsolutePath), LivyConf.SPARK_FILES)
-          mergeConfList(datanucleusJars(livyConf, sparkMajorVersion), LivyConf.SPARK_JARS)
+          if (!sparkFiles.isEmpty || yarnFiles.isEmpty) {
+            mergeConfList(List(file.getAbsolutePath), LivyConf.SPARK_FILES)
+          } else {
+            mergeConfList(List(file.getAbsolutePath), LivyConf.SPARK_YARN_DIST_FILES)
+          }
+          if (conf.get(LivyConf.SPARK_JARS) != None || conf.get(LivyConf.SPARK_YARN_JARS) == None) {
+            mergeConfList(datanucleusJars(livyConf, sparkMajorVersion), LivyConf.SPARK_JARS)
+          } else {
+            mergeConfList(datanucleusJars(livyConf, sparkMajorVersion), LivyConf.SPARK_YARN_JARS)
+          }
         case (None, false) =>
           warn("Enable HiveContext but no hive-site.xml found under" +
             " classpath or user request.")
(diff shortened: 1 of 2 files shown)


mgaido91:
actually the CI is failing...there are test failures...

yiheng:
We meet the same issue... Why close this patch?

Limmen:
I closed it because I've asked twice on the mailing-list with no response at all.. we maintain our own fork and this patch works fine.

Limmen:
https://www.mail-archive.com/user@livy.incubator.apache.org/msg00358.html

Limmen:
I remember unit tests pass locally, still the CI failed for some reason

mgaido91:
@Limmen the CI was failing as stated in my comment. First we need the CI to pass to review a patch. If you want to reopen it and make the CI passing, this patch can be reviewed.

Limmen:
@mgaido91 CI passes now, it was an issue with scalastyle

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
```diff
@@ -286,16 +286,33 @@ object InteractiveSession extends Logging {
...
 
     def mergeHiveSiteAndHiveDeps(sparkMajorVersion: Int): Unit = {
-      val sparkFiles = conf.get("spark.files").map(_.split(",")).getOrElse(Array.empty[String])
-      hiveSiteFile(sparkFiles, livyConf) match {
+      val yarnFiles = conf.get(LivyConf.SPARK_YARN_DIST_FILES)
+                          .map(_.split(",")).getOrElse(Array.empty[String])
+      val sparkFiles = conf.get(LivyConf.SPARK_FILES)
+                           .map(_.split(",")).getOrElse(Array.empty[String])
+      var files = Array.empty[String]
+      if (!sparkFiles.isEmpty || yarnFiles.isEmpty) files = sparkFiles else files = yarnFiles
```
yiheng: This code can be refactored like this
```scala
val files = if (!sparkFiles.isEmpty || yarnFiles.isEmpty) sparkFiles else yarnFiles
```

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
```diff
@@ -286,16 +286,33 @@ object InteractiveSession extends Logging {
...
+      hiveSiteFile(files, livyConf) match {
         case (_, true) =>
           debug("Enable HiveContext because hive-site.xml is found in user request.")
-          mergeConfList(datanucleusJars(livyConf, sparkMajorVersion), LivyConf.SPARK_JARS)
+          if (conf.get(LivyConf.SPARK_JARS) != None || conf.get(LivyConf.SPARK_YARN_JARS) == None) {
```
yiheng: I feel like it's not so straight forward here. In the JIRA description:
>Now what seems to happen is that if all of spark.files, spark.jars, spark.yarn.dist.files, and spark.yarn.jars are non-null when the job is submitted (spark.files spark.jars filled in by livy and spark.yarn.dist.files spark.yarn.jars filled in by the user-request from our platform),spark.yarn.dist.files gets set to spark.files and spark.yarn.jars gets set to spark.jars

Should we change the `set` behavior to `merge` where the job is submitted?

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
Limmen: Good point, done.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
Limmen: @yiheng can you elaborate? I am not sure I understand what you suggest. You mean to change the way that the nucleus jars are merged with the user-provided spark properties?The behavior is that if both spark.jars and spark.yarn.jars are set, spark ignores the yarn-property.

Limmen: ping @yiheng

jerryshao: Yes, the condition here `not A || B` seems not so straightforward, I would suggest to simplify or add some comments here and below to explain the purpose of such check.

yiheng: @Limmen Sorry for missing the message.
> The behavior is that if both spark.jars and spark.yarn.jars are set, spark ignores the yarn-property.

If it's spark behavior, it's fine to deal with it like this.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
```diff
@@ -332,7 +348,11 @@ object InteractiveSession extends Logging {
       LivySparkUtils.formatSparkVersion(livyConf.get(LivyConf.LIVY_SPARK_VERSION))
     val scalaVersion = livyConf.get(LivyConf.LIVY_SPARK_SCALA_VERSION)
 
-    mergeConfList(livyJars(livyConf, scalaVersion), LivyConf.SPARK_JARS)
+    if (conf.get(LivyConf.SPARK_JARS) != None || conf.get(LivyConf.SPARK_JARS) == None){
```
yiheng: Is there a typo here? Should it be
```scala
if (conf.get(LivyConf.SPARK_JARS) != None || conf.get(LivyConf. SPARK_YARN_JARS) == None)
```

yiheng: As that condition appears many times, I suggest do it like this
```scala
// Detect which config we should use and merge jars to that config.
// Please notice that if we set both, the yarn_ config will be ignored by Spark
val mergeJarsConf = if (conf.get(LivyConf.SPARK_JARS) != None || conf.get(LivyConf. SPARK_YARN_JARS) == None) LivyConf.SPARK_JARS else SPARK_YARN_JARS

mergeConfList(datanucleusJars(livyConf, sparkMajorVersion), mergeJarsConf)
...
mergeConfList(datanucleusJars(livyConf, sparkMajorVersion), mergeJarsConf)
...
mergeConfList(livyJars(livyConf, scalaVersion), mergeJarsConf)
```

yiheng: For the spark file conf
```scala
val (files, mergeFileConf) = if (!sparkFiles.isEmpty || yarnFiles.isEmpty) {
  (sparkFiles, LivyConf.SPARK_FILES)
} else {
  (yarnFiles, LivyConf.SPARK_YARN_DIST_FILES)
}
...
mergeConfList(List(file.getAbsolutePath), mergeFileConf)
```

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
```diff
@@ -285,17 +285,17 @@ object InteractiveSession extends Logging {
       }
     }
 
-    def mergeHiveSiteAndHiveDeps(sparkMajorVersion: Int): Unit = {
-      val sparkFiles = conf.get("spark.files").map(_.split(",")).getOrElse(Array.empty[String])
-      hiveSiteFile(sparkFiles, livyConf) match {
+    def mergeHiveSiteAndHiveDeps(sparkMajorVersion: Int, files: Array[String],
+                                 mergeJarsConf: String, mergeFilesConf: String): Unit = {
```
yiheng: indent?

yiheng:
This patch looks good to me except a minor format issue.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
Limmen: oops, will fix in a few hours, busy right now.

Limmen:
@yiheng CI pass now

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
jerryshao: Please change to "one parameter per line", thanks.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
```diff
@@ -332,6 +332,25 @@ object InteractiveSession extends Logging {
...
     val scalaVersion = livyConf.get(LivyConf.LIVY_SPARK_SCALA_VERSION)
 
+    // Detect which config we should use and merge jars and files to that config.
+    // Please notice that if we set both spark.X config values and the spark.yarn.X config values
+    // the spark.yarn.X config will be ignored by Spark
+    val yarnFiles = conf.get(LivyConf.SPARK_YARN_DIST_FILES)
+      .map(_.split(",")).getOrElse(Array.empty[String])
+    val sparkFiles = conf.get(LivyConf.SPARK_FILES)
+      .map(_.split(",")).getOrElse(Array.empty[String])
+    val mergeJarsConf = if (conf.get(LivyConf.SPARK_JARS) != None ||
```
jerryshao: You can change to `isDefined` or `isEmpty` other than comparing `None`.

PR #166, opened 2019-04-04
[LIVY-585][SERVER] Avoid duplicate stopping log when stop/interrupt a session

Ngone51:

There's duplicate stopping log when stop/interrupt a session. This happens because server logs stopping info when we stopping/interrupting a session at the first time. And when the session expired, it would be removed from session manager and call stop() and logs stopping info for the second time(Though, it does not really stop the session again at this time).

dup-stopping-log

The fix is quite simple by adding a `_stopped` flag in Session's `stop()` method. 
 
## How was this patch tested?

N.A.(?)

### DIFF
--- a/server/src/main/scala/org/apache/livy/sessions/Session.scala
+++ b/server/src/main/scala/org/apache/livy/sessions/Session.scala
@@ -185,32 +185,32 @@ abstract class Session(
   def start(): Unit
 
   def stop(): Future[Unit] = Future {
-    try {
-      if (_stopped.compareAndSet(false, true)) {
+    if (_stopped.compareAndSet(false, true)) {
+      try {
         info(s"Stopping $this...")
         stopSession()
         info(s"Stopped $this.")
+      } catch {
+        case e: Exception =>
+          warn(s"Error stopping session $id.", e)
       }
-    } catch {
-      case e: Exception =>
-        warn(s"Error stopping session $id.", e)
-    }
 
-    try {
-      if (stagingDir != null) {
-        debug(s"Deleting session $id staging directory $stagingDir")
-        doAsOwner {
-          val fs = FileSystem.newInstance(livyConf.hadoopConf)
-          try {
-            fs.delete(stagingDir, true)
-          } finally {
-            fs.close()
+      try {
+        if (stagingDir != null) {
+          debug(s"Deleting session $id staging directory $stagingDir")
+          doAsOwner {
+            val fs = FileSystem.newInstance(livyConf.hadoopConf)
+            try {
+              fs.delete(stagingDir, true)
+            } finally {
+              fs.close()
+            }
           }
         }
+      } catch {
+        case e: Exception =>
+          warn(s"Error cleaning up session $id staging dir.", e)
       }
-    } catch {
-      case e: Exception =>
-        warn(s"Error cleaning up session $id staging dir.", e)
     }
   }
 
--- a/server/src/main/scala/org/apache/livy/sessions/Session.scala
+++ b/server/src/main/scala/org/apache/livy/sessions/Session.scala
@@ -24,9 +24,11 @@ import java.util.UUID
 import java.util.concurrent.atomic.AtomicBoolean
 
 import scala.concurrent.{ExecutionContext, Future}
+
 import org.apache.hadoop.fs.{FileSystem, Path}
 import org.apache.hadoop.fs.permission.FsPermission
 import org.apache.hadoop.security.UserGroupInformation
+
 import org.apache.livy.{LivyConf, Logging, Utils}
 import org.apache.livy.utils.AppInfo
 
@@ -211,6 +213,8 @@ abstract class Session(
         case e: Exception =>
           warn(s"Error cleaning up session $id staging dir.", e)
       }
+    } else {
+      debug(s"Redundant call on stop(), ${toString()} has been already stopped.")
     }
   }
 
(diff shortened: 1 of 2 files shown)


Ngone51:
@vanzin  please have a look, thx.

server/src/main/scala/org/apache/livy/sessions/Session.scala
```diff
@@ -21,13 +21,12 @@ import java.io.InputStream
 import java.net.{URI, URISyntaxException}
 import java.security.PrivilegedExceptionAction
 import java.util.UUID
+import java.util.concurrent.atomic.AtomicBoolean
 
 import scala.concurrent.{ExecutionContext, Future}
-
```
mgaido91: these are causing scalastyle violations..

server/src/main/scala/org/apache/livy/sessions/Session.scala
```diff
@@ -184,30 +185,32 @@ abstract class Session(
...
           }
         }
+      } catch {
+        case e: Exception =>
+          warn(s"Error cleaning up session $id staging dir.", e)
       }
-    } catch {
-      case e: Exception =>
-        warn(s"Error cleaning up session $id staging dir.", e)
     }
```
mgaido91: I think it would be convenient to add a debug log in case the session is already stopped in order to track that this method is called multiple times.

server/src/main/scala/org/apache/livy/sessions/Session.scala
```diff
@@ -184,30 +185,32 @@ abstract class Session(
...
   def stop(): Future[Unit] = Future {
-    try {
-      info(s"Stopping $this...")
-      stopSession()
-      info(s"Stopped $this.")
-    } catch {
-      case e: Exception =>
-        warn(s"Error stopping session $id.", e)
-    }
+    if (_stopped.compareAndSet(false, true)) {
```
mgaido91: I am not sure about this. What about using the `state`? If there is an error while stopping the session, we might want to re-try to stop it, and this doesn't allow that..

Ngone51: > What about using the `state`? 

Previously, I used `state`, which probably likes *if the `state` is already `FinishedSessionState`, do not do stop again*. But I found there's code(see below) which transit `state` to `FinishedSessionState` first before calling stop(). So, this strategy can not work, then I try to use atomic boolean here.

```
if (serverSideState != SessionState.ShuttingDown) {
  transition(SessionState.Error())
  stop()
  app.foreach { a =>
     info(s"Failed to ping RSC driver for session $id. Killing application.")
        a.kill()
     }
}
```

> we might want to re-try to stop it, and this doesn't allow that..

Yeah, this can be a problem if we want to retry... Do we have some retry mechanism in some place by now ?

Ngone51:
@mgaido91 Thank you for your review. I've updated it.

server/src/main/scala/org/apache/livy/sessions/Session.scala
mgaido91: well, if we are calling the stop method several times, we are actually trying to perform this method multiple times. At very least, I think we should set the flag to true only at the end, once we are sure the session is closed. Still, iof the only advantage here is reducing the log lines, I am not sure whether the effort, the synchronization and the logic added are worth the very little gain we have. Are there any problems caused by these multiple invocations?

Ngone51: > well, if we are calling the stop method several times, we are actually trying to perform this method multiple times. 

Yeah, we're calling stop() for several times from current code, but I think this what we do not expect(e.g.  in `collectGarbage()` we may do not want to call stop() again for already finished session.) and this `several times` still remains different with `retries`.

> At very least, I think we should set the flag to true only at the end, once we are sure the session is closed. 

This is a good idea. I'll try it. Thanks.

> Are there any problems caused by these multiple invocations?

Nothing big deal, but some trivial things, e.g. duplicate logs, transiting an already Dead session to ShuttingDown state and then to Dead state again.

server/src/main/scala/org/apache/livy/sessions/Session.scala
```diff
@@ -184,30 +187,35 @@ abstract class Session(
...
   def stop(): Future[Unit] = Future {
-    try {
-      info(s"Stopping $this...")
-      stopSession()
-      info(s"Stopped $this.")
-    } catch {
-      case e: Exception =>
-        warn(s"Error stopping session $id.", e)
-    }
+    if (!_stopped.get()) {
```
mgaido91: now we should sync the whole block otherwise we may have 2 threads entering this, right?

Ngone51:
> If we are calling stop when it is not necessary, I think we should rather avoid calling it in those cases.

I was thinking about that way, but I was not sure whether it's the only place where calling stop unnecssary. So, I think it may be better to cover those cases(if any) in method stop (just like the thing we do in `IteractiveSession.stopSession`).

server/src/main/scala/org/apache/livy/sessions/Session.scala
Ngone51: Oh, yes. And this recalls me why I put `_stopped.compareAndSet(false, true)` in if statement firstly...

mgaido91:
>  I think it may be better to cover those cases(if any) in method stop

I don't really agree with you. Let me cc @vanzin for his opinion too on this.

PR #168, opened 2019-04-09
[LIVY-563] Propagate RSC configuration when creating sessions.

vanzin:
Even though not all RSC configs apply to the remote driver, a few do,
so propagate all of them when starting a new session.

Includes new unit test.


vanzin:
Merging to master.

PR #180, opened 2019-07-15
[LIVY-582][Tests] Hostname in python-api test should be lower case to avoid test failures

yiheng:
In the python-API test code, when returning a mocked response, the mock lib will compare the URL with the predefined URLs case sensitive. However, the `Request` lib used in the Livy python API will change the URL to lower case. This will cause test failures on a machine with an upper case hostname.

This patch turns the hostname in python-API test code into the lower case to avoid such test failures.

## How was this patch tested?
Existing test. Run test specifically on a machine with an upper case hostname.


jerryshao:
Merging to master branch, thanks for the fix!

PR #182, opened 2019-07-17
[LIVY-571] cleanupStatement should not throw exception when statementId not exist

yantzu:

`ThriftSessionState` is used to store query result by statementId. When an exception is thrown during execute query, no query result is stored. But when a statement is closed from beeline, a request is invoked to remove cached query result in `ThriftSessionState`.

Remove statement query result from `ThriftSessionState` currently throws an exception, hiding the original query failure exception and message. The PR makes it return cleanly, so the proper exception is reported to the end user.

## How was this patch tested?

Added unit tests.

### DIFF
--- a/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
+++ b/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
@@ -177,7 +177,7 @@ class LivyExecuteStatementOperation(
       val cleaned = rpcClient.cleanupStatement(sessionHandle, statementId).get()
       if (!cleaned) {
         warn(s"Fail to cleanup query $statementId (session = ${sessionHandle.getSessionId}), " +
-          s"can be ignored if query is failed")
+          s"this message can be ignored if the query failed.")
       }
     }
     setState(state)
--- a/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
+++ b/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
@@ -174,7 +174,11 @@ class LivyExecuteStatementOperation(
 
   private def cleanup(state: OperationState) {
     if (statementId != null && rpcClientValid) {
-      rpcClient.cleanupStatement(sessionHandle, statementId).get()
+      val cleaned = rpcClient.cleanupStatement(sessionHandle, statementId).get()
+      if (!cleaned) {
+        warn(s"Fail to cleanup query $statementId (session = ${sessionHandle.getSessionId}), " +
+          s"can be ignored if query is failed")
+      }
     }
     setState(state)
   }
--- a/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
+++ b/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
@@ -177,7 +177,7 @@ class LivyExecuteStatementOperation(
       val cleaned = rpcClient.cleanupStatement(sessionHandle, statementId).get()
       if (!cleaned) {
         warn(s"Fail to cleanup query $statementId (session = ${sessionHandle.getSessionId}), " +
-          s"this message can be ignored if the query failed.")
+          "this message can be ignored if the query failed.")
       }
     }
     setState(state)
--- a/thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
+++ b/thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
@@ -151,7 +151,7 @@ class BinaryThriftServerSuite extends ThriftServerBaseTest with CommonThriftTest
   }
 
   test("LIVY-571: returns a meaningful exception when global_temp table doesn't exist") {
-    withJdbcConnection(jdbcUri("default")) { c =>
+    withJdbcConnection { c =>
       val caught = intercept[SQLException] {
         val statement = c.createStatement()
         try {
--- a/thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
+++ b/thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
@@ -134,6 +134,29 @@ class BinaryThriftServerSuite extends ThriftServerBaseTest with CommonThriftTest
       statement.close()
     }
   }
+
+  test("invalid sql") {
+    assume(hiveSupportEnabled(formattedSparkVersion._1, livyConf))
+    withJdbcConnection(jdbcUri("default")) { c =>
+      var exceptionThrew = false
+
+      val statement = c.createStatement()
+      try {
+        try {
+          statement.executeQuery("use invalid_database")
+        } finally {
+          statement.close()
+        }
+      } catch {
+        case t: Throwable => {
+          exceptionThrew = true
+          assert(t.getMessage.contains("Database 'invalid_database' not found"))
+        }
+      }
+
+      assert(exceptionThrew)
+    }
+  }
 }
 
 class HttpThriftServerSuite extends ThriftServerBaseTest with CommonThriftTests {
--- a/thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
+++ b/thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
@@ -151,7 +151,6 @@ class BinaryThriftServerSuite extends ThriftServerBaseTest with CommonThriftTest
   }
 
   test("LIVY-571: returns a meaningful exception when global_temp table doesn't exist") {
-    assume(hiveSupportEnabled(formattedSparkVersion._1, livyConf))
     withJdbcConnection(jdbcUri("default")) { c =>
       val caught = intercept[SQLException] {
         val statement = c.createStatement()
--- a/thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
+++ b/thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
@@ -17,7 +17,7 @@
 
 package org.apache.livy.thriftserver
 
-import java.sql.{Date, Statement}
+import java.sql.{Date, SQLException, Statement}
 
 import org.apache.livy.LivyConf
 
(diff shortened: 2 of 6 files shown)


mgaido91:
HI @yantzu, thank you for your PR! I don't see how this PR fixes the JIRA issue referenced here. What is the behavior now? I think just nothing is returned, instead of the related exception... Could you add in the PR description some screenshot of the behavior before and after the PR? Moreover, could you add some UTs in order to enforce this behavior? Thanks.

yantzu:
Hi @mgaido91 , the rpc call from beeline when execute a query is:
1, send sql to thriftserver, beeline code: HiveStatement.runAsyncOnServer --> thriftserver code: ThriftCLIService.ExecuteStatement
2, get sql execute status from thriftserver, beeline code: HiveStatement.waitForOperationToComplete --> thriftserver code: ThriftCLIService.GetOperationStatus
3, get result from thriftserver if step 2 result success
4, close sql execute statement, beeline code: HiveStatement.close(called by Commands.executeInternal) --> thriftserver code: ThriftCLIService.CloseOperation

When exception happens running a query, step 2, thrift server will return exception error message to beeline, and beeline will throw an exception to caller. However, beeline will execute step 4 regardless of the result of step 2. When step 4 request send to thriftserver, it will send a request to Spark to remove statement result but as I mentioned there is no result in ThriftSessionState, so exception is thrown to beeline, this new exception conceal the exception of step 2. 
My fix is just to avoid exception of step 4.

I have one IT to cover such case.

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -134,6 +134,29 @@ class BinaryThriftServerSuite extends ThriftServerBaseTest with CommonThriftTest
       statement.close()
     }
   }
+
+  test("invalid sql") {
+    assume(hiveSupportEnabled(formattedSparkVersion._1, livyConf))
```
mgaido91: I think we can create an example which doesn't require hive support in order to have better test coverage, so we can get rid of this

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -134,6 +134,29 @@ class BinaryThriftServerSuite extends ThriftServerBaseTest with CommonThriftTest
...
+    withJdbcConnection(jdbcUri("default")) { c =>
+      var exceptionThrew = false
```
mgaido91: we can avoid this `var` using `intercept`

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -134,6 +134,29 @@ class BinaryThriftServerSuite extends ThriftServerBaseTest with CommonThriftTest
       statement.close()
     }
   }
+
+  test("invalid sql") {
```
mgaido91: please add the JIRA number and provide a better description
```suggestion
  test("LIVY-571: Invalid query returns a meaningful exception") {
```

thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ThriftSessionState.java
```diff
@@ -26,6 +26,9 @@
 import org.apache.spark.sql.SparkSession;
 import org.apache.spark.sql.types.StructType;
```
mgaido91: please remove this blank line

thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ThriftSessionState.java
```diff
@@ -104,9 +109,10 @@ StatementState findStatement(String statementId) {
   void cleanupStatement(String statementId) {
     checkNotNull(statementId, "No statement ID.");
     if (statements.remove(statementId) == null) {
-      throw statementNotFound(statementId);
+      LOG.warn("Statement {} not found in session {}",statementId, sessionId);
```
mgaido91: This code is execute on the spark side. What about throwing an exception here and catching in the Livy server side?

yantzu:
Before:
image

After:
image

thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ThriftSessionState.java
yantzu: I did consider other solution like catch exception in livy or don't CleanupStatementJob to spark when livy know it is a ERROR statement. But I think it make sense to change cleanupStatement, it like remove method of Map, or close method of InputStream, not matter how many times it is invoked, should always return success.

Other comments are fixed, and thanks a lot for your great suggestion!

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
mgaido91: @yantzu this is not resolved. Please use a query which doesn't require hive support so that we have wider coverage.

thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ThriftSessionState.java
mgaido91: The point is that this exception is logged on the spark job and not in the livy server, which is not what we want. We can maybe return a boolean here and check it on livy server... or throw the exception and catch it... but logging here makes no sense

yantzu: If we take ThriftSessionState as something like Map or InputStream, then it is not an exception. I think we can consider removing the log if you concern about it. On the other side, if we catch the exception in livy, we need to handle it carefully because there may be some other reason result in CleanupStatementJob failure, we need to distinguish them in livy code, which may make the code complex.  How do you think?

mgaido91: What about returning a boolean and give a boolean back to Livy server?

yantzu: I will look at if can returning a boolean. But I may have no time this week,  next week should be good. ^_^

mgaido91: no worries, thanks for your effort

mgaido91:
LGTM, cc @vanzin too. If I won't hear any comment in a couple of days I'll merge it. Thanks.

mgaido91:
@yantzu sorry, may you please improve the description of the PR? I'll merge once it is updated, thanks. I suggest:

```

`ThriftSessionState` is used to store query result by statementId. When an exception is thrown during execute query, no query result is stored. But when a statement is closed from beeline, a request is invoked to remove cached query result in `ThriftSessionState`.

Remove statement query result from `ThriftSessionState` currently throws an exception, hiding the original query failure exception and message. The PR makes it return cleanly, so the proper exception is reported to the end user.

## How was this patch tested?

Added unit tests.
```

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
mgaido91: moreover, this is yet to be addressed before I can merge this. Please do edit this UT so it doesn't need hive support and we have wider test coverage

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
```diff
@@ -174,7 +174,11 @@ class LivyExecuteStatementOperation(
 
   private def cleanup(state: OperationState) {
     if (statementId != null && rpcClientValid) {
-      rpcClient.cleanupStatement(sessionHandle, statementId).get()
+      val cleaned = rpcClient.cleanupStatement(sessionHandle, statementId).get()
+      if (!cleaned) {
+        warn(s"Fail to cleanup query $statementId (session = ${sessionHandle.getSessionId}), " +
+          s"can be ignored if query is failed")
```
mgaido91: nit:
```suggestion
          "this message can be ignored if the query failed.")
```

yantzu: Hi，I summit another commit. For the UT, I just remove the hive support check code. I am not sure whether it is you concern. Because as I my understanding, all test currently are running without hive context.  On the other hand, with/without hive context are happening on Spark internal code, which should not impact on Livy code.

yantzu: Is the build failed relevant?

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
```diff
@@ -174,7 +174,11 @@ class LivyExecuteStatementOperation(
 
   private def cleanup(state: OperationState) {
     if (statementId != null && rpcClientValid) {
-      rpcClient.cleanupStatement(sessionHandle, statementId).get()
+      val cleaned = rpcClient.cleanupStatement(sessionHandle, statementId).get()
+      if (!cleaned) {
+        warn(s"Fail to cleanup query $statementId (session = ${sessionHandle.getSessionId}), " +
+          s"this message can be ignored if the query failed.")
```
mgaido91:
```suggestion
          "this message can be ignored if the query failed.")
```

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -134,6 +134,35 @@ class BinaryThriftServerSuite extends ThriftServerBaseTest with CommonThriftTest
...
+        } finally {
+          statement.close()
+        }
+      }
+      assert(caught.getMessage.contains("Database 'invalid_database' not found"))
+    }
+  }
+
+  test("LIVY-571: returns a meaningful exception when global_temp table doesn't exist") {
+    withJdbcConnection(jdbcUri("default")) { c =>
```
mgaido91:
```suggestion
    withJdbcConnection { c =>
```

yantzu: Thanks! New commit created, and all tests passed.

mgaido91:
Thanks. Merging to master.

mgaido91:
Thanks for your first contribution @yantzu. May you please give me your JIRA username so I can assign the JIRA to you? Thanks.

yantzu:
Username: xilangyan . Thanks a lot!

mgaido91:
thank you, done!

PR #185, opened 2019-07-22
LIVY-612 Add Sessions Connect Documentation

patricker:

Documentation was missing for the Sessions Connect REST end point.
This came up as a work around for https://issues.apache.org/jira/browse/LIVY-547 in the discussion on https://github.com/apache/incubator-livy/pull/138.  It would be good to document it.

https://issues.apache.org/jira/browse/LIVY-612

## How was this patch tested?

Not Tested.

### DIFF
--- a/docs/rest-api.md
+++ b/docs/rest-api.md
@@ -242,6 +242,18 @@ Gets the log lines from this session.
   </tr>
 </table>
 
+### POST /sessions/{sessionId}/connect
+
+Connects to an existing session and updates its last activity time. This stops Livy from expiring a session due to inactivity.
+
+#### Request Parameters
+
+No Request Parameters
+
+#### Response Body
+
+The response body is the same as for `GET /sessions/{sessionId}`.
+
 ### GET /sessions/{sessionId}/statements
 
 Returns all the statements in a session.


jerryshao:
IIRC, this API and some others are only used for Livy client API, not for REST interaction. Seems not so necessary to document it.

>  // This endpoint is used by the client-http module to "connect" to an existing session and
  // update its last activity time. It performs authorization checks to make sure the caller
  // has access to the session, so even though it returns the same data, it behaves differently
  // from get("/:id").

patricker:
@jerryshao I'm using this outside of the API right now, specifically because it's listed as a work around to a bug.  Even if this the referenced PR is merged, it won't be available until v0.7; for those of us on the officially supported versions, such as v0.5, it would be nice to have documentation on how to resolve this issue.

jerryshao:
Then I would suggest to fix that bug first, rather than adding a document about non-public API.

mgaido91:
IMHO, what we would really need to document better are the configs available and their meaning. Currently, a user struggles to find the available configs and he/she has to guess their meaning/effect from the name (or by reading the code related to them).

I think a new section in the documentation explaining all the configs would be a great improvement.

jerryshao:
The bug is fixed in master branch, I'm going to close this PR.

PR #187, opened 2019-07-31
 LIVY-617: Livy session leak on Yarn when creating session duplicated names

shanyu:
When creating a session with duplicated name, instead of throw exception in SessionManager.register() method, we should stop the session. Otherwise the session driver process will keep running and end up creating a leaked Yarn application.

https://issues.apache.org/jira/browse/LIVY-617

## How was this patch tested?

This is just a simple fix and verified with manual end to end test.

### DIFF
--- a/server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
+++ b/server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
@@ -96,7 +96,10 @@ class SessionManager[S <: Session, R <: RecoveryMetadata : ClassTag](
     synchronized {
       session.name.foreach { sessionName =>
         if (sessionsByName.contains(sessionName)) {
-          throw new IllegalArgumentException(s"Duplicate session name: ${session.name}")
+          val errMsg = s"Duplicate session name: ${session.name} for session ${session.id}"
+          error(errMsg)
+          session.stop()
+          throw new IllegalArgumentException(errMsg)
         } else {
           sessionsByName.put(sessionName, session)
         }
@@ -104,6 +107,7 @@ class SessionManager[S <: Session, R <: RecoveryMetadata : ClassTag](
       sessions.put(session.id, session)
       session.start()
     }
+    info(s"Registered new session ${session.id}")
     session
   }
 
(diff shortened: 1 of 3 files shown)


jerryshao:
Reopen the PR to retest it.

jerryshao:
Thanks @shanyu for the fix. Would you please add an UT about your fix. Thanks!

shanyu:
@jerryshao UT added. Thanks for reviewing it.

server/src/test/scala/org/apache/livy/sessions/MockSession.scala
```diff
@@ -27,7 +27,8 @@ class MockSession(id: Int, owner: String, conf: LivyConf, name: Option[String] =
 
   override def start(): Unit = ()
 
-  override protected def stopSession(): Unit = ()
+  var stopped = false
+  override protected def stopSession(): Unit = {stopped = true}
```
jerryshao: nit: change the style to:

```
 = {
stopped = true
}
```

jerryshao:
CC @yiheng

yiheng:
LGTM

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
```diff
@@ -96,14 +96,18 @@ class SessionManager[S <: Session, R <: RecoveryMetadata : ClassTag](
     synchronized {
       session.name.foreach { sessionName =>
         if (sessionsByName.contains(sessionName)) {
-          throw new IllegalArgumentException(s"Duplicate session name: ${session.name}")
+          val errMsg = s"Duplicate session name: ${session.name} for session ${session.id}"
+          error(errMsg)
+          session.stop()
```
yiheng: It's a little weird that we call `session.stop()` before `session.start()`...

yiheng: From the code, when the duplicate name exception is thrown, the `session.start()` has not been called. Please let me know if I miss something.

jerryshao:
Reopen to test it.

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
shanyu: Yeah it a little odd.

createSession() calls InteractiveSession.create(), which calls LivyClientBuilder().build(), which creates a LivyClient, then create an InteractiveSession with this client. At this time the separate process is running which creates a Yarn application. This is before session.start() is called.

But to clean this Yarn application, we need to call session.stop().

jerryshao:
Reopen to test again.

PR #189, opened 2019-07-31
[LIVY-616] Livy Server discovery

o-shevchenko:
JIRA ticket https://issues.apache.org/jira/browse/LIVY-616.
Implement Livy Server discovery mechanism based on ZooKeeper. And provide convenient API for clients to get LivyServer URL without explicitly set it manually by "livy.server.host" property. Then we also can use it in LivyClientBuilder to set URL automatically. It will provide a more easy way for configuration. We don't need to set "livy.server.host" or "zeppelin.livy.url" manually in a case when we don't know where Livy Server will be started.

I have separated ZooKeeper logic from StateStore logic, as a result, we can use new ZooKeeperManager just to communicate with ZooKeeper and build new logic based on it (as I did for DiscoveryManager and ZooKeeperStateStore)

## How was this patch tested?
Added unit tests for new functionality.
Tested manually:
After starting LivyServer we can get Livy Server URI from ZooKeeper
image

image
Also verified on a secured cluster.

What the best way to add new API?
I see the couple of ways to do that:
1. Add it to LivyConf - in this case, we need to create DiscoveryManager in LivyConf which is not good.
2. Add it to LivyClient interface and set some variable during creating client via LivyClientBuilder.

Please, advise.


o-shevchenko:
What plugin do you use for import ordering?

o-shevchenko:
BinaryThriftServerSuite."fetch different data types" looks flaky

https://travis-ci.org/apache/incubator-livy/jobs/566398488
```
 BinaryThriftServerSuite:
 2328- Reuse existing session (30 seconds, 746 milliseconds)
 2329- fetch different data types *** FAILED *** (4 seconds, 270 milliseconds)
 2330 java.sql.SQLException: java.util.concurrent.ExecutionException: java.io.IOException: RSCClient instance stopped.
 2331 at org.apache.hive.jdbc.HiveStatement.waitForOperationToComplete(HiveStatement.java:401)
 2332 at org.apache.hive.jdbc.HiveStatement.execute(HiveStatement.java:266)
 2333 at org.apache.hive.jdbc.HiveStatement.executeQuery(HiveStatement.java:497)
 2334 at org.apache.livy.thriftserver.CommonThriftTests$class.dataTypesTest(ThriftServerSuites.scala:31)
 2335 at org.apache.livy.thriftserver.BinaryThriftServerSuite.dataTypesTest(ThriftServerSuites.scala:74)
 2336 at org.apache.livy.thriftserver.BinaryThriftServerSuite$$anonfun$2$$anonfun$apply$mcV$sp$2.apply(ThriftServerSuites.scala:98)
 2337 at org.apache.livy.thriftserver.BinaryThriftServerSuite$$anonfun$2$$anonfun$apply$mcV$sp$2.apply(ThriftServerSuites.scala:97)
 2338 at org.apache.livy.thriftserver.ThriftServerBaseTest$$anonfun$withJdbcStatement$1.apply(ThriftServerBaseTest.scala:102)
 2339 at org.apache.livy.thriftserver.ThriftServerBaseTest$$anonfun$withJdbcStatement$1.apply(ThriftServerBaseTest.scala:99)
 2340 at org.apache.livy.thriftserver.ThriftServerBaseTest.withJdbcConnection(ThriftServerBaseTest.scala:92)
```
Created separated ticket https://issues.apache.org/jira/browse/LIVY-618

o-shevchenko:
I have added API for getting Livy Server address to LivyClient, HttpClient, RSCClient and LivyScalaClient. I don't see a way to forward Livy address from DiscoveryManager to HttpClient or LivyScalaClient (we need to add a dependency on `livy-server` to `client-http` in this case, I'm not sure that this is acceptable).

So, from client code we can use:
```
val address = DiscoveryManager.getServerUri() - here we can use it in REST requests
val client = LivyClientBuilder.setURI(s"http://address").build - or create client 
val futureUri = client.getServerUri - and get address from client if needed

val scalaClient = new LivyScalaClient(client)
val futureUri = scalaClient.getServerUri
```
Added scaladocs, small refactoring.

o-shevchenko:
Refactored configuration to provide a more flexible mechanism for configuration:
```
# Livy Server discovery
# ZooKeeper quorum URLs, e.g. host1:port1,host2:port2
# livy.zookeeper.url =
# Name of base Livy zknode. Default livy
# livy.zookeeper.namespace = livy
# Name of Livy Server zknode. Uses livy.zookeeper.namespace as parent.
# By default will be /livy/server.uri
# livy.server.zookeeper.namespace = server.uri
# Number of trials to establish the connection to ZooKeeper quorum
# livy.server.zookeeper.connection.max.retries = 3
# Sleep time between connection retries to ZooKeeper quorum
# livy.server.zookeeper.connection.retry.interval.ms = 500
```

jerryshao:
Hi @o-shevchenko thanks a lot for your contribution. It would be better to have a design doc to describe what you're trying to resolve and what is your solution. It's really hard for the reviewer to understand your code without any documentation.

o-shevchenko:
Thanks a lot @jerryshao for your comment. I'll prepare documentation about the solution.
I'm just adding storing of Livy Server address/uri to ZooKeeper to be able to get it via some Livy API.

o-shevchenko:
@jerryshao you can find documentation in the Jira ticket: https://issues.apache.org/jira/browse/LIVY-616.
Please, let me know if something is unclear and I need to add more info.
Thanks!

o-shevchenko:
@jerryshao Any updates?
As far as I see, there are other PRs (#212 and #193) which adding similar functionality and trying to solve similar problems. We need to unify approaches and reuse the same properties and functionality.

server/src/main/scala/org/apache/livy/server/LivyServer.scala
```diff
@@ -385,6 +389,25 @@ class LivyServer extends Logging {
...
+      val host = resolvedSeverHost(livyConf)
+      val uri = new URI(s"http://$host:${livyConf.getInt(LivyConf.SERVER_PORT)}")
+      discoveryManager.setServerUri(uri)
+    }
+  }
+
+  private def resolvedSeverHost(livyConf: LivyConf) = {
+    val host = livyConf.get(LivyConf.SERVER_HOST)
+    if (host.equals(livyConf.get(LivyConf.SERVER_HOST.dflt.toString))) {
+      InetAddress.getLocalHost.getHostName
```
yiheng: I do not quite understand the if condition here. Could you please explain it a bit? Thanks.

And another question is why we return the hostname instead of IP here? IMHO, IP should be more convenient, as hostname user needs to update their hosts file or set up the name service.

o-shevchenko: You can find more detailed info in the specification which attached to the Jira ticket
```
To enable Livy Server discovery we need to set ​livy.zookeeper.url​. During start Livy Server we save URI to ZooKeeper. If we specify host by ​livy.server.host​ then we will use this value. if we use the default value “0.0.0.0” the local host name will be stored(InetAddress.getLocalHost.getHostName) to be able to communicate from other nodes. We will store address as java.net.URI (in format schema://host:port, e.g.http://livyhostname:8998​)
```
So, basically it's needed resolve Livy Server address by replacing `0.0.0.0` default value with the correct IP address if user didn't specify any Livy Server address in the conf file. In this case, we can use `InetAddress.getLocalHost.getHostName` to get the address from the node where LivyServer was started.


See your point, `InetAddress.getLocalHost.getHostAddress` might be the better choice here to store.

yiheng: From your description, should it be
```scala
if (host.equals(LivyConf.SERVER_HOST.dflt.toString)) {
...
```

o-shevchenko: oh, I see
it's my mistake.
good catch!
`livyConf.get(LivyConf.SERVER_HOST.dflt.toString)` not make a sense in this case, looks like it's typical copy problem

server/src/main/scala/org/apache/livy/server/discovery/LivyDiscoveryManager.scala
```diff
@@ -0,0 +1,65 @@
...
+  * Stores information about Livy Server location in ZooKeeper.
+  * The address will be stored in
+  * "/{@code LIVY_ZOOKEEPER_NAMESPACE}/{@code LIVY_SERVER_ZOOKEEPER_NAMESPACE}" znode
+  * By default, the full path to znode is /livy/server.uri.
+  * Need to set {@code livy.zookeeper.url} to be able to get information from ZooKeeper.
+  *
+  * @param livyConf - Livy configurations
+  * @param mockCuratorClient - used for testing
+  */
+class LivyDiscoveryManager(val livyConf: LivyConf,
```
yiheng: Should we consider providing a generic method to publishing configuration key/value to zookeeper?

In some other scenario, e.g. this PR https://github.com/apache/incubator-livy/pull/193 , a bunch of hive related configurations need to be published. 

If we have a generic method publish configuration key/value, it may be easier to let other components to leverage the code.

o-shevchenko: We have one in ZooKeeperManager
LivyDiscovery just uses it to simplify API

o-shevchenko: I just refactored StateStore to move functionality for communication with ZK from ZooKeeperStateStore.scala to separated trait ZooKeeperManager.scala. In ZooKeeperStateStore we will useZooKeeperManager to store data in ZK. ZooKeeperManager code was not changed except couple small improvements for configurations to make it more flexible.

We can use this trait everywhere when we need to store something to ZooKeeper since in generic, e.g:
```
def setData(key: String, value: Object): Unit = {
    val prefixedKey = prefixKey(key)
    val data = serializeToBytes(value)
    if (exist(prefixedKey)) {
      curatorClient.setData().forPath(prefixedKey, data)
    } else {
      curatorClient.create().creatingParentsIfNeeded().forPath(prefixedKey, data)
    }
  }

  def getData[T: ClassTag](key: String): Option[T] = {
    val prefixedKey = prefixKey(key)
    if (exist(prefixedKey)) {
      Option(deserialize[T](curatorClient.getData().forPath(prefixedKey)))
    } else {
      None
    }
  }
```

server/src/main/scala/org/apache/livy/server/LivyServer.scala
o-shevchenko: PR was updated

server/src/main/scala/org/apache/livy/server/discovery/LivyDiscoveryManager.scala
o-shevchenko: ZooKeeperManager should be used every time by design when we need to store something to ZK. I think we need to have one logic for ZK interaction for LivyServer discovery, ZK state store, Livy HA, Livy Thrift Server HA. That's why I moved this functionality to the separated trait, to make it reuseable for someone else.

o-shevchenko: Also, all configuration I have tried to make more generic and configurable on the top level to use it everywhere. As far as I see, for now, we have a couple of PRs with different ZK logic which can't be reused a lot and I see too specific confs like "livy.server.thrift.zookeeper.quorum". In this PR I want to not only add LivyServer discovery but also create init structure for subsequent work with ZooKeeper.

PR #193, opened 2019-08-02
[LIVY-621]add dynamic service discovery for thrift server

yantzu:

Add config and implementation to allow publish livy thrift server to zookeeper.  Configuration information are consistent with hiveserver2 for convenience.  Publish information is the same as hiveserver2 has, so beeline can discover thrift server.

https://issues.apache.org/jira/browse/LIVY-621

## How was this patch tested?

add unit test. And have tested manually.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -94,6 +94,7 @@
     <libthrift.version>0.9.3</libthrift.version>
     <kryo.version>2.22</kryo.version>
     <metrics.version>3.1.0</metrics.version>
+    <curator.version>2.12.0</curator.version>
     <mockito.version>1.9.5</mockito.version>
     <netty.spark-2.11.version>4.0.37.Final</netty.spark-2.11.version>
     <netty.version>${netty.spark-2.11.version}</netty.version>
@@ -500,6 +501,24 @@
         <groupId>org.json4s</groupId>
         <artifactId>json4s-jackson_${scala.binary.version}</artifactId>
         <version>${json4s.version}</version>
+      </dependency>
+
+      <dependency>
+        <groupId>org.apache.curator</groupId>
+        <artifactId>curator-framework</artifactId>
+        <version>${curator.version}</version>
+      </dependency>
+
+      <dependency>
+        <groupId>org.apache.curator</groupId>
+        <artifactId>curator-recipes</artifactId>
+        <version>${curator.version}</version>
+      </dependency>
+
+      <dependency>
+        <groupId>org.apache.curator</groupId>
+        <artifactId>curator-test</artifactId>
+        <version>${curator.version}</version>
       </dependency>
 
       <dependency>
--- a/server/pom.xml
+++ b/server/pom.xml
@@ -149,6 +149,16 @@
     </dependency>
 
     <dependency>
+      <groupId>org.apache.curator</groupId>
+      <artifactId>curator-framework</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.curator</groupId>
+      <artifactId>curator-recipes</artifactId>
+    </dependency>
+
+    <dependency>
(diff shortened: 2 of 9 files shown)


yantzu:
Hi @mgaido91 , could you please help to take a look at this PR.

jerryshao:
@yantzu what's the relation to this pr #189

mgaido91:
@jerryshao @vanzin I am pinging you because I am not sure about this PR. I mean, AFAIK Livy has no mechanism currently to create a "cluster" of Livy servers which interact among them, neither it has any rebalancing at all. I may be wrong, so that's why I am pinging you.

If I am not wrong on the above, I'd be against this PR which basically creates a cluster of thriftserver without having a cluster of Livy servers. I think we should first implement a feature like this for Livy server itself, and then eventually leverage it on the thriftserver part.

jerryshao:
Yes @mgaido91 , currently Livy doesn't have "cluster" itself, so the discovery mechanism seems more like a way to know LivyServer URL from ZK, #189 also has a similar proposal, and I left the similar comment in JIRA-616. Without Livy "cluster" support, a such discovery mechanism may not super useful/necessary.

yantzu:
> @yantzu what's the relation to this pr #189

Hi @jerryshao , #189 is for Rest API, and this one is for thrift API. This PR is to make livy thrift server compatibility with existing hiveserver2 client, jdbc or beeline, so user can move quickly from hiveserver to livy.

yantzu:
Hi @mgaido91 @jerryshao, thanks a lot for your comments!
If I understand correctly, "cluster" you mentioned should be something that communicate with each other and be able to transfer task from failed livy instance to active livy instance.  There are SessionStore code in livy, I am not sure if it is related to "cluster". However I think, no matter if "cluster" is supported, service discovery is necessary. 

Some considerations :
- This PR is to make livy thrift server compatibility  with hiveserver2, hiveserver2 has no "cluster" neither, but it works very well. 
- Hiveserver2 is quite stable, sometimes it may down, but we just rerun failed tasks in another hiveserver. And from architecture overview, livy should be more stable than hiveserver2. 
- Hiveserver2 jdbc/beeline has a RoundRobin based client rebalance mechanism
- Service discovery can hide backend server instances, and can enable HA.  

We have dozens of hiveserver2 in product deployment, it will help our users to move smoothly from hive to spark with service discovery, because it is almost impossible to let users to know all of these server instances.
Please feel free to advice.

mgaido91:
the point here is: from the project perspective we should first achieve the same level of HA and robustness for the Livy server part. Then, we can also have it for the thriftserver. It'd be very weird that HA is available only for the thrift part and not for the REST API for instance.

So I think that in order to have this, we should first wait for #189. Once that is over, we can port the same functionality to the thrift part, eventually leveraging the solution already present there.

yantzu:
Sure, that is a good point.

yantzu:
Livy user in my company is asking for Rest API Load Balance functionality these days. So I go throught #189 and #212 and then I realize service discovery for Rest is quite different with Thrift. The main difference is thrift is a long connection while rest is not. When thrift connect is break, the session is shutdown but rest session is not. And another difference is thrift's client are exsiting hive client while rest have many different clients.  
These differences make thrift service discovery much easier than rest, so maybe we should consider them seperately.

mgaido91:
I agree they are different and they may also lead to separate solutions, but I'd prefer to work on a full proposal which covers both cases. You may want to create a design doc and start a discussion maybe, proposing the solution you consider more appropriate. Once the design doc is approved I am happy to go ahead with this.

yantzu:
Hi @mgaido91 @jerryshao , I create a JIRA https://issues.apache.org/jira/browse/LIVY-698 to propose a cluster solution, could you please take a look. 
Thanks a lot!

PR #183, opened 2019-07-19
[LIVY-609] LDAP auth for Livy thriftserver

dockerzhang:
## What changes were proposed in this pull request?

To fix LIVY-609, add LDAP auth for Livy thriftserver, it's a extension for AuthenticationProvider.
https://issues.apache.org/jira/browse/LIVY-609

## How was this patch tested?
Tested manually

PR #195, opened 2019-08-09
[LIVY-634] To support reuse hive PasswdAuthenticationProvider implementation in ThriftServer

yantzu:

Currently Livy share same PasswdAuthenticationProvider interface with HiveServer2, but not able to reuse PasswdAuthenticationProvider implementation like LdapAuthenticationProviderImpl and PamAuthenticationProviderImpl. 

We can extend livy CustomAuthenticationProvider to support hive PasswdAuthenticationProvider  implementation. For example, to use LdapAuthenticationProviderImpl  we can add such configuration in livy.conf:
livy.server.thrift.authentication = CUSTOM
livy.server.thrift.custom.authentication.class = org.apache.hive.service.auth.LdapAuthenticationProviderImpl
livy.server.thrift.authentication.custom.hive.server2.authentication.ldap.url = xxxx
livy.server.thrift.authentication.custom.hive.server2.authentication.ldap.Domain = xxxx

https://issues.apache.org/jira/browse/LIVY-634

## How was this patch tested?

Tested manually.

### DIFF
--- a/thriftserver/server/pom.xml
+++ b/thriftserver/server/pom.xml
@@ -69,6 +69,10 @@
           <groupId>org.eclipse.jetty</groupId>
           <artifactId>*</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>org.apache.logging.log4j</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
       </exclusions>
     </dependency>
 
--- a/thriftserver/server/pom.xml
+++ b/thriftserver/server/pom.xml
@@ -56,6 +56,21 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hive</groupId>
+      <artifactId>hive-common</artifactId>
+      <version>${hive.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>org.apache.hive</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.eclipse.jetty</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
 
     <dependency>
       <groupId>org.apache.livy</groupId>
--- a/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/auth/AuthenticationProvider.scala
+++ b/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/auth/AuthenticationProvider.scala
@@ -20,6 +20,9 @@ package org.apache.livy.thriftserver.auth
 import java.lang.reflect.InvocationTargetException
 import javax.security.sasl.AuthenticationException
 
+import scala.collection.JavaConverters._
+
+import org.apache.hadoop.hive.conf.HiveConf
 import org.apache.hive.service.auth.PasswdAuthenticationProvider
 
 import org.apache.livy.LivyConf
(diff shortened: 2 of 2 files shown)


jerryshao:
Sorry there's some repo problems regarding to hive, will test again when the issue is fixed.

yiheng:
Looks like this patch has some overlap with this PR https://github.com/apache/incubator-livy/pull/183. We have a strong requirement of LDAP authentication in our environment. Would the community let us know which solution is preferred?

yiheng:
> Looks like this patch has some overlap with this PR #183. We have a strong requirement of LDAP authentication in our environment. Would the community let us know which solution is preferred?

@jerryshao @mgaido91 Can you give some suggestions here?

jerryshao:
I would suggest to reuse Hive's implementation instead of rewriting by ourselves, Hive's implementation is complete enough, we don't have to rewrite again. But here's the questions, 
 LivyServer also has a Filter based authentication mechanism, can they coexist, or can we unify them?

yantzu:
Hive's authentication interface is quite simple, it just accept a user & password pair, so there is no technical problem to use hive's implementation in livy-server. However, if we do it, it also means livy-server  module will have a dependency on hive-server.jar, not sure if it's acceptable.

thriftserver/server/pom.xml
```diff
@@ -56,6 +56,21 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hive</groupId>
+      <artifactId>hive-common</artifactId>
+      <version>${hive.version}</version>
+      <exclusions>
```
captainzmc: This will lead to the following jars in thriftserver, which will cause livy's logs to fail to display. We need to exclusion them.
log4j-1.2-api-2.10.0.jar
log4j-api-2.10.0.jar
log4j-core-2.10.0.jar
log4j-slf4j-impl-2.10.0.jar
log4j-web-2.10.0.jar

yantzu: thanks. issue fixed.

jerryshao:
Hi @mgaido91 , I think what I mean is whether we could unify two authentication both in REST and thrift, it doesn't mean that we have to use hive's authentication in REST, we can implement our own. My concern is that two authentications will introduce discrepancy. For example REST can enable Kerberos auth while thrift can enable others, is this acceptable, do we need a unified auth for both rest and thrift?

thriftserver/server/pom.xml
```diff
@@ -56,6 +56,25 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hive</groupId>
+      <artifactId>hive-common</artifactId>
```
mgaido91: we need to avoid all dependencies on hive classes...

mgaido91:
@jerryshao I see now what you mean.Yes, reusing the same auth would be ideal, but I am not sure whether it is possible...

thriftserver/server/pom.xml
captainzmc: I see  there is a separate ldap package under the auth in hive. If we support LDAP,  do we need to manually implement all the implementations under this package?

yantzu: Hi @captainzmc, could I know which package it is? I can not find such a package exist.

mgaido91: I'd say yes, but we can also come up with a different implementation which requires less classes

captainzmc: @yantzu I'm sorry if I didn't describe it clearly, it's not a separate package. Here it is： https://github.com/apache/hive/tree/master/service/src/java/org/apache/hive/service/auth/ldap

yantzu: I see, I misunderstood. Yes, Hive's LDAP implementation includes many classes, implement all of them is a huge work, that is why I reuse them. Another problem is LdapAuthenticationProviderImpl uses HiveConf class, which is in hive-common jar.  I cannot find a way to avoid those jar dependencies.

mgaido91: > implement all of them is a huge work
yes, I am aware. That's why in the original implementation this was left out for future work.
The approach in #183 is the right one IMO. We might also accept not to have all the functionalities Hive has for starting, but we need to clearly clarify the differences and to have an implementation which can be extended later to have the same functionalities.

yantzu: Yes. I will close this PR sometime later, may be when #183 finished.

mgaido91: if there is no answer from the author in #183, I think you might take it over too.

yantzu: OK, I may take a look later.

yiheng: I suggest that we split the [LIVY-609](https://issues.apache.org/jira/browse/LIVY-609)  into sub-tasks. In many scenarios, the basic LDAP function can meet the requirement.

yantzu:
close this PR as there is better solution.

PR #196, opened 2019-08-13
[LIVY-606][LIVY-152] fix the invalid conf of pyspark.archives and sparkr.package

runzhiwang:
1. Remove livy.pyspark.archives and livy.sparkr.package in livy.conf, which can be set in livy-client.conf.  

2. Replace "livyConf.get(RSCConf.Entry.PYSPARK_ARCHIVES)" with "builderProperties.get(RSCConf.Entry.PYSPARK_ARCHIVES.key())", because livyConf store the config of livy.conf which doesn't have  PYSPARK_ARCHIVES, and builderProperties has PYSPARK_ARCHIVES of livy-client.conf. The sames to SPARKR_PACKAGE.

## How was this patch tested?

Config the livy.rsc.sparkr.package by hdfs path, then livy will not upload zip from local to hdfs every time a session is started.

### DIFF
--- a/server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
+++ b/server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
@@ -183,7 +183,7 @@ object InteractiveSession extends Logging {
     }
 
     def findSparkRArchive(): Option[String] = {
-      Option(livyConf.get(RSCConf.Entry.SPARKR_PACKAGE.key())).orElse {
+      builderProperties.get(RSCConf.Entry.SPARKR_PACKAGE.key()).orElse {
         sys.env.get("SPARK_HOME").flatMap { case sparkHome =>
           val path = Seq(sparkHome, "R", "lib", "sparkr.zip").mkString(File.separator)
           val rArchivesFile = new File(path)
@@ -247,7 +247,7 @@ object InteractiveSession extends Logging {
     }
 
     def findPySparkArchives(): Seq[String] = {
-      Option(livyConf.get(RSCConf.Entry.PYSPARK_ARCHIVES))
+      builderProperties.get(RSCConf.Entry.PYSPARK_ARCHIVES.key())
         .map(_.split(",").toSeq)
         .getOrElse {
           sys.env.get("SPARK_HOME") .map { case sparkHome =>
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -79,14 +79,6 @@
 # during session creation.
 # livy.repl.jars =
 
-# Location of PySpark archives. By default Livy will upload the file from SPARK_HOME, but
-# by caching the file in HDFS, startup time of PySpark sessions on YARN can be reduced.
-# livy.pyspark.archives =
-
-# Location of the SparkR package. By default Livy will upload the file from SPARK_HOME, but
-# by caching the file in HDFS, startup time of R sessions on YARN can be reduced.
-# livy.sparkr.package =
-
 # List of local directories from where files are allowed to be added to user sessions. By
 # default it's empty, meaning users can only reference remote URIs when starting their
 # sessions.


conf/livy-client.conf.template
```diff
@@ -67,8 +67,6 @@
 # livy.rsc.jars =
 # Location of the SparkR package for running sparkr
 # livy.rsc.sparkr.package =
-# Location of the PySpark package for running pyspark
-# livy.rsc.pyspark.archives =
```
jerryshao: @runzhiwang would you please explain why should we remove this configuration?

jerryshao:
@yiheng would you please also review this patch, thanks!

conf/livy-client.conf.template
runzhiwang: @jerryshao Because spark will upload the pyspark.zip from local to hdfs, so it is useless to config it in livy-client.conf to avoid uploading . It is also reported in LIVY-152: Livy should use PYSPARK_ARCHIVES_PATH instead of having its own setting.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
```diff
@@ -247,30 +247,26 @@ object InteractiveSession extends Logging {
     }
 
     def findPySparkArchives(): Seq[String] = {
-      Option(livyConf.get(RSCConf.Entry.PYSPARK_ARCHIVES))
```
yiheng: The `RSCConf.Entry.PYSPARK_ARCHIVES` is also used [here](https://github.com/apache/incubator-livy/blob/master/rsc/src/main/java/org/apache/livy/rsc/ContextLauncher.java#L197), should we also remove it? Should we remove the `RSCConf.Entry.PYSPARK_ARCHIVES`?

conf/livy-client.conf.template
runzhiwang: @jerryshao Sorry, it's wrong to remove this configuration in livy-clent.conf. I have updated the pr.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
runzhiwang: @yiheng I have updated the pr. Please review it again. Thanks.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
```diff
@@ -183,7 +183,7 @@ object InteractiveSession extends Logging {
     }
 
     def findSparkRArchive(): Option[String] = {
-      Option(livyConf.get(RSCConf.Entry.SPARKR_PACKAGE.key())).orElse {
```
yiheng: Should we remove `RSCConf.Entry.SPARKR_PACKAGE` ?

runzhiwang: @yiheng I read the SPARKR_PACKAGE from livy-client.conf instead of livy.conf. The builderProperties has the config of  livy-client.conf.

yiheng: ok.

RogPodge:
+1 on getting this issue PR into Livy

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
RogPodge: Has this been addressed and how?

RogPodge:
Actually I can't seem to replicate the stated behavior on my machine after trying to incorporate this PR. Is there a reason why we want this config to be in livy-client.conf rather than livy.conf?

runzhiwang:
@RogPodge 1. This config should be defined by user, so in livy-client.conf.  2. Actually, this config in livy.conf has not been read in livy, so remove or not does not matter.  3. Sorry, what do you mean "I can't seem to replicate the stated behavior on my machine after trying to incorporate this PR"?

RogPodge:
> 
> 
>  Sorry, what do you mean "I can't seem to replicate the stated behavior on my machine after trying to incorporate this PR"?

After porting your PR to my local Livy instance, I still see the following in the Livy logs when I try to submit a pyspark application. 

INFO yarn.Client: Uploading resource file:/opt/spark/python/lib/pyspark.zip -> hdfs://nmnode-0-0.nmnode-0-svc:9000/user/root/.sparkStaging/application_1568331306042_0026/pyspark.zip

I noticed that the PR only makes changes to the InteractiveSession.scala, so how do we ensure that the pyspark.zip file is cached for Batch Sessions?

RogPodge:
Update: I'm submitting a batch job with a yarn master. When yarn is the master the pyspark files always seem to get copied

runzhiwang:
@RogPodge pyspark.zip is uploaded by spark, not livy.

shanyu:
@runzhiwang what do you mean by "pyspark.zip is uploaded by spark, not livy". I thought this PR is to let customer set livy.pyspark.archives to hdfs:///pyspark.zip, then spark submit won't upload the zip file for every session and every batch.

runzhiwang:
@shanyu I think we cannot achieve "spark submit won't upload the zip file for every session and every batch", because spark upload pyspark.zip in the following code and we cannot stop it in livy: https://github.com/apache/spark/blob/master/resource-managers/yarn/src/main/scala/org/apache/spark/deploy/yarn/Client.scala#L862.

yantzu:
This PR LGTM. Looking forward to have it mergered to master, is there other concerns about this PR?

jerryshao:
No any special concern, just need another careful revisit to avoid regression.

yantzu:
Hi @runzhiwang , when you test set livy.rsc.sparkr.package, do you set in livy-client.conf or in session request params. I tested set livy-client.conf, it saw sparkr.zip is always uploaded.

runzhiwang:
@yantzu Hi, I have test it again, and sparkr will not upload when set  livy.rsc.sparkr.package in livy-client.conf. The first image shows sparkr was uploaded when not set livy.rsc.sparkr.package, and the second image shows sparkr was not uploaded when set it. And the attach file is the source code, you can change  livy.rsc.sparkr.package to your hdfs path, and test it again.
image
image
[LivySparkR.zip](https://github.com/apache/incubator-livy/files/3937502/LivySparkR.zip)

yantzu:
@runzhiwang Did you test in dev mode? Could you please try build whole livy package and deploy to test. I think the difference may be because below code in InteractiveSession.scala
val sparkRArchive = if (!LivyConf.TEST_MODE) findSparkRArchive() else None

runzhiwang:
@yantzu I have test both TEST_MODE=true and TEST_MODE=false, they both will not upload sparkr. I think you can debug the code by `print(builderProperties.get(RSCConf.Entry.SPARKR_PACKAGE.key()))` before https://github.com/apache/incubator-livy/blob/master/server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala#L317, and see whether it print the path which config in livy-client.conf.

yantzu:
It is strange I got different result with you. I config livy-client.conf  as:
image
And add a log in InteractiveSession.scala as:
image

After session start, I see log:
image
image

And in spark UI I can see:
image

I looked in to code, the reason is findSparkRArchive will always find sparkr.zip in SPARK_HOME, set it to livy.rsc.sparkr.package config key. Then, when merge config with livy-client.conf, the config in livy-client.conf has lowest priority. So sparkr in SPARK_HOME is always used.

runzhiwang:
@yantzu Hi, could you ` error(builderProperties.get(RSCConf.Entry.SPARKR_PACKAGE.key()))` before `val sparkRArchive = if (!LivyConf.TEST_MODE) findSparkRArchive() else None` ? Then let's see what's the output.

yantzu:
@runzhiwang , I add 3 logs
image

And get log:
image

runzhiwang:
@yantzu Hi, The first log "ERROR InteractiveSession$: None" shows the config in your livy-client.conf was not passed to livy at all. Where your livy-client.conf placed ?

yantzu:
@runzhiwang in the conf path of livy install dir
image

yantzu:
@runzhiwang in the conf dir of livy install dir
image

runzhiwang:
@yantzu I think livy-client.conf should place in the src/main/resources of your client rather than in the conf of livy.  You can see the example in LivySparkR.zip I uploaded before.

runzhiwang:
@yantzu If you want to config it in livy rather than client, you can put livy.rsc.sparkr.packge=xx in livy.conf.  And livy will not read livy-client.conf at all, so your config in livy-client.conf cannot valid.

yantzu:
Oh, I see, we are talking about totally different issues. your PR is about set livy.rsc.sparkr.package in request params does not work, but my concern is set  livy.rsc.sparkr.package in sever side as default value doesn't work.

runzhiwang:
@yantzu The reason of: set livy.rsc.sparkr.package in sever side as default value doesn't work, is livy will not read livy-client.conf. So you can config it livy.rsc.sparkr.packge=xx in livy.conf

yantzu:
@runzhiwang  livy server will read livy-client.conf  in LivyClientBuilder, it should be used both in HttpClient and RSCClient. And livy.rsc.sparkr.packge set in livy.conf doesn't work neither.

yantzu:
@runzhiwang rollback your changes and set livy.rsc.sparkr.packge set in livy.conf, it works! Thanks. 
And this PR has side effect we need to be careful.

runzhiwang:
@yantzu Hi, the target of this PR is remove the  livy.rsc.sparkr.packge in livy.conf. Because I think it's a conf a client rather than server.

yantzu:
configuration relevant code in livy is a mess

yiheng:
> configuration relevant code in livy is a mess

+1

PR #197, opened 2019-08-13
[LIVY-574][TESTS][THRIFT] Add tests for metadata operations

yiheng:
Add unit test for existing meta operation: GetCatalogsOperation/GetTableTypesOperation/GetTypeInfoOperation.

We also fix issues we met.

## How was this patch tested?
Add new unit tests and existing test.

We also use SquirrelSQL test relate operations.
image

image

### DIFF
--- a/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/cli/ThriftCLIService.scala
+++ b/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/cli/ThriftCLIService.scala
@@ -397,7 +397,7 @@ abstract class ThriftCLIService(val cliService: LivyCLIService, val serviceName:
   override def GetTypeInfo(req: TGetTypeInfoReq): TGetTypeInfoResp = {
     val resp = new TGetTypeInfoResp
     try {
-      val operationHandle = cliService.getTypeInfo(new SessionHandle(req.getSessionHandle))
+      val operationHandle = cliService.getTypeInfo(createSessionHandle(req.getSessionHandle))
       resp.setOperationHandle(operationHandle.toTOperationHandle)
       resp.setStatus(ThriftCLIService.OK_STATUS)
     } catch {
@@ -412,7 +412,7 @@ abstract class ThriftCLIService(val cliService: LivyCLIService, val serviceName:
   override def GetCatalogs(req: TGetCatalogsReq): TGetCatalogsResp = {
     val resp = new TGetCatalogsResp
     try {
-      val opHandle = cliService.getCatalogs(new SessionHandle(req.getSessionHandle))
+      val opHandle = cliService.getCatalogs(createSessionHandle(req.getSessionHandle))
       resp.setOperationHandle(opHandle.toTOperationHandle)
       resp.setStatus(ThriftCLIService.OK_STATUS)
     } catch {
@@ -463,7 +463,7 @@ abstract class ThriftCLIService(val cliService: LivyCLIService, val serviceName:
   override def GetTableTypes(req: TGetTableTypesReq): TGetTableTypesResp = {
     val resp = new TGetTableTypesResp
     try {
-      val opHandle = cliService.getTableTypes(new SessionHandle(req.getSessionHandle))
+      val opHandle = cliService.getTableTypes(createSessionHandle(req.getSessionHandle))
       resp.setOperationHandle(opHandle.toTOperationHandle)
       resp.setStatus(ThriftCLIService.OK_STATUS)
     } catch {
--- a/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/GetTypeInfoOperation.scala
+++ b/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/GetTypeInfoOperation.scala
@@ -44,7 +44,7 @@ class GetTypeInfoOperation(sessionHandle: SessionHandle)
         val data = Array[Any](
           t.name,
           t.sqlType,
-          t.precision.getOrElse(null),
+          t.precision.orNull,
           null, // LITERAL_PREFIX
           null, // LITERAL_SUFFIX
           null, // CREATE_PARAMS
@@ -59,7 +59,7 @@ class GetTypeInfoOperation(sessionHandle: SessionHandle)
           0.toShort, // MAXIMUM_SCALE
           null, // SQL_DATA_TYPE
           null, // SQL_DATETIME_SUB
-          t.numPrecRadix.getOrElse(null))
+          t.numPrecRadix.orNull)
         rowSet.addRow(data)
       }
       setState(OperationState.FINISHED)
--- a/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/GetTypeInfoOperation.scala
+++ b/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/GetTypeInfoOperation.scala
@@ -44,21 +44,22 @@ class GetTypeInfoOperation(sessionHandle: SessionHandle)
         val data = Array[Any](
           t.name,
           t.sqlType,
-          t.precision,
+          t.precision.getOrElse(null),
           null, // LITERAL_PREFIX
           null, // LITERAL_SUFFIX
           null, // CREATE_PARAMS
-          DatabaseMetaData.typeNullable, // All types are nullable
+          DatabaseMetaData.typeNullable.toShort, // All types are nullable
           t.caseSensitive,
+          t.searchable,
           t.unsignedAttribute,
           false, // FIXED_PREC_SCALE
           false, // AUTO_INCREMENT
           null, // LOCAL_TYPE_NAME
-          0, // MINIMUM_SCALE
-          0, // MAXIMUM_SCALE
+          0.toShort, // MINIMUM_SCALE
+          0.toShort, // MAXIMUM_SCALE
           null, // SQL_DATA_TYPE
           null, // SQL_DATETIME_SUB
-          t.numPrecRadix)
+          t.numPrecRadix.getOrElse(null))
         rowSet.addRow(data)
       }
       setState(OperationState.FINISHED)
--- a/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/MetadataOperation.scala
+++ b/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/MetadataOperation.scala
@@ -33,6 +33,10 @@ abstract class MetadataOperation(sessionHandle: SessionHandle, opType: Operation
 
   protected def rowSet: ThriftResultSet
 
+  private var offset = 0
+
+  private lazy val EMPTY_ROW_SET = ThriftResultSet(getResultSetSchema, protocolVersion)
+
   @throws[HiveSQLException]
   override def close(): Unit = {
     setState(OperationState.CLOSED)
@@ -48,8 +52,14 @@ abstract class MetadataOperation(sessionHandle: SessionHandle, opType: Operation
     assertState(Seq(OperationState.FINISHED))
     validateFetchOrientation(orientation)
     if (orientation.equals(FetchOrientation.FETCH_FIRST)) {
-      rowSet.setRowOffset(0)
+      offset = 0
     }
-    rowSet
+
+    if (offset >= rowSet.numRows) {
+      EMPTY_ROW_SET
+    } else {
+      offset += rowSet.numRows
+      rowSet
+    }
   }
 }
--- a/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/MetadataOperation.scala
+++ b/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/MetadataOperation.scala
@@ -33,10 +33,6 @@ abstract class MetadataOperation(sessionHandle: SessionHandle, opType: Operation
 
   protected def rowSet: ThriftResultSet
 
-  private var offset = 0
-
-  private lazy val EMPTY_ROW_SET = ThriftResultSet(getResultSetSchema, protocolVersion)
-
   @throws[HiveSQLException]
   override def close(): Unit = {
     setState(OperationState.CLOSED)
@@ -52,14 +48,8 @@ abstract class MetadataOperation(sessionHandle: SessionHandle, opType: Operation
     assertState(Seq(OperationState.FINISHED))
     validateFetchOrientation(orientation)
     if (orientation.equals(FetchOrientation.FETCH_FIRST)) {
-      offset = 0
+      rowSet.setRowOffset(0)
     }
-
-    if (offset >= rowSet.numRows) {
-      EMPTY_ROW_SET
-    } else {
-      offset += rowSet.numRows
-      rowSet
-    }
+    rowSet.extractSubset(maxRows)
   }
 }
--- a/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/MetadataOperation.scala
+++ b/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/MetadataOperation.scala
@@ -50,6 +50,6 @@ abstract class MetadataOperation(sessionHandle: SessionHandle, opType: Operation
     if (orientation.equals(FetchOrientation.FETCH_FIRST)) {
       rowSet.setRowOffset(0)
     }
-    rowSet.extractSubset(maxRows)
+    rowSet.extractSubset(maxRows.toInt)
   }
 }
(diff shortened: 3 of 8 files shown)


thriftserver/server/src/main/scala/org/apache/livy/thriftserver/cli/ThriftCLIService.scala
```diff
@@ -397,7 +397,7 @@ abstract class ThriftCLIService(val cliService: LivyCLIService, val serviceName:
   override def GetTypeInfo(req: TGetTypeInfoReq): TGetTypeInfoResp = {
     val resp = new TGetTypeInfoResp
     try {
-      val operationHandle = cliService.getTypeInfo(new SessionHandle(req.getSessionHandle))
+      val operationHandle = cliService.getTypeInfo(createSessionHandle(req.getSessionHandle))
```
yiheng: Create a session handle with the real protocol version of this session. The original is using version_v1 as default, which will not pass the requirement when generating the thrift result set. see here

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/MetadataOperation.scala
```diff
@@ -42,8 +46,14 @@ abstract class MetadataOperation(sessionHandle: SessionHandle, opType: Operation
     assertState(Seq(OperationState.FINISHED))
     validateFetchOrientation(orientation)
     if (orientation.equals(FetchOrientation.FETCH_FIRST)) {
-      rowSet.setRowOffset(0)
+      offset = 0
+    }
+
+    if (offset >= rowSet.numRows) {
```
yiheng: Return am empty row set when no row should be returned. The original implementation will cause an infinite RowSet on the client side.

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -69,6 +69,51 @@ trait CommonThriftTests {
     }
     assert(!resultSetComplex.next())
   }
+
+  def getCatalogTest(connection: Connection): Unit = {
+    val metadata = connection.getMetaData
+    val catalogResultSet = metadata.getCatalogs()
```
mgaido91: it'd be great to have also a test with some data returned

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -69,6 +69,51 @@ trait CommonThriftTests {
...
+    assert(typeInfoResultSet.getBoolean(12) == false)
+    assert(typeInfoResultSet.getString(13) == null)
+    assert(typeInfoResultSet.getShort(14) == 0)
+    assert(typeInfoResultSet.getShort(15) == 0)
+    assert(typeInfoResultSet.getInt(16) == 0)
+    assert(typeInfoResultSet.getInt(17) == 0)
+    assert(typeInfoResultSet.getInt(18) == 0)
+    val LEFT_TYPE_NUM = 16
+    for (i <- 1 to LEFT_TYPE_NUM) {
+      assert(typeInfoResultSet.next())
```
mgaido91: it'd be great to test them all..

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/MetadataOperation.scala
mgaido91: I am not sure about this change honestly, may you please share in more details which was the problem? This way I hope we can find a better solution than this one.

yiheng: The problem is the last assert won't pass in the test.
```Scala
val metadata = connection.getMetaData
val tableTypesResultSet = metadata.getTableTypes()
tableTypesResultSet.next()
assert(tableTypesResultSet.getString(1) == "TABLE")
tableTypesResultSet.next()
assert(tableTypesResultSet.getString(1) == "VIEW")
assert(!tableTypesResultSet.next())
```
The `next()` of ResultSet will always be `true`. So when fetching table types or data types, the JDBC client will keep fetching forever. 

The root cause is in the [HiveColumnBasedSet](https://github.com/apache/hive/blob/master/jdbc/src/java/org/apache/hive/jdbc/HiveQueryResultSet.java#L352), the `hasNext()` is always be true.
```Java
if (!fetchedRowsItr.hasNext()) {
    return false;
}
```
The [Iterator.hasNext()](https://github.com/apache/hive/blob/master/service/src/java/org/apache/hive/service/cli/ColumnBasedSet.java#L179) implementation is this
```Java
public boolean hasNext() {
  return index < numRows();
}
```
And the [numRows](https://github.com/apache/hive/blob/master/service/src/java/org/apache/hive/service/cli/ColumnBasedSet.java#L130) is only related with column size
```Java
public int numRows() {
  return columns.isEmpty() ? 0 : columns.get(0).size();
}
```
As we always return a not empty colum result set, the `ResultSet.next()` is always true.

yiheng: Here are the two solutions I have thought:
1. Return an empty ResultSet when needed
2. Implement an `extractSubset()` method for ThriftResultSet, which is similar to what Spark thrift server do. In another patch, you have a concern of it's not efficient.

Please let me know your thought. Thanks.

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
yiheng: The catalog is not supported in Spark. It' a no-op. Please see [here](catalogs are actually not supported in spark, so this is a no-op)

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
yiheng: fixed

jerryshao:
Please solve the conflict, @yiheng .

yiheng:
@jerryshao resolved

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/MetadataOperation.scala
yiheng: ping @mgaido91 

Hi Marco, do you have any suggestion here?

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/GetTypeInfoOperation.scala
```diff
@@ -44,21 +44,22 @@ class GetTypeInfoOperation(sessionHandle: SessionHandle)
         val data = Array[Any](
           t.name,
           t.sqlType,
-          t.precision,
+          t.precision.getOrElse(null),
```
jerryshao: `orNull`.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/GetTypeInfoOperation.scala
```diff
@@ -44,21 +44,22 @@ class GetTypeInfoOperation(sessionHandle: SessionHandle)
...
           false, // AUTO_INCREMENT
           null, // LOCAL_TYPE_NAME
-          0, // MINIMUM_SCALE
-          0, // MAXIMUM_SCALE
+          0.toShort, // MINIMUM_SCALE
+          0.toShort, // MAXIMUM_SCALE
           null, // SQL_DATA_TYPE
           null, // SQL_DATETIME_SUB
-          t.numPrecRadix)
+          t.numPrecRadix.getOrElse(null))
```
jerryshao: ditto.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/MetadataOperation.scala
jerryshao: What's the pros and cons of these two methods?

yiheng: @jerryshao 

1. Return an empty ResultSet when needed
**pros**: No data copy. It's efficient
**cons**: Not straight forward. And we need to maintain the `row offset` in the `MetadataOperation`.

2. extractSubset()
**pros**: Straight forward. Maintain the `row offset` in the `ResultSet` itself. It's what spark thriftserver do.
**cons**: Data copy every time. Honestly, I think it's not a big issue as the metadata is small.  However, `ResultSet` is also used in other places, e.g. fetch table data. Not sure if it will be misused in the future.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/GetTypeInfoOperation.scala
yiheng: fixed

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/GetTypeInfoOperation.scala
yiheng: fixed

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/MetadataOperation.scala
jerryshao: I see, thanks! IMO I think current solution is OK. Maybe we need to add some comments on the code.

mgaido91: sorry for the late answer, I was on vacation last week. I just came back and I am processing the backlog :)

Since this is a quite delicate thing, I'd like to think carefully about it. Please give me a day or so to get back to this. Thanks.

mgaido91: Anyway, apart from this, looks fine to me. I'd still like to think a bit more on this. I really don't like this solution and I'd prefer to find another one. Sorry for the delay again.

mgaido91:
title:

`[LIVY-574][TESTS][THRIFT] Add tests for metadata operations`

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/MetadataOperation.scala
mgaido91: I see now the reason of the `extractSubset` you introduced in the other PR. While in the other case it is not needed, IMHO, here it is the right solution. May you please take that and use it here? Thanks!

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -177,6 +181,70 @@ trait CommonThriftTests {
     assert(columnsResultSet.getString(23) == "NO")
     assert(!columnsResultSet.next())
   }
+
+  def getCatalogTest(connection: Connection): Unit = {
+    val metadata = connection.getMetaData
+    val catalogResultSet = metadata.getCatalogs()
+    // Spark doesn't support getCatalog. In current implementation, it's a no-op and not return
```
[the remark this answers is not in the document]
yiheng: fixed

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -17,7 +17,11 @@
 
 package org.apache.livy.thriftserver
 
-import java.sql.{Connection, Date, SQLException, Statement}
+import java.sql.Connection
```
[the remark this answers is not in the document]
yiheng: fixed

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/operation/MetadataOperation.scala
yiheng: done

thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ColumnBuffer.java
```diff
@@ -204,6 +204,12 @@ public BitSet getNulls() {
     return nulls != null ? BitSet.valueOf(nulls) : new BitSet();
   }
 
+  /**
+   * Extract subset data to a new column buffer.
+   * @param start start row number
+   * @param end end row number, which is not included in the subset data
+   * @return
+   */
```
yiheng: I forgot to delete this in another patch... Please review this method...

thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ColumnBuffer.java
```diff
@@ -204,6 +204,12 @@ public BitSet getNulls() {
...
   public ColumnBuffer extractSubset(int start, int end) {
     ColumnBuffer subset = new ColumnBuffer(type);
     subset.currentSize = end - start;
```
mgaido91: I am not sure about this method, expecially regarding the handling of nulls. Nulls are not inserted in the vaues but the nulls bitset account for them. I am very surprised that the tests pass with this. I expect worng results with nulls. May you please check?

yiheng: Yes, you're right. I didn't handle the nulls when extractSubset.

I have fixed this and handle some illegal start/end input. I have also added some [test cases](https://github.com/apache/incubator-livy/pull/197/files#diff-275dca4a619dd9b0b7f3abbfdfeb00f3R233) to cover this. Can you please review? Thanks.

mgaido91: I am afraid that the `arraycopy` can go out of the boundaries (eg. a column with 110 nulls). Can you please check what was done in https://github.com/apache/hive/blob/master/serde/src/java/org/apache/hadoop/hive/serde2/thrift/ColumnBuffer.java and do something similar here?

Moreover, may you please add tests for the case mentioned above? Thanks.

yiheng: Have updated code based on the hive ColumnBuffer.

Please notice that the [addValue](https://github.com/apache/hive/blob/master/serde/src/java/org/apache/hadoop/hive/serde2/thrift/ColumnBuffer.java#L360) in hive ColumnBuffer is different from ours. We still need to handle when `values array length` is smaller than `currentSize` case.

Add two extra tests cases to cover those cases. See [here](https://github.com/apache/incubator-livy/pull/197/files#diff-275dca4a619dd9b0b7f3abbfdfeb00f3R233)

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/serde/ThriftResultSet.scala
```diff
@@ -121,4 +122,13 @@ class ColumnOrientedResultSet(
   override def setRowOffset(rowOffset: Long): Unit = this.rowOffset = rowOffset
 
   override def numRows: Int = columns.headOption.map(_.size).getOrElse(0)
+
+  override def extractSubset(maxRows: Int): ThriftResultSet = {
+    val nRows = Math.min(numRows, maxRows)
+    val result = new ColumnOrientedResultSet(
+      columns.map(_.extractSubset(nRows))
+    )
```
mgaido91: nit: on the line above

thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ColumnBuffer.java
```diff
@@ -204,39 +240,71 @@ public BitSet getNulls() {
     return nulls != null ? BitSet.valueOf(nulls) : new BitSet();
   }
 
-  public ColumnBuffer extractSubset(int start, int end) {
-    ColumnBuffer subset = new ColumnBuffer(type);
-    subset.currentSize = end - start;
-    subset.ensureCapacity();
+  /**
+   * Extract subset data to a new column buffer.
```
mgaido91: nit: since this also "removes" the extracted data from this column set, it would be great to add a note here about this.

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -190,6 +190,70 @@ trait CommonThriftTests {
...
+    val metadata = connection.getMetaData
+    val tableTypesResultSet = metadata.getTableTypes()
+    tableTypesResultSet.next()
+    assert(tableTypesResultSet.getString(1) == "TABLE")
+    tableTypesResultSet.next()
+    assert(tableTypesResultSet.getString(1) == "VIEW")
+    assert(!tableTypesResultSet.next())
+  }
+
+  def getTypeInforTest(connection: Connection): Unit = {
```
mgaido91: nit:
```suggestion
  def getTypeInfoTest(connection: Connection): Unit = {
```

mgaido91:
Thanks, merging to master.

PR #198, opened 2019-08-14
[LIVY-635][BUILD] Fix travis fail to build by excluding the dependency of hbase

runzhiwang:

Fix travis fail to build by excluding the dependency of hbase.

## How was this patch tested?

existed UT and IT.

### DIFF
--- a/thriftserver/server/pom.xml
+++ b/thriftserver/server/pom.xml
@@ -39,6 +39,10 @@
           <groupId>tomcat</groupId>
           <artifactId>*</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>org.apache.hbase</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
       </exclusions>
     </dependency>
     <dependency>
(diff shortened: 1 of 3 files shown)


thriftserver/server/pom.xml
```diff
@@ -39,6 +39,10 @@
           <groupId>tomcat</groupId>
           <artifactId>*</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>org.apache.hbase</groupId>
```
yiheng: This dependency scope is not test. Not sure if this exclusion is safe

jerryshao:
CC @mgaido91 , would you please help to review. Thanks!

jerryshao:
@runzhiwang can you please also verify if thriftserver can be worked  without issue?

runzhiwang:
@jerryshao It's ok to work.

jerryshao:
I see, thank!

mgaido91:
LGTM too, this should be safe as AFAIK, dependencies on HBase are needed only for using HBase as a metastore, but we do not use any metastore at all. Thanks.

PR #200, opened 2019-08-16
[LIVY-637][ThriftServer]Fix NullPointerException when create database using thriftserver

captainzmc:

Spark beeline use old hive-jdbc-client doesn’t do null point ref check. So  when new TRowSet, setColumes make sure column set not null.

## How was this patch tested?

Connect to livy's thriftserver using spark beeline. And create/use/drop database will no longer have nullpointerexceptions after execution.

image

### DIFF
--- a/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/serde/ThriftResultSet.scala
+++ b/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/serde/ThriftResultSet.scala
@@ -111,7 +111,10 @@ class ColumnOrientedResultSet(
   }
 
   override def toTRowSet: TRowSet = {
+    // Spark beeline use old hive-jdbc-client doesn’t do null point ref check. When we new TRowSet,
+    // setColumes make sure column set not null.
     val tRowSet = new TRowSet(rowOffset, new util.ArrayList[TRow])
+    tRowSet.setColumns(new util.ArrayList[TColumn]())
     columns.foreach { c =>
       tRowSet.addToColumns(ThriftResultSet.toTColumn(c))
     }


thriftserver/server/src/main/scala/org/apache/livy/thriftserver/serde/ThriftResultSet.scala
```diff
@@ -112,6 +112,7 @@ class ColumnOrientedResultSet(
 
   override def toTRowSet: TRowSet = {
     val tRowSet = new TRowSet(rowOffset, new util.ArrayList[TRow])
+    tRowSet.setColumns(new util.ArrayList[TColumn]())
```
yiheng: should we add a comment here to explain why we have this?

captainzmc: OK, now I had added a comment here to explain this.

TisonKun:
Hi @jerryshao shall we make progress on this thread?

@captainzmc may I ask why you intermittently close and reopen the issue? Do you want to retest the pull request? If so, maybe there is or we can implement a retest approach instead of close and reopen.

captainzmc:
Hi @TisonKun 
Now livy's Travis is unstable and needs several retries to succeed. So I turn this PR on and off each time to re-execute Travis. The problem of unstable Travis needs to be solved later.

TisonKun:
I think another solution should be use an auto-retest trigger instead of close-reopen which would pollute notification quite a lot...

captainzmc:
That sounds great，Thanks tison. This is the first time for me to submit PR, and I will adopt this method later.

jerryshao:
Merging to master branch, thanks for the fix.

mgaido91:
Sorry for being late, just noticed this one. I am not sure this is a valid use case/usage. Spark's beeline is based on Hive 1.2/2.3, as well as Spark thriftserver. While Livy uses Hive 3.0 as a base. Despite old client may be able to communicate to Livy, I don't think this should be a valid/supported use case. I think that the client should be a Hive 3.0 compatible one.

PR #201, opened 2019-08-16
[LIVY-639][REPL] add start time and completion time and duration to the statements web ui

OpenOpened:

add start time and completion time and duration to the statements web ui

JIRA: https://issues.apache.org/jira/browse/LIVY-639?filter=-2

## How was this patch tested?

Existing test set and our test environment

DeepinScreenshot_select-area_20190816153300

### DIFF
--- a/repl/src/main/scala/org/apache/livy/repl/Session.scala
+++ b/repl/src/main/scala/org/apache/livy/repl/Session.scala
@@ -160,13 +160,12 @@ class Session(
     val statement = new Statement(statementId, code, StatementState.Waiting, null)
     _statements.synchronized { _statements(statementId) = statement }
 
-    statement.started = System.currentTimeMillis()
-
     Future {
       setJobGroup(tpe, statementId)
       statement.compareAndTransit(StatementState.Waiting, StatementState.Running)
 
       if (statement.state.get() == StatementState.Running) {
+        statement.started = System.currentTimeMillis()
         statement.output = executeCode(interpreter(tpe), statementId, code)
       }
 
--- a/repl/src/main/scala/org/apache/livy/repl/Session.scala
+++ b/repl/src/main/scala/org/apache/livy/repl/Session.scala
@@ -17,7 +17,6 @@
 
 package org.apache.livy.repl
 
-
 import java.util.{LinkedHashMap => JLinkedHashMap}
 import java.util.Map.Entry
 import java.util.concurrent.Executors
@@ -148,10 +147,6 @@ class Session(
     _statements.toMap
   }
 
-  def nowTime(): Long = {
-    return System.currentTimeMillis()
-  }
-
   def execute(code: String, codeType: String = null): Int = {
     val tpe = if (codeType != null) {
       Kind(codeType)
@@ -165,8 +160,7 @@ class Session(
     val statement = new Statement(statementId, code, StatementState.Waiting, null)
     _statements.synchronized { _statements(statementId) = statement }
 
-    val start = System.currentTimeMillis()
-    statement.started = nowTime()
+    statement.started = System.currentTimeMillis()
 
     Future {
       setJobGroup(tpe, statementId)
@@ -179,7 +173,7 @@ class Session(
       statement.compareAndTransit(StatementState.Running, StatementState.Available)
       statement.compareAndTransit(StatementState.Cancelling, StatementState.Cancelled)
       statement.updateProgress(1.0)
-      statement.completed = nowTime()
+      statement.completed = System.currentTimeMillis()
     }(interpreterExecutor)
 
     statementId
@@ -227,7 +221,7 @@ class Session(
       }
 
       if (statement.state.get() == StatementState.Cancelled) {
-        statement.completed = nowTime()
+        statement.completed = System.currentTimeMillis()
         info(s"Statement $statementId cancelled.")
       }
     }(cancelExecutor)
--- a/repl/src/main/scala/org/apache/livy/repl/Session.scala
+++ b/repl/src/main/scala/org/apache/livy/repl/Session.scala
@@ -17,6 +17,8 @@
 
 package org.apache.livy.repl
 
+import java.text.SimpleDateFormat
+import java.time.LocalDateTime
 import java.util.{LinkedHashMap => JLinkedHashMap}
 import java.util.Map.Entry
 import java.util.concurrent.Executors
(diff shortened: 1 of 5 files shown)


repl/src/main/scala/org/apache/livy/repl/Session.scala
```diff
@@ -147,6 +149,12 @@ class Session(
     _statements.toMap
   }
 
+  def nowTime(): String = {
+    import java.time.format.DateTimeFormatter
+    val formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss")
+    return LocalDateTime.now().format(formatter)
```
yiheng: How do we handle the timezone? What is the timezone of the returned time representation?

My concern is if the server and web client are in different timezones, will there be some confusion?

rsc/src/main/java/org/apache/livy/rsc/driver/Statement.java
```diff
@@ -28,6 +28,9 @@
   @JsonRawValue
   public volatile String output;
   public double progress;
+  public String duration;
+  public String started;
+  public String completed;
```
jerryshao: It's not so efficient to use string to represent time internally, timestamp should be enough.

rsc/src/main/java/org/apache/livy/rsc/driver/Statement.java
```diff
@@ -56,4 +59,26 @@ public void updateProgress(double p) {
...
+    this.started = started;
+  }
+
+  public String getCompleted() {
+    return completed;
+  }
+
+  public void setCompleted(String completed) {
+    this.completed = completed;
+  }
```
jerryshao: As all the fields are public, it is not necessary to write all these setter/getters.

OpenOpened:
@jerryshao @yiheng I have modified the code. Time zone problem, I recorded the Greenwich time on the server, the time zone displayed is subject to the client.

repl/src/main/scala/org/apache/livy/repl/Session.scala
```diff
@@ -17,6 +17,7 @@
 
 package org.apache.livy.repl
 
+
```
jerryshao: useless blank line here.

repl/src/main/scala/org/apache/livy/repl/Session.scala
```diff
@@ -147,6 +148,10 @@ class Session(
     _statements.toMap
   }
 
+  def nowTime(): Long = {
+    return System.currentTimeMillis()
+  }
```
jerryshao: No need to add this method to simply wrap `System.currentTimeMillis()`.

rsc/src/main/java/org/apache/livy/rsc/driver/Statement.java
```diff
@@ -28,6 +28,9 @@
   @JsonRawValue
   public volatile String output;
   public double progress;
+  public double duration;
```
jerryshao: what's the purpose of using double instead of long?

repl/src/main/scala/org/apache/livy/repl/Session.scala
```diff
@@ -171,6 +179,9 @@ class Session(
       statement.compareAndTransit(StatementState.Running, StatementState.Available)
       statement.compareAndTransit(StatementState.Cancelling, StatementState.Cancelled)
       statement.updateProgress(1.0)
+      statement.completed = nowTime()
+      statement.duration = ((nowTime() - start) / 1000.0 / 60.0)
```
jerryshao: duration can be `statement.completed - statement.started`, why do we need a such conversion? I think a such conversion could let js to do it.

rsc/src/main/java/org/apache/livy/rsc/driver/Statement.java
OpenOpened: Because the unit is minutes and keeps a decimal

OpenOpened:
@jerryshao I have modified the code, thank you for your valuable suggestions.

server/src/main/resources/org/apache/livy/server/ui/static/js/session.js
```diff
@@ -34,6 +34,26 @@ function formatError(output) {
   return preWrap(errStr);
 }
 
+function timeWrap(time) {
+  time = (time / 1000.0 / 60.0).toFixed(1)
```
jerryshao: Usually statement maybe executed quite fast, I would suggest to use `second` rather than `minute`. More intelligently, we could change the unit automatically based on the time length. You could refer to Spark's code.

repl/src/main/scala/org/apache/livy/repl/Session.scala
```diff
@@ -160,6 +160,8 @@ class Session(
     val statement = new Statement(statementId, code, StatementState.Waiting, null)
     _statements.synchronized { _statements(statementId) = statement }
 
+    statement.started = System.currentTimeMillis()
+
     Future {
       setJobGroup(tpe, statementId)
       statement.compareAndTransit(StatementState.Waiting, StatementState.Running)
```
jerryshao: I think you should set `statement.started` here.

server/src/main/resources/org/apache/livy/server/ui/static/html/statements-table.html
```diff
@@ -56,6 +56,24 @@ <h4 id="statements-header" class="statements-template">Statements</h4>
         Output
       </span>
     </th>
+    <th>
+      <span data-toggle="tooltip"
+            title="The statement code of started time">
```
jerryshao: "The start time of statement code".

server/src/main/resources/org/apache/livy/server/ui/static/html/statements-table.html
```diff
@@ -56,6 +56,24 @@ <h4 id="statements-header" class="statements-template">Statements</h4>
...
+        Started
+      </span>
+    </th>
+    <th>
+      <span data-toggle="tooltip"
+            title="The statement code of completed time">
```
jerryshao: "The complete time of statement code".

server/src/main/resources/org/apache/livy/server/ui/static/html/statements-table.html
```diff
@@ -56,6 +56,24 @@ <h4 id="statements-header" class="statements-template">Statements</h4>
...
+        Completed
+      </span>
+    </th>
+    <th>
+      <span data-toggle="tooltip"
+            title="The statement code of duration time">
```
jerryshao: "The duration time of statement code".

OpenOpened:
@jerryshao please review, 3ks

rsc/src/main/java/org/apache/livy/rsc/driver/Statement.java
```diff
@@ -28,6 +28,8 @@
   @JsonRawValue
   public volatile String output;
   public double progress;
+  public long started = 0;
+  public long completed = 0;
```
jerryshao: Can you please also change the doc, since we now added two new fields.

OpenOpened:
@jerryshao document complete, by the way, complete the progress

PR #203, opened 2019-08-20
[hotfix][log] log stack trace properly

TisonKun:

Log stack trace properly. With an automatic check pass this is the only instance.

## How was this patch tested?

straightforward


TisonKun:
I can see a corresponding JIRA is necessary but given this is quite a nit change I am not sure whether we also create a dedicated JIRA. Let me know if it is expected.

zjffdu:
LGTM, I think it is fine without creating jira ticket for this trivial change.

TisonKun:
@zjffdu thanks for your reply. Shall we move forward?

jerryshao:
Thanks for the fix, merging to master branch.

TisonKun:
Thanks for your review!

PR #207, opened 2019-08-22
[LIVY-645]Add Session Name, Owner, Proxy User information to Web UI

yantzu:

1, Web UI, add Session Name to Interactive Sessions list
2, Web UI, add Session Name, Owner, Proxy User to Batch Sessions list
3,~~fix thrift server session doesn't set Session Name issue.~~ Move to PR #218

## How was this patch tested?

Update existing unit test, and have tested manually.

AddName
AddName2

### DIFF
--- a/server/src/main/scala/org/apache/livy/server/batch/BatchSessionServlet.scala
+++ b/server/src/main/scala/org/apache/livy/server/batch/BatchSessionServlet.scala
@@ -28,6 +28,8 @@ import org.apache.livy.utils.AppInfo
 case class BatchSessionView(
   id: Long,
   name: Option[String],
+  owner: String,
+  proxyUser: Option[String],
   state: String,
   appId: Option[String],
   appInfo: AppInfo,
@@ -73,8 +75,8 @@ class BatchSessionServlet(
       } else {
         Nil
       }
-    BatchSessionView(session.id, session.name, session.state.toString, session.appId,
-      session.appInfo, logs)
+    BatchSessionView(session.id, session.name, session.owner, session.proxyUser,
+      session.state.toString, session.appId, session.appInfo, logs)
   }
 
 }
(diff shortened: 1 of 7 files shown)


thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
```diff
@@ -228,7 +228,7 @@ class LivyThriftSessionManager(val server: LivyThriftServer, val livyConf: LivyC
       createInteractiveRequest.kind = Spark
       val newSession = InteractiveSession.create(
         server.livySessionManager.nextId(),
-        None,
+        createInteractiveRequest.name,
```
mgaido91: shall we do this in a separate PR and add UT for it, since this is a bug?

yantzu: OK, I will create separate PR.

jerryshao: @yantzu can you please revert the changes here, since you will create another PR to track this issue.

yantzu: Hi @jerryshao, I have submitted another commit to revert the change.

server/src/main/scala/org/apache/livy/server/batch/BatchSessionServlet.scala
```diff
@@ -73,8 +75,8 @@ class BatchSessionServlet(
       } else {
         Nil
       }
-    BatchSessionView(session.id, session.name, session.state.toString, session.appId,
-      session.appInfo, logs)
+    BatchSessionView(session.id, session.name, session.owner, session.proxyUser,
+      session.state.toString, session.appId, session.appInfo, logs)
```
jerryshao: I think you can match this with interactive session's view. 1. interactive session uses `null` for empty proxy user; 2. interactive session may have more field, maybe we can also add these in batch session.

yantzu: Hi @jerryshao , sorry for late. My consideration is:
1, I use Option[String] as proxyUser type is because BatchRecoveryMetadata's proxyUser is Option[String]. And name, appId are Option[String], my purpose it to make them consistent. Interactive session is a little different because its SessionInfo class is Java
2, The only field batch session doesn't have is "kind", which is a legacy field, batch session doesn't have the field information.

PR #208, opened 2019-08-22
[LIVY-643][Server] Support for a Livy Server running on Windows Operating System

davidOSUL:

The ability to start up a Livy server on a Windows Operating System. 
(https://issues.apache.org/jira/browse/LIVY-643)

This PR also partially resolves some of the test issues when building in a Windows environment as described in https://issues.apache.org/jira/browse/LIVY-611. It does not resolve all of the issues with running tests on a windows environment. For example integration tests will fail because of line 299 of MiniCluster.scala (sys.process.Process(s"pkill -f $simpleName")). I believe that with the changes in this PR, should be able to resolve LIVY-611 with a future PR. 

## How was this patch tested?

Unit tests for new class, and I was able to start up a livy server (with  java -cp "./jars/*;./conf;" org.apache.livy.server.LivyServer) on a Docker windows container and submit spark requests to the livy server successfully.

### DIFF
--- a/client-common/src/main/java/org/apache/livy/client/common/OperatingSystemUtils.java
+++ b/client-common/src/main/java/org/apache/livy/client/common/OperatingSystemUtils.java
@@ -1,3 +1,20 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


jerryshao:
@davidOSUL sorry for late response, can you please tell what are the left issues for Livy running on Windows?

>This PR also partially resolves some of the test issues when building in a Windows environment as described in https://issues.apache.org/jira/browse/LIVY-611.

davidOSUL:
@jerryshao And sorry for my late reply! I believe this PR should resolve issues with running the Livy server on Windows (at least with my tests). However it does not resolve *developing* on windows. LIVY-611 refers specifically to integration/unit tests. For example in the MiniCluster.scala integration test, some calls use unix based commands. However at the very least this PR should allow users to run a Livy server on a Windows machine, and with the changes it brings, it shouldn't be too hard to then resolve LIVY-611.

PR #209, opened 2019-08-23
[LIVY-640] Add tests for ThriftServer

captainzmc:

1、Added some tests in BinaryThriftServerSuite that ThriftServer is currently missing（The tests was Moved from spark ThriftServer）.
2、Set ENABLE_HIVE_CONTEXT=true in BinaryThriftServerSuite to support the creation of Hive tables and the creation of udf in test.
3、Upgrade spark2.2.0 to 2.2.3. To resolve the issue that session cannot be created during Travis test when ENABLE_HIVE_CONTEXT is set to true
4、Travis ITs is divided into two parts to avoid test timeout problems.

## How was this patch tested?

Test with Travis and see the results
image


captainzmc:
@jerryshao @mgaido91 @yiheng Please help to review this PR. The specific changes to the PR can be seen in the previous "What changes"

.travis.yml
```diff
@@ -23,17 +23,23 @@ language: scala
 matrix:
   include:
   - name: "Spark 2.2 Unit Tests"
-    env: MVN_FLAG='-Pthriftserver -DskipITs'
-  - name: "Spark 2.2 ITs"
+    env: MVN_FLAG='-Pthriftserver -Pintegration-test -Pcoverage -DskipITs'
+  - name: "Spark 2.2 ITs without integration-test and coverage"
```
yiheng: what about `Spark 2.2 Thrift ITs`?

.travis.yml
```diff
@@ -23,17 +23,23 @@ language: scala
...
     env: MVN_FLAG='-Pthriftserver -DskipTests'
+  - name: "Spark 2.2 ITs without thriftserver"
+    env: MVN_FLAG='-Pintegration-test -Pcoverage -DskipTests'
   - name: "Spark 2.3 Unit Tests"
-    env: MVN_FLAG='-Pspark-2.3 -Pthriftserver -DskipITs'
-  - name: "Spark 2.3 ITs"
+    env: MVN_FLAG='-Pspark-2.3 -Pthriftserver -Pintegration-test -Pcoverage -DskipITs'
+  - name: "Spark 2.3 ITs without integration-test and coverage"
```
yiheng: same with the above

.travis.yml
```diff
@@ -23,17 +23,23 @@ language: scala
...
     env: MVN_FLAG='-Pspark-2.3 -Pthriftserver -DskipTests'
+  - name: "Spark 2.3 ITs without thriftserver"
+    env: MVN_FLAG='-Pspark-2.3 -Pintegration-test -Pcoverage -DskipTests'
   - name: "Spark 2.4 Unit Tests"
-    env: MVN_FLAG='-Pspark-2.4 -Pthriftserver -DskipITs'
-  - name: "Spark 2.4 ITs"
+    env: MVN_FLAG='-Pspark-2.4 -Pthriftserver -Pintegration-test -Pcoverage -DskipITs'
+  - name: "Spark 2.4 ITs without integration-test and coverage"
```
yiheng: same with the above

.travis.yml
yiheng: so we do not generate coverage report on thrift integration test?

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -17,12 +17,21 @@
...
-import org.apache.hive.jdbc.HiveStatement
+import scala.collection.mutable.ArrayBuffer
+
+import org.apache.hive.jdbc.{HiveDriver, HiveStatement}
 
 import org.apache.livy.LivyConf
 
+object TestData {
+  def getTestDataFilePath(name: String): URL = {
+    Thread.currentThread().getContextClassLoader.getResource(s"data/files/$name")
```
yiheng: Can we use `getClass.getResource(s"data/files/$name")`?

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -119,13 +130,17 @@ trait CommonThriftTests {
     assert(tablesResultSet.getString(3) == "test_get_tables")
     assert(tablesResultSet.getString(4) == "TABLE")
     assert(!tablesResultSet.next())
+    statement.execute("DROP TABLE IF EXISTS test_get_tables")
```
yiheng: duplicated?

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -178,6 +193,8 @@ trait CommonThriftTests {
     assert(columnsResultSet.getString(22) == null)
     assert(columnsResultSet.getString(23) == "NO")
     assert(!columnsResultSet.next())
+    statement.execute("DROP TABLE IF EXISTS test_get_columns")
```
yiheng: duplicated?

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -259,6 +276,268 @@ trait CommonThriftTests {
 class BinaryThriftServerSuite extends ThriftServerBaseTest with CommonThriftTests {
   override def mode: ServerMode.Value = ServerMode.binary
   override def port: Int = 20000
+  // In BinaryThriftServerSuite, we set ENABLE_HIVE_CONTEXT=true to support the creation
+  // of Hive tables and the creation of udf.
+  livyConf.set(LivyConf.ENABLE_HIVE_CONTEXT, true)
```
yiheng: should we test with an `ENABLE_HIVE_CONTEXT = false` version?

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -259,6 +276,268 @@ trait CommonThriftTests {
...
+
+      val queries = Seq(
+        "DROP TABLE IF EXISTS test_map",
+        "DROP TABLE IF EXISTS db1.test_map2",
+        "DROP DATABASE IF EXISTS db1")
+
+      queries.foreach(statement.execute)
+    }
+
+  }
```
yiheng: why not test enable hive in http test cases?

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -259,6 +276,268 @@ trait CommonThriftTests {
...
+
+  test("test add jar and udf") {
+    withJdbcStatement { statement =>
+      val jarPath = "src/test/resources/TestUDTF.jar"
+      val jarURL = s"file://${System.getProperty("user.dir")}/$jarPath"
+      Seq(
+        "SET foo=bar",
+        s"ADD JAR $jarURL",
+        s"""CREATE TEMPORARY FUNCTION udtf_count2
+           |AS 'org.apache.spark.sql.hive.execution.GenericUDTFCount2'""".stripMargin
```
yiheng: The statement is not so long. Do we really need the stripMargin here?

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -259,6 +276,268 @@ trait CommonThriftTests {
...
+
+      queries.foreach(statement.execute)
+
+      assertResult(5, "Row count mismatch") {
+        val resultSet = statement.executeQuery("SELECT COUNT(*) FROM test")
+        resultSet.next()
+        resultSet.getInt(1)
+      }
+      statement.executeQuery("DROP TABLE IF EXISTS test")
+      statement.close()
```
yiheng: we do not need to close the statement here. The statement will be closed in the out of the withJdbcStatement scope. See [here](https://github.com/apache/incubator-livy/blob/master/thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerBaseTest.scala#L98)

.travis.yml
captainzmc: "Spark 2.2 ITs without integration-test and coverage" I will change to "Spark 2.2 Thrift ITs".
 coverage has dependencies on integration-test, so it needs to be taken out

.travis.yml
captainzmc: I will modify this place

.travis.yml
captainzmc: I will modify this place

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
captainzmc: After setting ENABLE_HIVE_CONTEXT=true, the tables created by the previous test are still visible in the next test. This can affect the results of unit tests. So at the end of each test, the currently created table needs to be deleted

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
captainzmc: Same as above

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
captainzmc: In spark-thrift tests ENABLE_HIVE_CONTEXT = true is default . this can override all test scenarios where ENABLE_HIVE_CONTEXT = false. And in livy-thrift, the IT test ran for nearly 40 minutes, and there was a risk of timeout. (Travis default timeout is 50 minutes)

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
captainzmc: yes，I will modify this place

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
captainzmc: This is consistent with spark. In the spark-thrift test, most of the tests were in BinaryServerSuite, and only a small number were in HttpServerSuite. Maybe they did it to prevent Travis from timeout.

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
captainzmc: use "getClass.getResource(s"data/files/$name")" I will get exception "LOAD DATA input path does not exist: null"

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
mgaido91: we are already testing the case when the hive context is supported. There are some tests which are skipped if it is not, but this happens only in specific builds. I am not sure we need this at all

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
mgaido91: that's not clear...seems like they are deleted twice, one at the beginning, if they exist, and once at the end...I don0t think both are needed

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -259,6 +276,268 @@ trait CommonThriftTests {
...
+    }
+  }
+
+  test("test add jar and udf") {
```
mgaido91: I am not sure about many of these tests: what are we testing? The point is: Livy leverages spark to execute all these operations. So the question is: are we testing the Livy thriftserver here or are we testing Spark?

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
captainzmc: yes，StripMargin can be removed here

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
captainzmc: Okay, I'm going to delete the previous drop statement

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
captainzmc: Okay, I'm going to delete the previous drop statement

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
captainzmc: We added this section because we found that none of livy-thrift's UTs had created hive tables for testing. But we can see this in spark-thrift UTs. So we added this for alignment

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
captainzmc: What I understand here is that we tested livy thriftserver to perform some common SQL usage.

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
mgaido91: well, I think Spark has to test Spark and Livy has to test Livy. I don't think we necessarily have to run all Spark tests. As you also noted, we are already pretty close to time limits, so useless tests are not harmless. My point is: are we testing a specific behavior of the Livy thriftserver? If so, then it is good to add the tests, otherwise no. For instance, if you create a hive table in a session, then you create another connection to Livy thrisever reusing the spark session and read from that table, you are testing Livy IMHO. Whiel if you just perform a LOAD DATA, you're testing Spark IMO.

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
mgaido91: Yes, because for instance we need to ensure all datatypes are properly written over the wire, that nulls are handled properly, etc.etc.

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
captainzmc: Get it, thanks Marco. I will delete some tests that are not related to livy. Keep only livy-related tests.

captainzmc:
Fix issue in code review, and only livy-related tests are retained.

.travis.yml
```diff
@@ -23,17 +23,23 @@ language: scala
 matrix:
   include:
   - name: "Spark 2.2 Unit Tests"
-    env: MVN_FLAG='-Pthriftserver -DskipITs'
-  - name: "Spark 2.2 ITs"
+    env: MVN_FLAG='-Pthriftserver -Pintegration-test -Pcoverage -DskipITs'
```
jerryshao: We use `skipITs` to check whether IT should be enabled, why do we need another profile `integration-test`?

pom.xml
```diff
@@ -1009,6 +1007,23 @@
   </reporting>
 
   <profiles>
+    <profile>
+      <id>integration-test</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
```
jerryshao: What's the purpose of making this as a profile and activate by default?

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -17,12 +17,21 @@
...
+
+import org.apache.hive.jdbc.{HiveDriver, HiveStatement}
 
 import org.apache.livy.LivyConf
 
+object TestData {
+  def getTestDataFilePath(name: String): URL = {
+    Thread.currentThread().getContextClassLoader.getResource(s"data/files/$name")
+  }
+  val smallKv = getTestDataFilePath("small_kv.txt")
```
jerryshao: Weird to add object field here, this will be initialized when class is loaded, this can be a local variable or just a method.

pom.xml
captainzmc: It's my fault. I added this for testing, but it didn't work. I'm going to delete this

.travis.yml
captainzmc: I confirmed that there are no Unit Tests in integration-test and coverage(Coverage has depends on integration-test). So I'm going to delete ntegration-test and coverage in this place.

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
captainzmc: Got it, I’ll change it here

captainzmc:
fix

.travis.yml
```diff
@@ -24,16 +24,22 @@ matrix:
   include:
   - name: "Spark 2.2 Unit Tests"
     env: MVN_FLAG='-Pthriftserver -DskipITs'
-  - name: "Spark 2.2 ITs"
+  - name: "Spark 2.2 Thrift ITs"
```
mgaido91: this seems to run only Thrift, which is not true. I think we can avoid this change, since you already specify `Without Thrift` when needed

pom.xml
```diff
@@ -1009,6 +1007,20 @@
   </reporting>
 
   <profiles>
+    <profile>
```
mgaido91: why do we need these 2 new profiles?

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
mgaido91: what if an exception happens?

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -259,6 +263,171 @@ trait CommonThriftTests {
...
+      assertResult(1, "Row count mismatch") {
+        val resultSet = statement.executeQuery("SELECT COUNT(*) FROM test_null WHERE key IS NULL")
+        resultSet.next()
+        resultSet.getInt(1)
+      }
+      statement.executeQuery("DROP TABLE IF EXISTS test_null")
+    }
+  }
+
+  test("Binary type support") {
```
mgaido91: shall we rather improve `dataTypesTest` in order to contain also a test for binary data type?

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -259,6 +263,171 @@ trait CommonThriftTests {
...
   override def port: Int = 20000
+  // In BinaryThriftServerSuite, we set ENABLE_HIVE_CONTEXT=true to support the creation
+  // of Hive tables.
+  livyConf.set(LivyConf.ENABLE_HIVE_CONTEXT, true)
+
+  def getTestDataFilePath(): URL = {
+    Thread.currentThread().getContextClassLoader.getResource("data/files/small_kv.txt")
+  }
+
+  test("result set containing NULL") {
```
mgaido91: what are you testing here? If the count with null values works? This is a Spark test rather than a Livy one...

.travis.yml
captainzmc: You're right. This place should not need to be highlighted. I'll change this to "Spark 2.2 ITs"

pom.xml
captainzmc: These two parts are added in order to separate the thriftserver and the ITs of integration-test(Coverage has depends on integration-test).  then manually specified with -p in travis.yml

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
captainzmc: I'm going to add a try finally here and put this part in finally

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
captainzmc: Yes, this can test in dataTypesTest.I'll change it here

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
captainzmc: And there is a overlap here with the dataTypesTest. I'm going to delete this test.

pom.xml
mgaido91: isn't it enough to use the existing `-DskipITs`?

captainzmc: Without the two profiles, both thriftserver and integration-test will be executed in ITs, which may cause timeouts.
So we need these two profiles. Then separate thriftserver and integration-test in travis.yml. 
 Like this:
-name: "Spark 2.2 ITs"
Env: MVN_FLAG = '- Pthriftserver - DskipTests'
- name: "Spark 2.2 ITs Without Thrift"
Env: MVN_FLAG = '- Pintegration - test - Pcoverage - DskipTests'

captainzmc: Without the two profiles, both thriftserver and integration-test will be executed in ITs, which may cause timeouts.
So we need these two profiles. Then separate thriftserver and integration-test in travis.yml.
Like this:
Env: MVN_FLAG = '- Pthriftserver - DskipTests'
Env: MVN_FLAG = '- Pintegration - test - Pcoverage - DskipTests'

mgaido91: ah I think now I got it. We need a way to execute only thrifserver ITs basically... I am wondering if there is a better way to do this...

captainzmc:
fix

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -259,6 +276,137 @@ trait CommonThriftTests {
...
+
+        data = buf1
+      }
+
+      // first session, we get the default value of the session status
+      withJdbcStatement { statement =>
+        val rs1 = statement.executeQuery(s"SET spark.sql.shuffle.partitions")
+        rs1.next()
+        defaultV1 = rs1.getString(1)
+        assert(defaultV1 != "200")
```
mgaido91: we should test the actual value, rather than what it is not

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -259,6 +276,137 @@ trait CommonThriftTests {
...
+        rs1.close()
+
+        val rs2 = statement.executeQuery("SET hive.cli.print.header")
+        rs2.next()
+
+        defaultV2 = rs2.getString(1)
+        assert(defaultV1 != "true")
```
mgaido91: ditto

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -259,6 +276,137 @@ trait CommonThriftTests {
...
+        rs2.close()
+      }
+
+      // second session, we update the session status
+      withJdbcStatement { statement =>
+        val queries = Seq(
+          s"SET spark.sql.shuffle.partitions=291",
```
mgaido91:
```suggestion
          "SET spark.sql.shuffle.partitions=291",
```

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -259,6 +276,137 @@ trait CommonThriftTests {
...
+        assert(defaultV1 === rs1.getString(1))
+        rs1.close()
+
+        val rs2 = statement.executeQuery("SET hive.cli.print.header")
+        rs2.next()
+        assert(defaultV2 === rs2.getString(1))
+        rs2.close()
+      }
+
+      // switch another database
```
mgaido91: seems this and the following are redundant

pom.xml
mgaido91: @jerryshao @vanzin any suggestion regarding this?

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -17,9 +17,13 @@
...
 package org.apache.livy.thriftserver
 
+import java.net.URL
 import java.sql.{Connection, Date, SQLException, Statement, Types}
 
-import org.apache.hive.jdbc.HiveStatement
+import scala.collection.mutable.ArrayBuffer
+import scala.io.Source
+
+import org.apache.hive.jdbc.{HiveDriver, HiveStatement}
```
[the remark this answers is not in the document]
captainzmc: My fault. I forgot to delete it after I used it as a test

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
captainzmc: Got it. I'll change this test

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
captainzmc: I’ll fix this

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
captainzmc: For testing multiple sessions, this section is really not relevant. I'll delete this part

captainzmc:
fix

pom.xml
jerryshao: If there's no timeout currently, I would suggest to not change here and in travis file, let's do it in a separate PR.

captainzmc: Now we really don't have a timeout problem after simplifying the test. I will restore this place. If a timeout occurs later, we can modify it in a separate PR

captainzmc:
fix

jerryshao:
+1, LGTM. @mgaido91 would you please help to take another review.

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -260,6 +277,104 @@ class BinaryThriftServerSuite extends ThriftServerBaseTest with CommonThriftTest
...
 
+  test("test multiple session") {
+    var defaultV1: String = null
+    var defaultV2: String = null
+    var data: ArrayBuffer[Int] = null
+    try {
+      // create table
+      withJdbcStatement { statement =>
+        val queries = Seq(
+          "CREATE TABLE test_map(key INT, value STRING) USING json",
```
mgaido91: I don't see where and when data is inserted into it, may you please explain me?

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -260,6 +277,104 @@ class BinaryThriftServerSuite extends ThriftServerBaseTest with CommonThriftTest
...
+        rs2.close()
+
+        assert(buf1 === buf2)
+
+        data = buf1
+      }
+
+      // first session, we get the default value of the session status
+      withJdbcStatement { statement =>
+        val rs1 = statement.executeQuery(s"SET spark.sql.shuffle.partitions")
```
mgaido91:
```suggestion
        val rs1 = statement.executeQuery("SET spark.sql.shuffle.partitions")
```

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -260,6 +277,104 @@ class BinaryThriftServerSuite extends ThriftServerBaseTest with CommonThriftTest
...
+      // create table
+      withJdbcStatement { statement =>
+        val queries = Seq(
+          "CREATE TABLE test_map(key INT, value STRING) USING json",
+          "CACHE TABLE test_table AS SELECT key FROM test_map ORDER BY key DESC")
+
+        queries.foreach(statement.execute)
+
+        val plan = statement.executeQuery("explain select * from test_table")
+        plan.next()
```
mgaido91: I am afraid that in a future version of spark, the plan may be a bit different and this test may fail. What about creating a string with the whole plan and then checking for the `InMemoryTableScan`?

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -260,6 +277,104 @@ class BinaryThriftServerSuite extends ThriftServerBaseTest with CommonThriftTest
...
+
+      // second session, we update the session status
+      withJdbcStatement { statement =>
+        val queries = Seq(
+          "SET spark.sql.shuffle.partitions=291",
+          "SET hive.cli.print.header=true"
+        )
+
+        queries.map(statement.execute)
+        val rs1 = statement.executeQuery(s"SET spark.sql.shuffle.partitions")
```
mgaido91:
```suggestion
        val rs1 = statement.executeQuery("SET spark.sql.shuffle.partitions")
```

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -260,6 +277,104 @@ class BinaryThriftServerSuite extends ThriftServerBaseTest with CommonThriftTest
...
+        rs2.next()
+        assert("hive.cli.print.header" === rs2.getString(1))
+        assert("true" === rs2.getString(2))
+        rs2.close()
+      }
+
+      // third session, we get the latest session status, supposed to be the
+      // default value
+      withJdbcStatement { statement =>
+        val rs1 = statement.executeQuery(s"SET spark.sql.shuffle.partitions")
```
mgaido91:
```suggestion
        val rs1 = statement.executeQuery("SET spark.sql.shuffle.partitions")
```

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -260,6 +277,104 @@ class BinaryThriftServerSuite extends ThriftServerBaseTest with CommonThriftTest
...
+          "CACHE TABLE test_table AS SELECT key FROM test_map ORDER BY key DESC")
+
+        queries.foreach(statement.execute)
+
+        val plan = statement.executeQuery("explain select * from test_table")
+        plan.next()
+        plan.next()
+        assert(plan.getString(1).contains("InMemoryTableScan"))
+
+        val rs1 = statement.executeQuery("SELECT key FROM test_table ORDER BY KEY DESC")
```
mgaido91: Shall we read form the table in a different connection? otherwise I don't see how this is useful to test multiple sessions...

captainzmc: In fact, I think this create table test hasn’t much to do with multiple sessions. And this test overlaps with other tests. I should delete this.

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -260,6 +277,66 @@ class BinaryThriftServerSuite extends ThriftServerBaseTest with CommonThriftTest
...
+      assert(defaultV1 === rs1.getString(2))
+      rs1.close()
+
+      val rs2 = statement.executeQuery("SET hive.cli.print.header")
+      rs2.next()
+      assert("hive.cli.print.header" === rs2.getString(1))
+      assert(defaultV2 === rs2.getString(2))
+      rs2.close()
+    }
+
```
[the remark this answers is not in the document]
captainzmc: Got it

mgaido91:
Thanks. Merging to master.

PR #211, opened 2019-08-23
[LIVY-649] Use HTTP method DELETE instead of POST for statement cancellation API

o-shevchenko:

Unify Livy REST API to use more common DELETE method instead of POST for statement cancellation API.
Before:
POST /sessions/{sessionId}/statements/{statementId}/cancel
After:
DELETE /sessions/{sessionId}/statements/{statementId}

https://issues.apache.org/jira/browse/LIVY-649

## How was this patch tested?
unit tests
manual verification:
image

### DIFF
--- a/docs/rest-api.md
+++ b/docs/rest-api.md
@@ -295,7 +295,7 @@ Returns a specified statement in a session.
 The <a href="#statement">statement</a> object.
 
 
-### POST /sessions/{sessionId}/statements/{statementId}/cancel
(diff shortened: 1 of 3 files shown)


yiheng:
Maybe we need to consider the API compatibility with the old version.

jerryshao:
Yes, agree with @yiheng , this will simply break the compatibility, I would not change this unless we have a sufficient reason.

o-shevchenko:
Can it be a part of some major release?

PR #212, opened 2019-08-24
[LIVY-11] Enable HA support

RogPodge:

This pull request enables Active/Passive HA in Livy through the use of Zookeeper. The CuratorElectorService class coordinates leadership election between multiple Livy instances

https://issues.apache.org/jira/browse/LIVY-11

## How was this patch tested?

unit tests
manual verification using 2 Livy instances


yantzu:
How does this pr related to pr #189 ?

OpenOpened:
Do you need a proxy-like thing to find the latest active node?

server/src/main/scala/org/apache/livy/server/CuratorElectorService.scala
```diff
@@ -0,0 +1,123 @@
...
+  val retryPattern = """\s*(\d+)\s*,\s*(\d+)\s*""".r
+  val retryPolicy = retryValue match {
+    case retryPattern(n, sleepMs) => new RetryNTimes(n.toInt, sleepMs.toInt)
+    case _ => throw new IllegalArgumentException(
+      s"$HA_KEY_PREFIX_CONF contains bad value: $retryValue. " +
+        "Correct format is <max retry count>,<sleep ms between retry>. e.g. 5,100")
+  }
+
+  val client: CuratorFramework = CuratorFrameworkFactory.newClient(haAddress, retryPolicy)
+  val leaderKey = "/$haKeyPrefix/leader"
```
lcaaaat: May be val leaderKey = s"/$haKeyPrefix/leader" ?

server/src/main/scala/org/apache/livy/server/CuratorElectorService.scala
```diff
@@ -0,0 +1,123 @@
...
+{
+
+  import CuratorElectorService._
+
+  val haAddress = livyConf.get(LivyConf.HA_ZOOKEEPER_URL)
+  require(!haAddress.isEmpty, s"Please config ${LivyConf.HA_ZOOKEEPER_URL.key}.")
+  val haKeyPrefix = livyConf.get(HA_KEY_PREFIX_CONF)
+  val retryValue = livyConf.get(HA_RETRY_CONF)
+  // a regex to match patterns like "m, n" where m and n both are integer values
+  val retryPattern = """\s*(\d+)\s*,\s*(\d+)\s*""".r
```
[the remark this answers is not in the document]
o-shevchenko: Hi, 
just FYI
I have moved this configuration to ZooKeeper manager in #189. Replaced with two separated configs.
Perhaps, we need to merge this PR after #189 or I need to refactor my code to use your approach if this PR will be merged first (but I think that two separated properties more clear then parsing one property).

yiheng:
How do the clients know the latest active node? Do they need to ask apache curator? If so, do we need to update the client code?

yiheng:
Another question is if the current leader crash and a new leader is elected, should the session state in the old leader be restored in the new leader?

OpenOpened:
> Another question is if the current leader crash and a new leader is elected, should the session state in the old leader be restored in the new leader?

The code does not do the state synchronization related to the Session State. Is it a good choice to put it in zk?

o-shevchenko:
> How do the clients know the latest active node? Do they need to ask apache curator? If so, do we need to update the client code?
> 
> When people use a restful client(e.g. curl) to communicate with Livy, should they query zk every time?

#189 implements Livy Server discovery to provide the ability to get current Livy Server address by client to use it in REST requests. There is Java/Scala API for it. If someone uses not Java world client then he should ask ZK to get the address.

RogPodge:
When HA mode is enabled, the following happens.

Once new Livy Server starts, it uses Curator to connect to the Zookeeper state store to try to acquire leadership. The process will wait until it is able to acquire leadership. The leader latch library will automatically ensure that only 1 Livy Server among those that are connect to that state store will be elected Leader at any point.

Once it acquires leadership, it starts the Livy Web Server in the same way that it's currently handled, **recovering session state** if it is configured and all other associated configurations.

The client ideally doesn't need to do anything with respect to contacting the zookeeper server, as ideally you have a load balancer of some sort failover controller to eventually link the client to the actively running Livy instance. This solution is **primarily concerned** with ensuring that given a group of Livy Servers deployed in HA mode, **exactly one** of them will be active and running at any one time.

shanyu:
> > Another question is if the current leader crash and a new leader is elected, should the session state in the old leader be restored in the new leader?
> 
> The code does not do the state synchronization related to the Session State. Is it a good choice to put it in zk?

This is already implemented previously. You can set livy.server.recovery.state-store to "zookeeper" and livy.server.recovery.state-store.url to zookeeper quorum. This PR mainly concerns the leader election part for the active-passive Livy Server.

server/src/main/scala/org/apache/livy/server/CuratorElectorService.scala
```diff
@@ -0,0 +1,123 @@
...
+
+  def start() : Unit = {
+    transitionToStandby()
+
+    client.start()
+    leaderLatch.start()
+    leaderLatch.await()
+
+    // This instance is now the leader. Joining the webserver to the main thread
+    info("starting join")
```
shanyu: This doesn't seem right. If networking issue caused current active server to be standby, then transitionToStandby() will be called and server will be stopped, and current server process will exit.

I think we should do "Thread.currentThread.join()" right after leaderLatch.start() to block forever, with try catch block, and if any exception occurs we transitionToStandby() and exit.

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperStateStore.scala
```diff
@@ -46,7 +46,7 @@ class ZooKeeperStateStore(
   }
 
   private val zkAddress = livyConf.get(LivyConf.RECOVERY_STATE_STORE_URL)
-  require(!zkAddress.isEmpty, s"Please config ${LivyConf.RECOVERY_STATE_STORE_URL.key}.")
+  require(!zkAddress.isEmpty, s"Please configure ${LivyConf.RECOVERY_STATE_STORE_URL.key}.")
```
shanyu: let's not make this typo change. there are other places using "config" as well.

conf/livy.conf.template
```diff
@@ -100,6 +100,16 @@
 # on user request and then livy server classpath automatically.
 # livy.repl.enable-hive-context =
 
+# High Availability mode of Livy. Possible values:
+# off: Default. Turn off High Availability. 
+# on: Livy uses Zookeeper as a state store to ensure a livy server is always available with the
+#     correct state.
+# Must set livy.server.ha.zookeeper-url to configure HA
+# livy.server.ha.mode = off
```
ajbozarth: is `off`/`on` how HA configs usually work? I feel like `livy.server.ha.mode.enabled = true|false` would be more consistent with the current configs. This isn't an issue, just an observation

server/src/main/scala/org/apache/livy/server/CuratorElectorService.scala
```diff
@@ -0,0 +1,122 @@
...
+    case retryPattern(n, sleepMs) => new RetryNTimes(n.toInt, sleepMs.toInt)
+    case _ => throw new IllegalArgumentException(
+      s"$HA_KEY_PREFIX_CONF contains bad value: $retryValue. " +
+        "Correct format is <max retry count>,<sleep ms between retry>. e.g. 5,100")
+  }
+
+  val client: CuratorFramework = CuratorFrameworkFactory.newClient(haAddress, retryPolicy)
+  val leaderKey = s"/$haKeyPrefix/leader"
+
+  var server : LivyServer = livyServer
```
ajbozarth: no space before `:`

server/src/main/scala/org/apache/livy/server/CuratorElectorService.scala
```diff
@@ -0,0 +1,122 @@
...
+    type HAState = Value
+    val Active, Standby = Value
+  }
+  var currentState = HAState.Standby
+
+  def isLeader() {
+    transitionToActive();
+  }
+
+  def notLeader(){
```
ajbozarth: space before `{`

server/src/main/scala/org/apache/livy/server/CuratorElectorService.scala
```diff
@@ -0,0 +1,122 @@
...
+    transitionToStandby();
+  }
+
+  def start() : Unit = {
```
ajbozarth: no space before `:`

server/src/main/scala/org/apache/livy/server/CuratorElectorService.scala
```diff
@@ -0,0 +1,122 @@
...
+    leaderLatch.start()
+
+    try {
+      Thread.currentThread.join()
+    } finally {
+      transitionToStandby()
+    }
+  }
+
+  def close() : Unit = {
```
ajbozarth: no space before `:`

server/src/main/scala/org/apache/livy/server/CuratorElectorService.scala
```diff
@@ -0,0 +1,122 @@
...
+    transitionToStandby();
+    leaderLatch.close();
+  }
+
+  def transitionToActive() : Unit = {
```
ajbozarth: no space before `:`

server/src/main/scala/org/apache/livy/server/CuratorElectorService.scala
```diff
@@ -0,0 +1,122 @@
...
+    info("Transitioning to Active state")
+    if(currentState == HAState.Active){
```
ajbozarth: space before `{`

server/src/main/scala/org/apache/livy/server/CuratorElectorService.scala
```diff
@@ -0,0 +1,122 @@
...
+      info("Already in Active State")
+    }
+    else {
+      server.start()
+      currentState = HAState.Active
+      info("Transition complete")
+    }
+  }
+
+  def transitionToStandby() : Unit = {
```
ajbozarth: no space before `:`

server/src/main/scala/org/apache/livy/server/CuratorElectorService.scala
```diff
@@ -0,0 +1,122 @@
...
+    info("Transitioning to Standby state")
+    if(currentState == HAState.Standby){
```
ajbozarth: space before `{`

server/src/main/scala/org/apache/livy/server/LivyServer.scala
```diff
@@ -394,16 +395,29 @@ class LivyServer extends Logging {
...
   def main(args: Array[String]): Unit = {
     val server = new LivyServer()
-    try {
-      server.start()
-      server.join()
-    } finally {
-      server.stop()
+    val livyConf = new LivyConf().loadFromFile("livy.conf")
+
+    if(livyConf.get(LivyConf.HA_MODE) == HighAvailabilitySettings.HA_ON){
```
ajbozarth: space before `{`

ajbozarth (commented):
I don't have the knowledge to give the functionality of the code a proper review, but I did go through and check the code for consistency with the current codebase's style and naming conventions. 

I will defer to another committer to check the code functionality and eventual merging

yiheng:
> > How do the clients know the latest active node? Do they need to ask apache curator? If so, do we need to update the client code?
> > When people use a restful client(e.g. curl) to communicate with Livy, should they query zk every time?
> 
> #189 implements Livy Server discovery to provide the ability to get current Livy Server address by client to use it in REST requests. There is Java/Scala API for it. If someone uses not Java world client then he should ask ZK to get the address.

Another solution is we can put all active/standby services behind a load balancer. The client only needs to communicate with that load balancer. Then there's no need to change the client-side code. The server switch should be transparent to the client.

conf/livy.conf.template
yiheng: This patch implement active-standby HA. Should we change the possible values to:
1. off
2. active-standby

As in the future, we may implement other types of HA mode, e.g. active-active.

RogPodge:
Can someone help get this feature approved and checked in?

@jerryshao

jerryshao:
How do you do state transfer from old server to newly elected  livy server? How do you do server discovery? How do you do session recovery? I was wondering such a simple solution is good enough. Also there's no UT and IT for the code, I don't think it is ready to merge.

RogPodge:
State transfer and session recovery are handled by the pre-existing livy.server.recovery.mode and livy.server.recovery.state-store components, provided they are configured to a Zookeeper instance. 

Server discovery is decentralized though, would it be better to have an explicit configuration? I feel like having that wouldn't give any user-side benefits, given that there is no external command line tool to check the status of Livy instances.

RogPodge:
Livy HA with redirection:

In this pull request we are adding Active/Standby High Availability (HA) to Livy Server. The goal of HA is to ensure that amoung a group of Livy Servers, only 1 is Active at 
any given time while the others are on Standby. Our changes are as following:

Leveraging Zookeeper to handle Leader Election:
In order to achieve HA, we need a service to handle keeping track of the current leader. To do this, we created a CuratorElectorService class which will register the Livy Server 
instance into the Leader election group and keep track of the leader. This class leverages the Apache Curator LeaderLatch Library to keep track of the current leader. 
This class is also responsible for restarting the Livy Webserver when it transitions from Standby state to Active. This restart is where the Livy Webserver performs session recovery.



Redirecting to the active webserver:
With the CuratorElectorService class, Livy Server can now determine whether it is Active or on Standby. If it is active, it proceeds as usual. However, if it is on standby, it will
use the configuration in livy.conf to determine the endpoint to redirect the request to.


Creating configurations to enable HA:
We added user configurations to enable HA. These include specifying whether or not HA is enabled, the endpoints of the Livy Server instances used for HA, and the zookeeper instances 
used for leader election.



Splitting Livy's start() function to init() and start():
Currently, Livy's start() function is overloaded to both initalize all of its components (the webserver, its servlets, and session managers) and then safely start them. This PR
separates the start() function into an init() component and a start() components. This is to allow us to restart the session managers and trigger session recovery without having
to safely tear down and rebuild the Livy Webserver every time. 

@jerryshao can I get some eyes for a code review/suggestions?

RogPodge:
@jerryshao small ping for a review

RogPodge:
@jerryshao Another ping

RogPodge:
@jerryshao pinging again

RogPodge:
@jerryshao another ping

RogPodge:
@jerryshao

RogPodge:
@jerryshao

RogPodge:
@jerryshao

jameschen1519:
Taking over the pull request with @RogPodge 's permission. A bug involving query omission on failover was fixed, along with a bit of minor refactoring and renaming.

I'm not too familiar with the code review structure of this repo, but can someone take a look at the PR and its changes? (Pinging @jerryshao based on prior messages as well)

PR #215, opened 2019-08-26
[LIVY-586] Fix batch state from starting to dead when startup fail

runzhiwang:

Fix batch state from starting to dead when startup fail

## How was this patch tested?

Create batch session with error parameter, the batch state will turn from starting to dead.

### DIFF
--- a/server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
+++ b/server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
@@ -151,6 +151,10 @@ class SparkYarnApp private[utils] (
     }
   }
 
+  private def isProcessErrExit(): Boolean = {
+    process.isDefined && !process.get.isAlive && process.get.exitValue() != 0
+  }
+
   private def changeState(newState: SparkApp.State.Value): Unit = {
     if (state != newState) {
       listener.foreach(_.stateChanged(state, newState))
@@ -171,6 +175,10 @@ class SparkYarnApp private[utils] (
       appTag: String,
       pollInterval: Duration,
       deadline: Deadline): ApplicationId = {
+    if (isProcessErrExit()) {
+      throw new IllegalStateException("spark-submit start failed")
+    }
+
     val appTagLowerCase = appTag.toLowerCase()
 
     // FIXME Should not loop thru all YARN applications but YarnClient doesn't offer an API.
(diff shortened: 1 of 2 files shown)


server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -169,6 +169,10 @@ class SparkYarnApp private[utils] (
       appTag: String,
       pollInterval: Duration,
       deadline: Deadline): ApplicationId = {
+    if (process.isDefined && !process.get.isAlive && process.get.exitValue() != 0) {
```
yiheng: create a unit test to cover this case?

runzhiwang: okay

runzhiwang: @yiheng UT has been added.

yiheng:
+1

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
jerryshao: Seems a bit weird to add check here in this method, maybe we can add such logic in `yarnAppMonitorThread`.

runzhiwang: @jerryshao The process is alive when the yarnAppMonitorThread was created, so I cannot check it in the begining of yarnAppMonitorThread. The process turns to dead when getAppIdFromTag which always fail and cost 120 seconds, so I check it in getAppIdFromTag  to make the state change as soon as possible.

jerryshao: I see. Can we please make such check as a method, I saw there's another place you changed previously which has the same check.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -151,6 +151,13 @@ class SparkYarnApp private[utils] (
     }
   }
 
+  private def checkProcessErrExit(): Boolean = {
+    if (process.isDefined && !process.get.isAlive && process.get.exitValue() != 0) {
+      return true
+    }
+    return false
+  }
```
jerryshao:
```
private def isProcessErrExit(): Boolean = {
    process.isDefined && !process.get.isAlive && process.get.exitValue() != 0
}
```

runzhiwang: okay

runzhiwang: @jerryshao I have updated the commit.

PR #218, opened 2019-08-27
[LIVY-652]Thrifserver doesn't set session name correctly

yantzu:

Thriftserver should set session name as the value passed in livy.session.name , but it current always set it NONE

## How was this patch tested?

add IT

https://issues.apache.org/jira/browse/LIVY-652

### DIFF
--- a/thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
+++ b/thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
@@ -404,10 +404,13 @@ class BinaryThriftServerSuite extends ThriftServerBaseTest with CommonThriftTest
     withJdbcConnection("default", Seq(s"livy.session.name=${testSessionName}")) { c =>
       // execute a statement and block until session is ready
       val statement = c.createStatement()
-      statement.executeQuery("select current_database()")
-      statement.close()
+      try {
+        statement.executeQuery("select current_database()")
+      } finally {
+        statement.close()
+      }
 
-      assert(!livySessionManager.get(testSessionName).isEmpty)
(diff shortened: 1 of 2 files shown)


thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -396,6 +396,20 @@ class BinaryThriftServerSuite extends ThriftServerBaseTest with CommonThriftTest
...
   }
+
+  test("LIVY-652: should set session name correctly") {
+    val livySessionManager = LivyThriftServer.getInstance.get.livySessionManager
+    val testSessionName = "MySessionName"
+    assert(livySessionManager.get(testSessionName).isEmpty)
+    withJdbcConnection("default", Seq(s"livy.session.name=${testSessionName}")) { c =>
+      // execute a statement and block until session is ready
+      val statement = c.createStatement()
+      statement.executeQuery("select current_database()")
```
jerryshao: Please wrap this with `try...finally` like other tests.

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -396,6 +396,20 @@ class BinaryThriftServerSuite extends ThriftServerBaseTest with CommonThriftTest
...
+      statement.close()
+
+      assert(!livySessionManager.get(testSessionName).isEmpty)
```
mgaido91: shall we rather test that the value is the expected one?

yantzu:
@jerryshao @mgaido91 fix code review, please take a look.

PR #219, opened 2019-08-27
[LIVY-653][TESTS] scalatest 3.0.8

pjfanning:

https://issues.apache.org/jira/browse/LIVY-653

Upgrading test jars.

## How was this patch tested?

unit tests run

### DIFF
--- a/core/src/test/scala/org/apache/livy/LivyBaseUnitTestSuite.scala
+++ b/core/src/test/scala/org/apache/livy/LivyBaseUnitTestSuite.scala
@@ -17,9 +17,9 @@
 
 package org.apache.livy
 
-import org.scalatest.{Outcome, Suite}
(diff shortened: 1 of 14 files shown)


jerryshao:
Hi @pjfanning thanks a lot for the contribution, may I know what's the purpose of this change?

pjfanning:
@jerryshao I was thinking that updating scalatest would make it easier to support a scala 2.12 build -- and ultimately, a scala 2.13 build

jerryshao:
I see. I would suggest to do Scala 2.12 support first, then we could upgrade the scalatest version.

pjfanning:
@jerryshao one other reason to upgrade scalatest sooner rather than later is that v3.0.8 is maintained while v2 is not.

PR #220, opened 2019-08-28
[LIVY-654] Ability to validate string before creating state/kind of session/statement

o-shevchenko:
https://issues.apache.org/jira/browse/LIVY-654

Kind.scala, SessionState.scala and StatementState.java don't include a method for string validation. It might be very useful to verify if some string can be converted to an enum type.

For example, to parse JSON string we need to write something like this (to prevent IllegalStateException):
```
Try {
  value match {
  case JsString(state) => SessionState(state)
  case _ => throw MyException
  }
} 
match {
 case Success(state) => state
 case Failure(ex) => throw MyException
}
```
it can be simplified to:
```
value match {
  case JsString(state) if SessionState.isValid(state) => SessionState(state)
  case _ => throw MyException
}
```

The second problem is StatementState.java enum use Pascal case instead of upper case
https://github.com/apache/incubator-livy/blob/412ccc8fcf96854fedbe76af8e5a6fec2c542d25/rsc/src/main/java/org/apache/livy/rsc/driver/StatementState.java#L26

For example, Waiting instead of WAITING. This isn't fit Java code style.

In this case, we can't use StatementState.valueOf("waiting".toUpperCase()) as usually but

StatementState.valueOf(state.toLowerCase().capitalize). Not sure if we can change it now.

## How was this patch tested?
Existed tests

### DIFF
--- a/core/src/main/scala/org/apache/livy/sessions/Kind.scala
+++ b/core/src/main/scala/org/apache/livy/sessions/Kind.scala
@@ -45,6 +45,12 @@ object Kind {
     case "sql" => SQL
     case other => throw new IllegalArgumentException(s"Invalid kind: $other")
   }
+
+  val kinds: Seq[Kind] = Seq(Spark, PySpark, SparkR, Shared, SQL)
+
+  def isValid(kind: String): Boolean = {
+    kinds.map(_.name).contains(kind)
+  }
 }
 
 class SessionKindModule extends SimpleModule("SessionKind") {
--- a/core/src/main/scala/org/apache/livy/sessions/SessionState.scala
+++ b/core/src/main/scala/org/apache/livy/sessions/SessionState.scala
@@ -30,13 +30,13 @@ class FinishedSessionState(
 object SessionState {
 
   def apply(s: String): SessionState = s match {
-    case "not_started" => NotStarted
-    case "starting" => Starting
-    case "recovering" => Recovering
-    case "idle" => Idle
-    case "running" => Running
-    case "busy" => Busy
-    case "shutting_down" => ShuttingDown
+    case "not_started" => NotStarted()
+    case "starting" => Starting()
+    case "recovering" => Recovering()
+    case "idle" => Idle()
+    case "running" => Running()
+    case "busy" => Busy()
+    case "shutting_down" => ShuttingDown()
     case "error" => Error()
     case "dead" => Dead()
     case "killed" => Killed()
@@ -44,19 +44,19 @@ object SessionState {
     case _ => throw new IllegalArgumentException(s"Illegal session state: $s")
   }
 
-  object NotStarted extends SessionState("not_started", true)
+  case class NotStarted() extends SessionState("not_started", true)
 
-  object Starting extends SessionState("starting", true)
+  case class Starting() extends SessionState("starting", true)
 
-  object Recovering extends SessionState("recovering", true)
+  case class Recovering() extends SessionState("recovering", true)
 
-  object Idle extends SessionState("idle", true)
+  case class Idle() extends SessionState("idle", true)
 
-  object Running extends SessionState("running", true)
+  case class Running() extends SessionState("running", true)
 
-  object Busy extends SessionState("busy", true)
+  case class Busy() extends SessionState("busy", true)
 
-  object ShuttingDown extends SessionState("shutting_down", false)
+  case class ShuttingDown() extends SessionState("shutting_down", false)
 
   case class Killed(override val time: Long = System.nanoTime()) extends
     FinishedSessionState("killed", false, time)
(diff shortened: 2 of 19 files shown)


core/src/main/scala/org/apache/livy/sessions/Kind.scala
```diff
@@ -45,6 +45,12 @@ object Kind {
     case "sql" => SQL
     case other => throw new IllegalArgumentException(s"Invalid kind: $other")
   }
+
+  val kinds: Seq[Kind] = Seq(Spark, PySpark, SparkR, Shared, SQL)
```
o-shevchenko: Can we also move object SQL and other inside object Kind to be able to use:
`import org.apache.livy.sessions.Kind.Spark` instead of `import org.apache.livy.sessions.Spark` 
It looks like more common.
Is it critical for backward compatibility?

rsc/src/main/java/org/apache/livy/rsc/driver/StatementState.java
```diff
@@ -73,13 +74,18 @@ static void put(StatementState key,
...
   }
 
-  static boolean isValid(StatementState from, StatementState to) {
+  static boolean isAllowed(StatementState from, StatementState to) {
     return PREDECESSORS.get(to).contains(from);
   }
 
+  public static boolean isValid(String state) {
+    return Arrays.stream(values())
+            .map(x -> StringUtils.capitalize(x.state)).anyMatch(state::equals);
```
o-shevchenko: Can we change naming (upper case instead of Pascal case) for enum values. For example, `WAITING` instead of `Waiting`. This isn't fit Java code style.
Also, in this case, we can't use `StatementState.valueOf("waiting".toUpperCase())` as usually but

`StatementState.valueOf(state.toLowerCase().capitalize)`.

core/src/main/scala/org/apache/livy/sessions/Kind.scala
o-shevchenko: Is better to use `values` everywhere (instead of `kinds`/`states`)?

core/src/main/scala/org/apache/livy/sessions/SessionState.scala
```diff
@@ -30,33 +30,33 @@ class FinishedSessionState(
...
+  case class Idle() extends SessionState("idle", true)
 
-  object Running extends SessionState("running", true)
+  case class Running() extends SessionState("running", true)
 
-  object Busy extends SessionState("busy", true)
+  case class Busy() extends SessionState("busy", true)
 
-  object ShuttingDown extends SessionState("shutting_down", false)
+  case class ShuttingDown() extends SessionState("shutting_down", false)
```
jerryshao: What's purpose of changes here to switch to case class?

o-shevchenko: To unify access to enumeration type and to be able to create sequents of all states
```
val states: Seq[SessionState] = Seq(NotStarted(), Starting(), Recovering(), Idle(), Running(),
    Busy(), ShuttingDown(), Killed(), Error(), Dead(), Success())
```

jerryshao: The changes here will create many small `SessionState` objects, which is unnecessary and can be replaced by singleton.

o-shevchenko: we need to use either object or class to be able to create a sequence of states. as far as I see,  you changed case class to object for FinishedSessionState classes to fix the problem with time-variable initialization, so we can't use singleton everywhere. so, I thought this is not critical to use case class instead of object for other states.

core/src/main/scala/org/apache/livy/sessions/Kind.scala
```diff
@@ -45,6 +45,12 @@ object Kind {
...
+
+  def isValid(kind: String): Boolean = {
```
jerryshao: Where did this get called?

o-shevchenko: this is public API to use as I described in the description

jerryshao: Then how do users use this public api?

o-shevchenko: to verify that we can create StateSession object from a string without additional logic for catching exceptions.
There is the example in the description.

o-shevchenko: you need to add livy-core dependency in your project if you asking about it

jerryshao: Why do you need to create StateSession or Kind object?

o-shevchenko: to simply use it in the code instead of hardcoding string values like "idle", sometimes I need to check session state or wait for state, for example, create a session and wait for status "idle" to be able to run statement. I don't want to use string variables everywhere. I need an enumeration. I can create my own in my code but better to use it from Livy. it is the typical case, isn't it?

jerryshao: Yes, that's the case. But from my point I never expect to expose this to users

o-shevchenko: hmm, having a method to check if the object is a part of the enum is common practice for Scala, it can simplify using of your enum to don't force user to catch exception or write own method like to check if object exists in enum values (if exist, we don't have it now in Kind.scala, SessionState.scala that's why I have added it)

PR #221, opened 2019-09-01
[LIVY-519][TEST] Fix travis failed on should kill yarn app

runzhiwang:

Fix travis failed on "should kill yarn app"

1. When create SparkYarnApp, the yarnAppMonitorThread will be created, which change app state to Failed. Because before recent commit https://github.com/apache/incubator-livy/commit/a90f4fac8be27a38cc961c24043a494a739ff188, the pair <RUNNING: getYarnApplicationState, SUCCEEDED: getFinalApplicationStatus> which was mocked in test, but was not defined in mapYarnState, so the state of app will be changed to failed.

2. Then the test kills app, which will call killApplication when the app is running. However the app has been changed to failed in step 1, so killApplication won't be called, and verify(mockYarnClient).killApplication(appId) failed.

3. So if yarnAppMonitorThread changes app state before main thread kills app, the test will failed. If not, the test will succeed.

4. Though the recent commit https://github.com/apache/incubator-livy/commit/a90f4fac8be27a38cc961c24043a494a739ff188 fixed the bug accidentally, it is necessary to ensure the app is running before kill app.

## How was this patch tested?

Existed UT and IT.

### DIFF
--- a/server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
+++ b/server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
@@ -209,7 +209,8 @@ class SparkYarnApp private[utils] (
       .getOrElse(IndexedSeq.empty)
   }
 
-  def isRunning: Boolean = {
+  // Exposed for unit test.
+  private[utils] def isRunning: Boolean = {
     state != SparkApp.State.FAILED && state != SparkApp.State.FINISHED &&
       state != SparkApp.State.KILLED
   }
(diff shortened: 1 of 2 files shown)


server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -209,7 +209,7 @@ class SparkYarnApp private[utils] (
       .getOrElse(IndexedSeq.empty)
   }
 
-  private def isRunning: Boolean = {
+  def isRunning: Boolean = {
```
yiheng: Is it better to change the scope to package-private? e.g.
```scala
private[utils] def isRunning: Boolean = {
```

runzhiwang: okay

runzhiwang: @yiheng PR has been updated.

jerryshao:
Run the test again to avoid flakiness.

PR #222, opened 2019-09-02
[LIVY-647][TEST]Fix travis failed on batch session should not be gc-ed until application is finished

runzhiwang:

Fix travis failed on "batch session should not be gc-ed until application is finished"

1. When create BatchSessionManager, the GarbageCollector thread will be created, which collect garbage according to session.state. However the session was mocked, and the test thread execute doReturn(s).when(session).state several times, if the test thread do half of the statement doReturn(s).when(session).state, and GarbageCollector thread check session.state at this time, the exception will threw by GarbageCollector thread.

2. So the fix avoid executing doReturn(s).when(session).state after session has been registered into SessionManager.

## How was this patch tested?

Existed UT and IT.


jerryshao:
Run the test again.

PR #223, opened 2019-09-02
[LIVY-658]RSCDriver should catch exception if cancel job failed during shutdown

yantzu:

Currently, if startup meet exception, exception will trigger spark to shutdown, then trigger cancel job, but cancel job will throw another exception due to spark is not initialized. The new exception will swallow the old exception.

https://issues.apache.org/jira/browse/LIVY-658

Before changes:
cancel job exception

After changes:
cancel job exception after fix

## How was this patch tested?

Tested manually， and  add unit test.

### DIFF
--- a/rsc/src/test/java/org/apache/livy/rsc/driver/TestRSCDriver.java
+++ b/rsc/src/test/java/org/apache/livy/rsc/driver/TestRSCDriver.java
@@ -1,0 +1,37 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.livy.rsc.driver;
+
+import org.apache.spark.SparkConf;
+
+import org.junit.Test;
+
+import org.apache.livy.rsc.BaseProtocol;
+import org.apache.livy.rsc.RSCConf;
+
+public class TestRSCDriver {
+  @Test(expected = IllegalArgumentException.class)
+  public void testCancelJobFailedAfterInitializeFailedShouldThrowOriginalException()
+      throws Exception {
+    //use empty Conf to trigger initialize throw IllegalArgumentException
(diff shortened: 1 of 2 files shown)


jerryshao:
CC @yiheng @runzhiwang please help to review.

runzhiwang:
It's easy to find the exception was caused by canceling job whether catch or not ?

yantzu:
Sorry @runzhiwang , I cannot catch your point, could you please explain more?

runzhiwang:
@yantzu If do not catch the exception, it's also easy to find cancel job will cause exception. Is it necessary to catch ?

yantzu:
Without catch cancel exception, the original exception("scala.refrect.internal.FactalError: object Predef does not have a member classOf" in the screen-shoot) is concealed by NullPointerException, it is not easy to find application error root cause(which is FactalError in this case). @runzhiwang

runzhiwang:
@yantzu Is the original exception helpful to find the reason is fail to startup.

yantzu:
original exception is the cause of application failure. @runzhiwang

yiheng:
can we fix the annoying `NullPointerException`?

yantzu:
The unit test I add demonstrate how NullPointerException happens: a request send to RSCDriver while it is initializing, after initialization failed, RSCDriver try to cancel job. In real case, it is beeline send a "use default" SQL to thrift server then RSCDriver. 
I think it is possible to avoid NPE in this case, but not sure if it is what we should do, because:
1, there may be other reason cause cancel job failure, to fix all those would be diffcult
2, catch and log exception when cancel job give RSCDriver a chance to cancel other job and shutdown context correctly
3, RSCDriver already catch shutdownContext() and shutdownServer() call

jerryshao:
Retrigger the test.

rsc/src/test/java/org/apache/livy/rsc/driver/TestRSCDriver.java
```diff
@@ -0,0 +1,37 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.livy.rsc.driver;
+
+import org.apache.spark.SparkConf;
+
```
jerryshao: Remove this blank line.

rsc/src/test/java/org/apache/livy/rsc/driver/TestRSCDriver.java
```diff
@@ -0,0 +1,37 @@
...
+import org.junit.Test;
+
+import org.apache.livy.rsc.BaseProtocol;
+import org.apache.livy.rsc.RSCConf;
+
+public class TestRSCDriver {
+  @Test(expected = IllegalArgumentException.class)
+  public void testCancelJobFailedAfterInitializeFailedShouldThrowOriginalException()
```
jerryshao: Can you please shorten the method name?

yantzu: fixed

jerryshao:
LGTM, merging to master branch.

PR #224, opened 2019-09-04
[LIVY-633][Server] session should not be gc-ed for long running queries

yiheng:
Currently, Livy records the last activity time of the session before statement execution. If a statement runs too long, exceeding then the session timeout, the session will be garbage collected after the statement execution.

This should not be the expected behavior. The statement execution time should not be count into idle. We should update the last activity time after the statement execution.

We cannot be updated when session changes state from busy to idle in the Session class. So in this patch, we add a replLastActivity field into the rscClient, which will be updated when the repl state changes. So when session changes its state from busy to idle, this field will catch the time and finally reflect on the session last activity.

## How was this patch tested?
Manual test. Also, add a new unit test.

Existing unit tests and integration tests.

### DIFF
--- a/rsc/pom.xml
+++ b/rsc/pom.xml
@@ -49,6 +49,11 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.livy</groupId>
+      <artifactId>livy-core_${scala.binary.version}</artifactId>
+      <version>${project.version}</version>
+    </dependency>
 
     <dependency>
       <groupId>com.esotericsoftware.kryo</groupId>
--- a/rsc/pom.xml
+++ b/rsc/pom.xml
@@ -53,6 +53,8 @@
       <groupId>org.apache.livy</groupId>
       <artifactId>livy-core_${scala.binary.version}</artifactId>
       <version>${project.version}</version>
+      <!-- Or it will conflict with repl jars -->
+      <scope>provided</scope>
     </dependency>
 
     <dependency>
--- a/rsc/src/main/java/org/apache/livy/rsc/RSCClient.java
+++ b/rsc/src/main/java/org/apache/livy/rsc/RSCClient.java
@@ -34,7 +34,6 @@ import io.netty.channel.nio.NioEventLoopGroup;
 import io.netty.util.concurrent.GenericFutureListener;
 import io.netty.util.concurrent.ImmediateEventExecutor;
 import io.netty.util.concurrent.Promise;
-import org.apache.livy.sessions.SessionState;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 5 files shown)


rsc/src/main/java/org/apache/livy/rsc/RSCClient.java
```diff
@@ -411,6 +425,10 @@ private void handle(ChannelHandlerContext ctx, JobStarted msg) {
 
     private void handle(ChannelHandlerContext ctx, ReplState msg) {
       LOG.trace("Received repl state for {}", msg.state);
+      // Update last activity timestamp when there's state change.
+      if (replState == null || !replState.equals(msg.state)) {
```
jerryshao: Shall we only track the state from `busy` to others?

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSession.scala
```diff
@@ -626,4 +626,16 @@ class InteractiveSession(
   }
 
   override def infoChanged(appInfo: AppInfo): Unit = { this.appInfo = appInfo }
+
+  override def lastActivity: Long = {
+    val serverSideLastActivity = super.lastActivity
+    if (serverSideState == SessionState.Running) {
+      // If the rsc client is running, we compare the lastActivity of the session and the repl,
+      // and return the more latest one
+      client.flatMap(s => Option(s.getReplLastActivity)).filter(_ > serverSideLastActivity)
```
jerryshao: nit: `client.flatMap { s => Option(s.getxxx) }`

server/src/test/scala/org/apache/livy/server/interactive/InteractiveSessionSpec.scala
```diff
@@ -247,6 +247,21 @@ class InteractiveSessionSpec extends FunSpec
       }
     }
 
+    withSession("should refresh last activity time when statement finished") { session =>
```
jerryshao: Can we use UT instead, I'm afraid adding more IT will lead to travis timeout.

rsc/src/main/java/org/apache/livy/rsc/RSCClient.java
yiheng: sure

server/src/test/scala/org/apache/livy/server/interactive/InteractiveSessionSpec.scala
yiheng: This test is running in unit test

rsc/pom.xml
```diff
@@ -49,6 +49,11 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
yiheng: We need to access the state class in rsc now.

rsc/src/main/java/org/apache/livy/rsc/RSCClient.java
```diff
@@ -411,6 +424,11 @@ private void handle(ChannelHandlerContext ctx, JobStarted msg) {
 
     private void handle(ChannelHandlerContext ctx, ReplState msg) {
       LOG.trace("Received repl state for {}", msg.state);
+      // Update last activity timestamp when state change is from busy to idle.
+      if (SessionState.Busy$.MODULE$.state().equals(replState) && msg != null &&
```
yiheng: According to [this](https://github.com/apache/incubator-livy/blob/master/repl/src/main/scala/org/apache/livy/repl/Session.scala#L263), the repl state can be changed to idle or error. However, error is a finished state. So we will check idle here.

yiheng:
@jerryshao Please help review, thanks

PR #225, opened 2019-09-04
[LIVY-657][TEST]Fix travis failed on should not create sessions with duplicate names

runzhiwang:

Fix travis failed on "should not create sessions with duplicate names"

1. When session2 was stopped, it will call stop() method in Session.scala asynchronously by Scala Future. So when assert(session2.stopped), if the stop() method has not been executed asynchronously, the test will fail.

## How was this patch tested?

Existed UT and IT.

### DIFF
--- a/server/src/test/scala/org/apache/livy/sessions/SessionManagerSpec.scala
+++ b/server/src/test/scala/org/apache/livy/sessions/SessionManagerSpec.scala
@@ -17,8 +17,6 @@
 
 package org.apache.livy.sessions
 
-import java.util.concurrent.TimeUnit
-
 import scala.concurrent.{Await, ExecutionContext, Future}
 import scala.concurrent.duration._
 import scala.language.postfixOps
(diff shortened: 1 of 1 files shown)


server/src/test/scala/org/apache/livy/sessions/SessionManagerSpec.scala
```diff
@@ -92,6 +94,7 @@ class SessionManagerSpec extends FunSpec with Matchers with LivyBaseUnitTestSuit
       an[IllegalArgumentException] should be thrownBy manager.register(session2)
       manager.get(session1.id).isDefined should be(true)
       manager.get(session2.id).isDefined should be(false)
+      Utils.waitUntil({ () => session2.stopped }, Duration(10, TimeUnit.SECONDS))
```
jerryshao: I think we can we scalatest's `Eventually` for such waiting test.

runzhiwang: @jerryshao I have updated PR.

server/src/test/scala/org/apache/livy/sessions/SessionManagerSpec.scala
```diff
@@ -27,7 +29,7 @@ import org.scalatest.{FunSpec, Matchers}
 import org.scalatest.concurrent.Eventually._
 import org.scalatest.mock.MockitoSugar.mock
 
-import org.apache.livy.{LivyBaseUnitTestSuite, LivyConf}
+import org.apache.livy.{LivyBaseUnitTestSuite, LivyConf, Utils}
```
jerryshao: This import is not used anymore, please remove it.

PR #228, opened 2019-09-09
[LIVY-663][API] exposing job id and session id in livy client

grarkydev:

Exposing job id and session id in livy client.

This change will make recognizing of job state change events' source possible.


jerryshao:
I think such APIs should also be added in Scala and Python Client. Also would you please add some unit tests, thanks.

celadari:
any updates on this ?

PR #229, opened 2019-09-11
[LIVY-644][TEST] Flaky test: Failed to execute goal org.jacoco:jacoco-maven-plugin:0.8.2:report-aggregate (jacoco-report) on project livy-coverage-report

yiheng:
This patch fixes the flaky test: Failed to execute goal org.jacoco:jacoco-maven-plugin:0.8.2:report-aggregate (jacoco-report) on project livy-coverage-report.

When JVM shutdown no gracefully in a test, the code coverage data file generated by jacoco may be corrupt. Jacoco will throw an exception when generate code coverage report.

In Livy integration test, two test cases shut down no gracefully(one of them uses System.exit). We can find random failure when jacoco process code coverage data file generated by that test case.

In this patch, we turn off the code coverage analysis on these two test cases.

## How was this patch tested?
Compare the jacoco data file generated in the integration test. Before the fix, there're 18 files, and after the fix there're 16 files, which means the fix works.

Run 10 builds on Travis each before and after the fix:
1. Before the fix: 3 builds failed due to the jacoco code coverage exception
2. After the fix: No build failed

Existing UTs and ITs.

### DIFF
--- a/rsc/src/main/java/org/apache/livy/rsc/ContextLauncher.java
+++ b/rsc/src/main/java/org/apache/livy/rsc/ContextLauncher.java
@@ -207,7 +207,7 @@ class ContextLauncher {
     if (!conf.getBoolean(CLIENT_IN_PROCESS) &&
         // For tests which doesn't shutdown RscDriver gracefully, JaCoCo exec isn't dumped properly.
         // Disable JaCoCo for this case.
-        !conf.getBoolean(TEST_STUCK_END_SESSION)) {
+        !conf.getBoolean(TEST_NO_CODE_COVERAGE_ANALYSIS)) {
       // For testing; propagate jacoco settings so that we also do coverage analysis
       // on the launched driver. We replace the name of the main file ("main.exec")
       // so that we don't end up fighting with the main test launcher.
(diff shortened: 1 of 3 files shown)


rsc/src/main/java/org/apache/livy/rsc/ContextLauncher.java
```diff
@@ -207,7 +207,7 @@ private static ChildProcess startDriver(final RSCConf conf, Promise<?> promise)
     if (!conf.getBoolean(CLIENT_IN_PROCESS) &&
         // For tests which doesn't shutdown RscDriver gracefully, JaCoCo exec isn't dumped properly.
         // Disable JaCoCo for this case.
-        !conf.getBoolean(TEST_STUCK_END_SESSION)) {
```
yiheng: The `conf.getBoolean(TEST_STUCK_END_SESSION)` actually always return false even the existing code set it to true. Because it uses `s"${RSCConf.LIVY_SPARK_PREFIX}${RSCConf.Entry.TEST_STUCK_END_SESSION.key()}"`. In the `ContextLauncher`, the livy spark prefix hasn't been removed.

In this patch, we use a new configuration key `TEST_NO_CODE_COVERAGE_ANALYSIS`. It should be used in places where the code coverage should be turned off.

jerryshao:
Try to run several times for flakiness.

jerryshao:
Thanks for the fix, I'm going to merge this PR.

PR #186, opened 2019-07-24
[LIVY-613][REPL] Livy can't handle the java.sql.Date type correctly.

397090770:
## What changes were proposed in this pull request?

When Spark table has `java.sql.Date` type column, Livy can't handle the `java.sql.Date` type correctly. e.g
```
create table test(
    name string,
    birthday date
);

insert into test values ('Livy', '2019-07-24')

curl -H "Content-Type:application/json" -X POST -d '{"code":"select * from test", "kind":"sql"}' 192.168.1.6:8998/sessions/48/statements
{"id":1,"code":"select * from test","state":"waiting","output":null,"progress":0.0}

curl 192.168.1.6:8998/sessions/48/statements/1
{"id":1,"code":"select * from test","state":"available","output":{"status":"ok","execution_count":1,"data":{"application/json":{"schema":{"type":"struct","fields":[{"name":"name","type":"string","nullable":true,"metadata":{}},{"name":"birthday","type":"date","nullable":true,"metadata":{}}]},"data":[["Livy",{}]]}}},"progress":1.0}
```
as you can see, the output of `select * from test` is `["Livy",{}]` , birthday column's value isn't handle  correctly.

The reason is that json4s can't handle `java.sql.Date`, so we should define the `CustomSerializer` for `java.sql.Date`.

This PR  add a `DateSerializer` to support `java.sql.Date` parser.

## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)
(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

Please review https://livy.incubator.apache.org/community/ before opening a pull request.

PR #234, opened 2019-09-17
[LIVY-683] Livy SQLInterpreter get empty array when extract date format rows to json

Jassy1994:

The issue link: [LIVY-683](https://issues.apache.org/jira/browse/LIVY-683)
to avoid this situation, we can transform date data to string data in rows, and this time rows can be extracted correctly.

## How was this patch tested?

manual test

### DIFF
--- a/repl/src/main/scala/org/apache/livy/repl/SQLInterpreter.scala
+++ b/repl/src/main/scala/org/apache/livy/repl/SQLInterpreter.scala
@@ -91,6 +91,9 @@ class SQLInterpreter(
             // Convert java BigDecimal type to Scala BigDecimal, because current version of
             // Json4s doesn't support java BigDecimal as a primitive type (LIVY-455).
             case i: java.math.BigDecimal => BigDecimal(i)
+            // Convert java.sql.Date type data to String, because java.sql.Date cannot extracted
+            // correctly with Json4s (LIVY-683).
+            case d: java.sql.Date => d.toString
             case e => e
           }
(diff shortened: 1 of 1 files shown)


repl/src/main/scala/org/apache/livy/repl/SQLInterpreter.scala
```diff
@@ -91,6 +91,9 @@ class SQLInterpreter(
             // Convert java BigDecimal type to Scala BigDecimal, because current version of
             // Json4s doesn't support java BigDecimal as a primitive type (LIVY-455).
             case i: java.math.BigDecimal => BigDecimal(i)
+            // Convert java.sql.Date type data to String, because java.sql.Date cannot extracted
+            // correctly with Json4s (LIVY-683).
+            case d: java.sql.Date => d.toString
```
yiheng: Can we have a unit test to cover this case?

Jassy1994: OK，I will make it

jerryshao:
This should have been fixed in #186 , what is difference here?

Jassy1994:
@jerryshao Thanks, I find it,  I think we have encountered many of the same questions, maybe we can communicate about livy usage in production.

mgaido91:
@Jassy1994 can we close this?

Jassy1994:
@mgaido91 yes, please close.

PR #235, opened 2019-09-23
[LIVY-666] Support named interpreter groups

namanmishra91:
Currently, a session contains only one interpreter group. For a new REPL of the same type, a new session has to be created. In order to support use case of multiple REPLs attached to the same spark application with separate variable scoping (similar to scoped interpreter mode in Zeppelin: https://zeppelin.apache.org/docs/0.8.0/usage/interpreter/interpreter_binding_mode.html#scoped-mode), this change adds support for having multiple interpreter groups in a session. The interpreter group on which the execution has to done can be specified via the statement execution API.

The detailed proposal is on the Jira: https://issues.apache.org/jira/browse/LIVY-666

The changes are backward compatible.

## How was this patch tested?
Tested manually on laptop by running Livy standalone. Tried running a few commands and verified the functionality.
Created a scala session:
`curl -i -X POST -H "Content-Type: application/json" -d '{"name":"test", "kind":"scala"}' localhost:8998/sessions`

`curl localhost:8998/sessions/0/statements -X POST -H 'Content-Type: application/json' -d '{"code":"val a=5", "interpreterGroup":"g1"}'`
`curl localhost:8998/sessions/0/statements -X POST -H 'Content-Type: application/json' -d '{"code":"print(a)", "interpreterGroup":"g1"}'`
prints 5
`curl localhost:8998/sessions/0/statements -X POST -H 'Content-Type: application/json' -d '{"code":"print(a)", "interpreterGroup":"g2"}'`
'a' is not defined

I have added a few unit tests but couldn't run those since other tests are failing. Verifying that the tests work as expected is in progress.

### DIFF
--- a/repl/scala-2.11/src/main/scala/org/apache/livy/repl/SparkInterpreter.scala
+++ b/repl/scala-2.11/src/main/scala/org/apache/livy/repl/SparkInterpreter.scala
@@ -21,6 +21,7 @@ import java.io.File
 import java.net.URLClassLoader
 import java.nio.file.{Files, Paths}
 
+import scala.collection.JavaConversions._
 import scala.tools.nsc.Settings
 import scala.tools.nsc.interpreter.Completion.ScalaCompleter
 import scala.tools.nsc.interpreter.IMain
@@ -45,16 +46,22 @@ class SparkInterpreter(protected override val conf: SparkConf) extends AbstractS
     require(sparkILoop == null)
 
     val rootDir = conf.get("spark.repl.classdir", System.getProperty("java.io.tmpdir"))
-    val outputDir = Files.createTempDirectory(Paths.get(rootDir), "spark").toFile
-    outputDir.deleteOnExit()
-    conf.set("spark.repl.class.outputDir", outputDir.getAbsolutePath)
+    SparkInterpreter.sharedInterpreterLock.synchronized {
+      if (SparkInterpreter.outputDir == null) {
+        SparkInterpreter.outputDir = Files.createTempDirectory(Paths.get(rootDir), "spark").toFile
+        SparkInterpreter.outputDir.deleteOnExit()
+      }
+    }
 
+    val outputDir = SparkInterpreter.outputDir
     val settings = new Settings()
+    conf.set("spark.repl.class.outputDir", outputDir.getAbsolutePath)
     settings.processArguments(List("-Yrepl-class-based",
       "-Yrepl-outdir", s"${outputDir.getAbsolutePath}"), true)
     settings.usejavacp.value = true
     settings.embeddedDefaults(Thread.currentThread().getContextClassLoader())
 
+    System.setProperty("scala.repl.name.line", "$line"+this.hashCode().toString());
     sparkILoop = new SparkILoop(None, new JPrintWriter(outputStream, true))
     sparkILoop.settings = settings
     sparkILoop.createInterpreter()
@@ -132,4 +139,9 @@ class SparkInterpreter(protected override val conf: SparkConf) extends AbstractS
       sparkILoop.bind(name, tpe, value, modifier)
     }
   }
+}
+
+object SparkInterpreter {
+  var outputDir: File = null
+  val sharedInterpreterLock: Int = 0
 }
--- a/repl/scala-2.11/src/main/scala/org/apache/livy/repl/SparkInterpreter.scala
+++ b/repl/scala-2.11/src/main/scala/org/apache/livy/repl/SparkInterpreter.scala
@@ -21,7 +21,6 @@ import java.io.File
 import java.net.URLClassLoader
 import java.nio.file.{Files, Paths}
 
-import scala.collection.JavaConversions._
 import scala.tools.nsc.Settings
 import scala.tools.nsc.interpreter.Completion.ScalaCompleter
 import scala.tools.nsc.interpreter.IMain
@@ -61,7 +60,7 @@ class SparkInterpreter(protected override val conf: SparkConf) extends AbstractS
     settings.usejavacp.value = true
     settings.embeddedDefaults(Thread.currentThread().getContextClassLoader())
 
-    System.setProperty("scala.repl.name.line", "$line"+this.hashCode().toString());
+    System.setProperty("scala.repl.name.line", "$line" + this.hashCode().toString());
     sparkILoop = new SparkILoop(None, new JPrintWriter(outputStream, true))
     sparkILoop.settings = settings
     sparkILoop.createInterpreter()
--- a/repl/src/main/scala/org/apache/livy/repl/Session.scala
+++ b/repl/src/main/scala/org/apache/livy/repl/Session.scala
@@ -57,6 +57,8 @@ class Session(
   extends Logging {
   import Session._
 
+  private val DEFAULT_INTERPRETER_GROUP_KEY: String = "default"
+
   private val interpreterExecutor = ExecutionContext.fromExecutorService(
     Executors.newSingleThreadExecutor())
 
@@ -81,6 +83,8 @@ class Session(
   private val defaultInterpKind = Kind(livyConf.get(RSCConf.Entry.SESSION_KIND))
 
   private val interpGroup = new mutable.HashMap[Kind, Interpreter]()
+  // Map of interpreterGroupKey -> InterpreterGroup
+  private val sessionsMap = new mutable.HashMap[String, mutable.HashMap[Kind, Interpreter]]()
 
   @volatile private var entries: SparkEntries = _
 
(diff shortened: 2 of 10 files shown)


namanmishra91:
@jerryshao Please help reviewing this

jerryshao:
@namanmishra91 thanks for your work, did you finish all your works?

@yiheng @runzhiwang would you please help to review, thanks.

jerryshao:
Also have you tried multiple scala/python/R interpreters in one context?

namanmishra91:
Basic backend support and API modifications have been done as part of this PR. It will be good to modify/add few new APIs for better debuggability etc. as described in the proposal that are not included yet.

 Should I add those as well in this PR itself? I thought that will make this PR huge and will be difficult to review. Hence I think its better to have a separate PR for enhancements on top of this.


> @namanmishra91 thanks for your work, did you finish all your works?

namanmishra91:
> Also have you tested multiple scala/python/R interpreters in one context?

As shown in [proposal](https://issues.apache.org/jira/secure/attachment/12980485/12980485_multiple_interpreter_groups.png), this implementation supports only one interpreter for each type per context. Is there a use-case to have multiple interpreters of same type under one context?

repl/scala-2.11/src/main/scala/org/apache/livy/repl/SparkInterpreter.scala
```diff
@@ -45,16 +45,22 @@ class SparkInterpreter(protected override val conf: SparkConf) extends AbstractS
     require(sparkILoop == null)
 
     val rootDir = conf.get("spark.repl.classdir", System.getProperty("java.io.tmpdir"))
-    val outputDir = Files.createTempDirectory(Paths.get(rootDir), "spark").toFile
-    outputDir.deleteOnExit()
-    conf.set("spark.repl.class.outputDir", outputDir.getAbsolutePath)
+    SparkInterpreter.sharedInterpreterLock.synchronized {
```
namanmishra91: SparkInterpreter changes fix an issue discovered during testing this change . Fix is along the lines for https://issues.apache.org/jira/browse/ZEPPELIN-1453

namanmishra91:
The travis errors seem unrelated. I see "java.lang.OutOfMemoryError: GC overhead limit exceeded" in all the failed travis jobs. Is there a configuration that needs to be set?

runzhiwang:
@namanmishra91 hi, the travis error has not seen before, it maybe caused by this pr.

namanmishra91:
@runzhiwang Sure, looking into it.

Tagar:
Would this need some changes on Zeppelin too?

cc @zjffdu 

Would this feature be useful outside of Zeppelin? (if yes, then perhaps "interpreter group" should be called as something more generic)

Thank you

repl/scala-2.11/src/main/scala/org/apache/livy/repl/SparkInterpreter.scala
```diff
@@ -133,3 +139,8 @@ class SparkInterpreter(protected override val conf: SparkConf) extends AbstractS
     }
   }
 }
+
+object SparkInterpreter {
+  var outputDir: File = null
```
zjffdu: Why outputDir is shared across InterpreterGroup ? Wouldn't it cause issues between multiple scala repl ?

repl/src/main/scala/org/apache/livy/repl/Session.scala
```diff
@@ -81,6 +83,8 @@ class Session(
   private val defaultInterpKind = Kind(livyConf.get(RSCConf.Entry.SESSION_KIND))
 
   private val interpGroup = new mutable.HashMap[Kind, Interpreter]()
```
zjffdu: Is this `interpGroup` still needed ?

repl/src/main/scala/org/apache/livy/repl/Session.scala
```diff
@@ -161,12 +179,12 @@ class Session(
     _statements.synchronized { _statements(statementId) = statement }
 
     Future {
-      setJobGroup(tpe, statementId)
+      setJobGroup(tpe, statementId, interpreterGroup)
```
zjffdu: How about using `interpreterGroupId` instead of `interpreterGroup` ?

zjffdu:
Thanks @Tagar for ping me. I think this feature would be useful for Zeppelin. And you are right, the concept in livy here is different from Zeppelin. In Zeppelin one InterpreterGroup includes multiple sessions, while in Livy here one Session includes multiple InterpreterGroup. I am afraid it is difficult for 2 projects align concepts, but it may be fine as long as each project contributor what is the underlying difference. If anyone has another proper name to replace `interpreter group`, I would love to hear.

repl/src/main/scala/org/apache/livy/repl/Session.scala
```diff
@@ -91,29 +95,42 @@ class Session(
...
-          case SparkR => SparkRInterpreter(sparkConf, entries)
-          case SQL => new SQLInterpreter(sparkConf, livyConf, entries)
+      interpreterGroupKey
+    }
+    var intpGroupOpt = sessionsMap.get(key)
+    if (intpGroupOpt == None) {
+      sessionsMap.put(key, new mutable.HashMap[Kind, Interpreter])
+    }
+
+    val intpGroup = sessionsMap.get(key).get
```
yiheng: From 106 to 111, what about?
```scala
val intpGroup = sessionsMap.getOrElseUpdate(key, new mutable.HashMap[Kind, Interpreter])
```

repl/src/main/scala/org/apache/livy/repl/Session.scala
```diff
@@ -91,29 +95,42 @@ class Session(
     entries.sc().sc
   }
 
-  private[repl] def interpreter(kind: Kind): Option[Interpreter] = interpGroup.synchronized {
-    if (interpGroup.contains(kind)) {
-      Some(interpGroup(kind))
+  private[repl] def interpreter(kind: Kind, interpreterGroupKey: String =
```
yiheng: can we break the line after the first parameter?

repl/src/main/scala/org/apache/livy/repl/Session.scala
```diff
@@ -147,7 +164,8 @@ class Session(
     _statements.toMap
   }
 
-  def execute(code: String, codeType: String = null): Int = {
+  def execute(code: String, codeType: String = null, interpreterGroup: String =
```
yiheng: can we break the line after the first parameter?

repl/src/main/scala/org/apache/livy/repl/Session.scala
```diff
@@ -91,29 +95,42 @@ class Session(
...
+    val intpGroup = sessionsMap.get(key).get
+    intpGroup.synchronized {
+      if (intpGroup.contains(kind)) {
+        Some(intpGroup(kind))
+      } else {
+        try {
+          require(entries != null,
+            "SparkEntries should not be null when lazily initialize other interpreters.")
+
+          val interp = kind match {
```
yiheng: So for each interpreter group, for each kind, we can have at most one interpreter? I think this restriction is a little surprise...

repl/src/test/scala/org/apache/livy/repl/ReplDriverSuite.scala
```diff
@@ -53,7 +53,7 @@ class ReplDriverSuite extends FunSuite with LivyBaseUnitTestSuite {
       // This is sort of what InteractiveSession.scala does to detect an idle session.
       client.submit(new PingJob()).get(60, TimeUnit.SECONDS)
 
-      val statementId = client.submitReplCode("1 + 1", "spark").get
+      val statementId = client.submitReplCode("1 + 1", "spark", null).get
```
yiheng: can we overload the submitReplCode with a two-parameter version? So we don't need to change the old code.

yiheng:
Do we need to change the Session recovery process?

repl/scala-2.11/src/main/scala/org/apache/livy/repl/SparkInterpreter.scala
namanmishra91: OutputDir needs to be at the application level, else any code that uses case classes in scala was throwing ClassCastExceptions from Spark. On searching a bit, found that similar issue was found and fixed in Zeppelin. Check https://issues.apache.org/jira/browse/ZEPPELIN-1453

repl/src/main/scala/org/apache/livy/repl/Session.scala
namanmishra91: Thanks for pointing. It is not needed now. Will also replace its occurrences with `sessionsMap` equivalent

repl/scala-2.11/src/main/scala/org/apache/livy/repl/SparkInterpreter.scala
zjffdu: This ticket is 2 years ago, I don't see this in the current zeppelin code base. Have you verified it in multiple interpreter group mode ?

repl/src/main/scala/org/apache/livy/repl/Session.scala
namanmishra91: Thanks, will change.

repl/src/main/scala/org/apache/livy/repl/Session.scala
namanmishra91: @yiheng The intent is to share a session and have multiple namespaces/contexts. Since the namespace is defined at the interpreter level, one can always use another interpreter group for having a different context. Is there any specific use case in which we would want multiple interpreters of a `Kind` within an interpreter group?

repl/src/main/scala/org/apache/livy/repl/Session.scala
namanmishra91: Actually `interpreterGroup` is the id here. Will modify the variable name.

repl/src/test/scala/org/apache/livy/repl/ReplDriverSuite.scala
namanmishra91: Sure

repl/src/main/scala/org/apache/livy/repl/Session.scala
yiheng: Please correct me if I'm wrong. If a user creates a python interpreter and another user want to reuse the session but with different variable scop, the second user cannot use python interpreter. Is it reasonable?

repl/scala-2.11/src/main/scala/org/apache/livy/repl/SparkInterpreter.scala
namanmishra91: Yes, I had verified it with multiple interpreter groups. I am not sure how its working with separate outputDirs currently in Zeppelin but that was there until recently. I checked and I still see this line with the comment stating that the compiled class names need to be different: https://github.com/apache/zeppelin/blob/master/spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala#L91

Let me verify once again.

repl/src/main/scala/org/apache/livy/repl/Session.scala
namanmishra91: Not sure if I understand it correctly but in the above scenario, the second user can create another interpreter group that will have a totally different set of scala, python, r and sql interpreters. So all interpreters can be used by him in new context. 

The limit of one interpreter is per interpreter group and not per session. Please refer (https://issues.apache.org/jira/secure/attachment/12980485/12980485_multiple_interpreter_groups.png). Let me know if this answers your question.

yiheng: Yes. I misunderstand. Thanks for the clarification.

PR #237, opened 2019-09-25
[LIVY-688] Error message of BypassJobStatus should contains cause information of Exception

WeiWenda:
Change the implement of org.apache.livy.rsc.Utils.stackTraceAsString to guava Throwables.getStackTraceAsString, so that user can receive details of error message by calling org.apache.livy.client.http.JobHandleImpl.get.

https://issues.apache.org/jira/browse/LIVY-688

### DIFF
--- a/rsc/src/main/java/org/apache/livy/rsc/Utils.java
+++ b/rsc/src/main/java/org/apache/livy/rsc/Utils.java
@@ -20,6 +20,7 @@ package org.apache.livy.rsc;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import com.google.common.base.Throwables;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.GenericFutureListener;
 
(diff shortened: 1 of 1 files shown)


rsc/src/main/java/org/apache/livy/rsc/Utils.java
```diff
@@ -91,13 +92,7 @@ public static String join(Iterable<String> strs, String sep) {
...
 
   public static String stackTraceAsString(Throwable t) {
-    StringBuilder sb = new StringBuilder();
-    sb.append(t.getClass().getName()).append(": ").append(t.getMessage());
-    for (StackTraceElement e : t.getStackTrace()) {
-      sb.append("\n");
-      sb.append(e.toString());
-    }
-    return sb.toString();
+    return Throwables.getStackTraceAsString(t);
```
jerryshao: Would you please change to not use Guava? We removed the Guava support in Livy server side, and if this method will be used in server side in future, this will potentially cause some issues. CC @runzhiwang .

WeiWenda: no problem. Done!

PR #238, opened 2019-09-25
[LIVY-689] Deliver stage process message to the end user using thriftserver

boneanxs:
When we use beeline to connect with thriftserver, we must open the spark history server to get the stage process, which is unconvenient, we can deliver process message to the end user.

## How was this patch tested?
In our environment, the log to the end user is below,

```log
0: jdbc:hive2://localhost:10090> select count(*) from dp_dw.tmp;
RSC client is executing SQL query: select count(*) from dp_dw.tmp, statementId = 8de21e99-562d-46d9-a9f7-3007f82d5828, session = SessionHandle [89eb625f-ffdc-4dce-b23c-c715f4d6444d]
2019-11-19 20:34:10.963 percentage:   0% active: 0 completed: 0 failed: 0 all: 0
2019-11-19 20:34:12.965 percentage:   0% active: 0 completed: 0 failed: 0 all: 294
2019-11-19 20:34:14.951 percentage:   0% active: 0 completed: 0 failed: 0 all: 294
2019-11-19 20:34:16.95  percentage:   0% active: 0 completed: 0 failed: 0 all: 294
2019-11-19 20:34:18.951 percentage:   0% active: 0 completed: 0 failed: 0 all: 294
2019-11-19 20:34:20.951 percentage:   0% active: 1 completed: 0 failed: 0 all: 294
2019-11-19 20:34:22.958 percentage:   0% active: 2 completed: 1 failed: 0 all: 294
2019-11-19 20:34:24.952 percentage:  17% active: 2 completed: 51 failed: 0 all: 294
2019-11-19 20:34:26.95  percentage:  42% active: 2 completed: 124 failed: 0 all: 294
2019-11-19 20:34:28.951 percentage:  60% active: 2 completed: 177 failed: 0 all: 294
2019-11-19 20:34:30.951 percentage:  61% active: 6 completed: 180 failed: 0 all: 294
2019-11-19 20:34:32.951 percentage:  66% active: 13 completed: 194 failed: 0 all: 294
2019-11-19 20:34:34.95  percentage:  99% active: 6 completed: 290 failed: 0 all: 294
2019-11-19 20:34:36.952 percentage:  99% active: 2 completed: 292 failed: 0 all: 294
2019-11-19 20:34:38.952 percentage:  99% active: 2 completed: 292 failed: 0 all: 294
2019-11-19 20:34:40.952 percentage:  99% active: 2 completed: 292 failed: 0 all: 294
2019-11-19 20:34:42.951 percentage:  99% active: 2 completed: 292 failed: 0 all: 294
2019-11-19 20:34:44.953 percentage:  99% active: 2 completed: 292 failed: 0 all: 294
2019-11-19 20:34:46.957 percentage:  99% active: 2 completed: 292 failed: 0 all: 294
2019-11-19 20:34:48.953 percentage:  99% active: 2 completed: 292 failed: 0 all: 294
2019-11-19 20:34:50.96  percentage:  99% active: 2 completed: 292 failed: 0 all: 294
2019-11-19 20:34:52.954 percentage:  99% active: 2 completed: 292 failed: 0 all: 294
2019-11-19 20:34:54.953 percentage:  99% active: 2 completed: 292 failed: 0 all: 294
2019-11-19 20:34:56.953 percentage:  99% active: 2 completed: 292 failed: 0 all: 294
2019-11-19 20:34:58.953 percentage:  99% active: 2 completed: 292 failed: 0 all: 294
2019-11-19 20:35:00.954 percentage:  99% active: 2 completed: 292 failed: 0 all: 294
2019-11-19 20:35:02.953 percentage:  99% active: 2 completed: 292 failed: 0 all: 294
2019-11-19 20:35:04.954 percentage:  99% active: 2 completed: 292 failed: 0 all: 294
2019-11-19 20:35:06.953 percentage:  99% active: 2 completed: 292 failed: 0 all: 294
2019-11-19 20:35:08.954 percentage: 100% active: 1 completed: 293 failed: 0 all: 294
2019-11-19 20:35:10.954 percentage: 100% active: 1 completed: 293 failed: 0 all: 294
RSC client is fetching result schema for statementId = 8de21e99-562d-46d9-a9f7-3007f82d5828
+------------+
|  count(1)  |
+------------+
| 182138357  |
+------------+
1 row selected (61.118 seconds)

### CODE REVIEW

**Comment by Tagar** [comment]
Is there is a way to disable this output? Something like `spark.ui.showConsoleProgress`=true/false

Also, would be nice if this could use some terminal capabilities to overwrite progress output using single terminal line, instead of writing each progress output on a separate line 

```
[Stage 0:>                                                        (0 + 25) / 25]
[Stage 0:==>                                                      (1 + 24) / 25]
[Stage 0:====>                                                    (2 + 23) / 25]
[Stage 0:=========>                                               (4 + 21) / 25]
[Stage 0:==================>                                      (8 + 17) / 25]
[Stage 0:========================>                               (11 + 14) / 25]
[Stage 0:========================>                               (11 + 14) / 25]
[Stage 0:=============================>                          (13 + 12) / 25]
[Stage 0:======================================>                  (17 + 8) / 25]
[Stage 0:==================================================>      (22 + 3) / 25]
```

**Comment by Tagar** [comment]
Beeline from Hive 2.2 has in-place progress bar `hive.server2.in.place.progress`

https://cwiki.apache.org/confluence/display/Hive/Configuration+Properties#ConfigurationProperties-hive.server2.in.place.progress

(I think it works only with tez - would be nice if this would work with Livy too )

Thanks.

File: rsc/src/main/java/org/apache/livy/rsc/RSCClient.java [hunk 1]
```diff
@@ -330,6 +335,9 @@ public long getReplLastActivity() {
 
   private class ClientProtocol extends BaseProtocol {
 
+    private  final int TerminalWidth =
```
yiheng: remove a space after private. And `TERMINAL_WIDTH`?

rsc/src/main/java/org/apache/livy/rsc/RSCClient.java
```diff
@@ -381,6 +389,11 @@ String bypass(ByteBuffer serializedJob, String jobType, boolean sync) {
       return jobId;
     }
 
+    ConcurrentBoundedLinkedQueue<String> operationMessage;
```
yiheng: private?

rsc/src/main/java/org/apache/livy/rsc/RSCClient.java
```diff
@@ -413,6 +426,28 @@ private void handle(ChannelHandlerContext ctx, JobResult msg) {
       }
     }
 
+    private void handle(ChannelHandlerContext ctx, JobProcessMessage msg){
+      if (operationMessage != null){
```
yiheng: space between ) and {

rsc/src/main/java/org/apache/livy/rsc/RSCClient.java
```diff
@@ -413,6 +426,28 @@ private void handle(ChannelHandlerContext ctx, JobResult msg) {
...
+        String tail = "(" + msg.completed + " + " + msg.actived + ") / " + msg.all + "]";
+        int processWid = TerminalWidth - head.length() - tail.length();
+        int barWid = processWid * msg.completed / msg.all;
+        bar.append(head);
+        for (int i = 0; i < processWid; i++) {
+          if (i < barWid){
+            bar.append("=");
+          }else if (i == barWid){
+            bar.append(">");
+          }else {
```
yiheng: space

rsc/src/main/java/org/apache/livy/rsc/RSCClient.java
```diff
@@ -413,6 +426,28 @@ private void handle(ChannelHandlerContext ctx, JobResult msg) {
...
+        StringBuilder bar = new StringBuilder();
+        String head = "[Stage " + msg.stageId + ":";
+        String tail = "(" + msg.completed + " + " + msg.actived + ") / " + msg.all + "]";
+        int processWid = TerminalWidth - head.length() - tail.length();
+        int barWid = processWid * msg.completed / msg.all;
+        bar.append(head);
+        for (int i = 0; i < processWid; i++) {
+          if (i < barWid){
+            bar.append("=");
+          }else if (i == barWid){
```
yiheng: space

rsc/src/main/java/org/apache/livy/rsc/RSCClient.java
```diff
@@ -413,6 +426,28 @@ private void handle(ChannelHandlerContext ctx, JobResult msg) {
       }
     }
 
+    private void handle(ChannelHandlerContext ctx, JobProcessMessage msg){
+      if (operationMessage != null){
+        StringBuilder bar = new StringBuilder();
+        String head = "[Stage " + msg.stageId + ":";
+        String tail = "(" + msg.completed + " + " + msg.actived + ") / " + msg.all + "]";
+        int processWid = TerminalWidth - head.length() - tail.length();
+        int barWid = processWid * msg.completed / msg.all;
```
yiheng: what about `barWidth`?

rsc/src/main/java/org/apache/livy/rsc/RSCClient.java
```diff
@@ -413,6 +426,28 @@ private void handle(ChannelHandlerContext ctx, JobResult msg) {
       }
     }
 
+    private void handle(ChannelHandlerContext ctx, JobProcessMessage msg){
```
yiheng: In our scenario, we have a different format requirement. Can we wrap the format logic to make it easy to change?

rsc/src/main/java/org/apache/livy/rsc/RSCConf.java
```diff
@@ -78,6 +78,8 @@
     JOB_CANCEL_TRIGGER_INTERVAL("job-cancel.trigger-interval", "100ms"),
     JOB_CANCEL_TIMEOUT("job-cancel.timeout", "30s"),
 
+    JOB_PROCESS_MSG_UPDATE_INTERVAL("job.process.msg.update.interval", 2000L),
```
yiheng: what's the time unit? Can we make the style consistent with others like the above `JOB_CANCEL_TRIGGER_INTERVAL`?

huianyi:
@Tagar thanks for your reply. You can set `livy.server.thrift.logging.operation.enabled = false` to disable it. For question 2, it might need rewrite getProgressedPercentage method in ThriftClIService and add new class implemented ProgressMonitor.

huianyi:
Currently I use operationMessage in  to achieve this feature, but maybe take out  operationMessage in LivyThriftSeesionManager as a separate class is a good way?

rsc/src/main/java/org/apache/livy/rsc/driver/JobWrapper.java
```diff
@@ -42,10 +45,17 @@
 
   private Future<?> future;
 
+  private long firstDelayMSec = 500L;
+  private long updatePeriodMSec;
```
yiheng: final?

rsc/src/main/java/org/apache/livy/rsc/driver/JobWrapper.java
```diff
@@ -42,10 +45,17 @@
 
   private Future<?> future;
 
+  private long firstDelayMSec = 500L;
```
yiheng: final?

rsc/src/main/java/org/apache/livy/rsc/driver/RSCDriver.java
```diff
@@ -212,6 +215,33 @@ public void onSaslComplete(Rpc client) {
     setupIdleTimeout();
   }
 
+  /**
+   * calculate each stage's process and broadcast message to the end user
```
yiheng: Calculate ?

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
```diff
@@ -196,7 +196,7 @@ class LivyThriftSessionManager(val server: LivyThriftServer, val livyConf: LivyC
       sessionHandle: SessionHandle,
       livySession: InteractiveSession,
       initStatements: List[String]): Unit = {
-    val rpcClient = new RpcClient(livySession)
+    val rpcClient = new RpcClient(livySession, None)
```
yiheng: Maybe we can write an auxiliary constructor for RpcClient then there's no change here.

rsc/src/main/java/org/apache/livy/rsc/driver/RSCDriver.java
```diff
@@ -212,6 +215,33 @@ public void onSaslComplete(Rpc client) {
...
+   */
+  void handleProcessMessage(String jobId){
+    SparkStatusTracker sparkStatusTracker;
+    synchronized (jcLock){
+      if (jc == null){
+        return;
+      }
+      sparkStatusTracker = jc.sc().sc().statusTracker();
+    }
+    int[] activeStageIds = sparkStatusTracker.getActiveStageIds();
```
yiheng: Is it possible these activeStageIds belong to different jobs?

huianyi: We can only submit one job using beeline at a time, so this may not happened?

huianyi:
@yiheng Thanks for your review, Could you also help me check why the test failed? I did not create another spark context in the patch. 
I'm gonna fix these problems in the next commit.

yiheng:
> @yiheng Thanks for your review, Could you also help me check why the test failed? I did not create another spark context in the patch.
> I'm gonna fix these problems in the next commit.

I'm not sure why. I guess the create another spark context is not the root cause. It should be caused by another reason. Can you reproduce the error in your local?

rsc/src/main/java/org/apache/livy/rsc/RSCClient.java
huianyi: sorry, I can not catch your point, what does `format requirement` mean. Could you please explain this more specificly?

yiheng: currently, the output of the log is like this
```
[Stage 0:==================>                                      (8 + 17) / 25]
```

We hope it can be customized. For example
```
Stage:0 Finished:8 Total:25
```

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
huianyi: Do you think it might be better if we can seprate object `operationMessage` as a class out of `LivyExecuteStatementOperation`? Then we don not need deliver it to others.

captainzmc:
Hi @huianyi ，I have saw the ERROR in your uts. The root cause of your UT failure is:
Can't find  class ConcurrentBoundedLinkedQueue in thriftserver .

You can avoid this error by adding the following code in thriftserver/session/pom.xml：
<dependency>
  <groupId>org.apache.livy</groupId>
  <artifactId>livy-core_${scala.binary.version}</artifactId>
  <version>${project.version}</version>
  <scope>provided</scope>
</dependency>

captainzmc:
Hi @huianyi ，I have saw the ERROR in your uts. The root cause of your UT failure is:
Can't find class ConcurrentBoundedLinkedQueue in thriftserver .

You can avoid this error by adding the following code in thriftserver/session/pom.xml：
`<dependency>
  <groupId>org.apache.livy</groupId>
  <artifactId>livy-core_${scala.binary.version}</artifactId>
  <version>${project.version}</version>
  <scope>provided</scope>
</dependency>`

rsc/src/main/java/org/apache/livy/rsc/RSCClient.java
captainzmc: It feels better to use text descriptions. It is also easier to parse if we need to collect logs:

Stage:0 Finished:8 Total:25; Running job progress: 32%

huianyi:
> Hi @huianyi ，I have saw the ERROR in your uts. The root cause of your UT failure is:
> Can't find class ConcurrentBoundedLinkedQueue in thriftserver .
> 
> You can avoid this error by adding the following code in thriftserver/session/pom.xml：
> `<dependency> <groupId>org.apache.livy</groupId> <artifactId>livy-core_${scala.binary.version}</artifactId> <version>${project.version}</version> <scope>provided</scope> </dependency>`
Thanks for your very kind support, gonna push a new commit.

rsc/src/main/java/org/apache/livy/rsc/RSCClient.java
huianyi: gonna change this format soon.

huianyi:
@yiheng Could you please review the new commit again? All checks have passed now.

core/src/main/scala/org/apache/livy/OperationMessageManager.scala
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  def get(): ConcurrentBoundedLinkedQueue[String] = {
+    threadLocalLogQueue.get()
+  }
+
+  def set(threadLocalLogQueue: ConcurrentBoundedLinkedQueue[String]){
+    this.threadLocalLogQueue.set(threadLocalLogQueue)
+  }
+  private val threadLocalLogQueue = new ThreadLocal[ConcurrentBoundedLinkedQueue[String]]
```
yiheng: I think it's odd to use a ThreadLocal here...

I'm afraid it may be not correct in some scenarios. For example, the threads run on a thread pool. The thread is reused and the ThreadLocal variable will return ConcurrentBoundedLinkedQueue of another rpc client, right?

I think the old way should be good. What's the reason we change it like this?

yiheng:
> @yiheng Could you please review the new commit again? All checks have passed now.

I only have one comment. Others look good to me.

core/src/main/scala/org/apache/livy/OperationMessageManager.scala
huianyi: Using this threadLocal variable gives us an easy way to get operationMessage object in other classes, we do not need transfer it in other classes. For example, Class ClientProtocol need operationMessage to deliver process message to the end user, we can just call `OperationMessageManager.get()` to get it.

```java
private final ConcurrentBoundedLinkedQueue<String> operationMessages =
        OperationMessageManager.get();
```
I don't think this might be reused of another rpc client, ConcurrentBoundedLinkedQueue would be initialized in `LivyThriftSessionManager.openSession` method, for every new session, we will create only one RSCClient for it, and both of them are in the same thread, so for every RSCClient, they will have the right ConcurrentBoundedLinkedQueue. 

Please correct me if I miss something.

yiheng: In the current implementation, it should be safe.

But we bring in an implicit constraint on how to use the `OperationMessageManager.get()` and `OperationMessageManager.set()` methods, that you must make sure they're called in the same thread. How can we guarantee developer will not break it in the future?

IMHO, compared to the old code, it's not straightforward to understand and maintain.

huianyi: I think this gives us another more convenient way to get operationMessage object in other classes if they are in the same thread(just like `Sessionstate.get()` method in hive), for those classes are not in the same thread, we need deliver this object to them, like Class `LivyExecuteStatementOperation`.

Actually, if we do not use this threadLocal variable, for class `ClientProtocol` needing operationMessage to deliver process message, we should first deliver operationMessage to RSCClient, then RSCClient deliver it to ClientProtocol, which needs lots of set and get method.

Would it be aIright if I add detailed comment in `OperationMessageManager` to call developers' attention to the constraint?

jerryshao: Yes, it is strange to use threadlocal variable here? I would suggest to a more decent way even if we need more code changes.

rsc/src/main/java/org/apache/livy/rsc/BaseProtocol.java
```diff
@@ -126,6 +126,28 @@ public JobResult() {
...
   }
 
+  protected static class JobProcessMessage {
+    public final String id;
+    public final int stageId;
+    public final int completed;
+    public final int all;
+    public final int active;
+    public final int failed;
+    public JobProcessMessage(String id, int stageId, int completed, int active,
```
jerryshao: Please change the style to one parameter per line.

rsc/src/main/java/org/apache/livy/rsc/driver/JobWrapper.java
```diff
@@ -60,6 +70,14 @@ public Void call() throws Exception {
         }
       }
 
+        timer.schedule(new TimerTask() {
+            @Override
+            public void run() {
+                driver.handleProcessMessage(jobId);
+            }
+        }, firstDelayMSec, updatePeriodMSec);
```
jerryshao: Two space indent here.

jerryshao:
> Beeline from Hive 2.2 has in-place progress bar hive.server2.in.place.progress
>
>https://cwiki.apache.org/confluence/display/Hive/Configuration+Properties#ConfigurationProperties-hive.server2.in.place.progress
>
>(I think it works only with tez - would be nice if this would work with Livy too )

Would be better to support this if it is not a big effort.

core/src/main/scala/org/apache/livy/ConcurrentBoundedLinkedQueue.scala
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.livy
+
+import java.util.concurrent.ConcurrentLinkedQueue
+
+class ConcurrentBoundedLinkedQueue[E](capacity: Long) extends ConcurrentLinkedQueue[E]{
```
jerryshao: Space before `{`. Please fix here and various other places.

core/src/main/scala/org/apache/livy/ConcurrentBoundedLinkedQueue.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  override def offer(e: E): Boolean = {
+    if (this.size() ==  capacity){
```
jerryshao: I guess there has two spaces before `capacity`

rsc/src/main/java/org/apache/livy/rsc/driver/RSCDriver.java
jerryshao: I also have the same concern. Beeline can only submit one query a time, but what if user submits query programmatically? We can hardly say that all the active stages belong to one query, so there should be a way to figure out job id -> active stages mapping.

rsc/src/main/java/org/apache/livy/rsc/BaseProtocol.java
huianyi: Looks like all java code writes all parameter one line, should I change this?

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
```diff
@@ -252,6 +259,8 @@ class LivyThriftSessionManager(val server: LivyThriftServer, val livyConf: LivyC
               }
             }
             initSession(sessionHandle, livySession, initStatements)
+            operationMessages.foreach(
+              _.offer(s"tracking URL: ${livySession.appInfo.sparkUiUrl.orNull}"))
```
captainzmc: Is this code sure to get the tracking URL? I get null even though my livy is on yarn.

huianyi: I can get the application URL in my environment, which should be set after `initSession` method, could you please show me how your produce this issue? May `initSession` method not run right?

rsc/src/main/java/org/apache/livy/rsc/driver/RSCDriver.java
jerryshao: NVM, just remembered that Livy cannot submit one job a time. But there's another issue, some queries may split into multiple jobs, then the measurement here may not be accurate.

huianyi:
> > Beeline from Hive 2.2 has in-place progress bar hive.server2.in.place.progress
> > https://cwiki.apache.org/confluence/display/Hive/Configuration+Properties#ConfigurationProperties-hive.server2.in.place.progress
> > (I think it works only with tez - would be nice if this would work with Livy too )
> 
> Would be better to support this if it is not a big effort.

After reading hive related code, this needs write some new classes to implement this feature, which might cause this PR large, if you don not mind, I'd love to open a new PR to support this.

jerryshao:
Sure, I'm fine with it.

rsc/src/main/java/org/apache/livy/rsc/driver/RSCDriver.java
huianyi: Yes, I find another way by using jobGroupId(which is livy jobId) may fix this problem, gonna test it and pull a new commit if it would work.

huianyi: > NVM, just remembered that Livy cannot submit one job a time. But there's another issue, some queries may split into multiple jobs, then the measurement here may not be accurate.

If Livy cannot submit one job a time, stage id in one query is unique, I already log stage process message with stage id, so this should not be a problem?

huianyi: >  just remembered that Livy cannot submit one job a time

Do you mean one session should only run a livy job a time? if this is true, current implementation should not be a problem?

jerryshao: Just rethink about the current solution, I think there should be no issue about the current way. Sorry about it.

huianyi: : ), ok, I'm gonna fix other problems soon, very thanks for your review.

rsc/src/main/java/org/apache/livy/rsc/driver/JobWrapper.java
```diff
@@ -60,6 +70,14 @@ public Void call() throws Exception {
         }
       }
 
+      timer.schedule(new TimerTask() {
```
jerryshao: It is super cost to start a timer per job.

huianyi:
> Rethink about the way this ticket doing, I think current way may not be the best way.
> 
> 1. `jobId` may not exactly match the currently running jobs, there's no issue for beeline, but could be an issue for multi-thread usage. Currently when getting active jobs, there's no direct relation to jobId.
> 2. There's already a way to calculate job progress in ReplDriver for statement, we should extend the current way to also apply to job API.
> 3. Using LivyServer to store the session/job status may not be fault tolerant. It would better to extend the job API to get job progress synchronously.

Thanks for your reply, if I have understand your meaning correctly, actually I can use `jobId`(which is spark jobs' group id delivered by livy) in `handleProcessMessage` method of RSCDriver, then get related stages using this `jobId`, to handle these circumstances above. May this be a good way?

rsc/src/main/java/org/apache/livy/rsc/driver/JobWrapper.java
huianyi: Maybe Creating this time instance in `RSCDriver` is a good way?  Notify this timer when job submit, other times would be wait.

rsc/src/main/java/org/apache/livy/rsc/driver/JobWrapper.java
```diff
@@ -42,10 +45,17 @@
 
   private Future<?> future;
 
+  private final long firstDelayMSec = 500L;
+  private final long updatePeriodMSec;
+
+  private Timer timer = new Timer("refresh progress", true);
```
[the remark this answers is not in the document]
huianyi: sure.

rsc/src/main/java/org/apache/livy/rsc/RSCClient.java
```diff
@@ -330,6 +332,14 @@ public long getReplLastActivity() {
 
   private class ClientProtocol extends BaseProtocol {
 
+    private final String PROCESS_FORMAT = "stage %3s: percentage: %3d%% active: %s " +
+            "completed: %s failed: %s all: %s";
+    private ConcurrentBoundedLinkedQueue<String> operationMessages;
```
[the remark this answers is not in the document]
huianyi: Yes, this could be an issue, gonna change it.

huianyi:
> > Rethink about the way this ticket doing, I think current way may not be the best way.
> > 
> > 1. `jobId` may not exactly match the currently running jobs, there's no issue for beeline, but could be an issue for multi-thread usage. Currently when getting active jobs, there's no direct relation to jobId.
> > 2. There's already a way to calculate job progress in ReplDriver for statement, we should extend the current way to also apply to job API.
> > 3. Using LivyServer to store the session/job status may not be fault tolerant. It would better to extend the job API to get job progress synchronously.
> 
> Thanks for your reply, if I have understood your meaning correctly, actually I can use `jobId`(which is spark jobs' group id delivered by livy) in `handleProcessMessage` method of RSCDriver, then get related stages using this `jobId`, to handle these circumstances above. May this be a good way?

@jerryshao After a deep dive into this implement, I find it is difficult to get job group id from RSCDriver, at first I thought `jobId` might be the spark job group id(because in JobWrapper class, it calls set JobGroup with `jobId`), actually I'm wrong. Spark JobGroup is set as SqlJob's statementId, and if I want to get this statementId in JobWrapper or RscDriver, I must convert JobWrapper's Job to SqlJob, this needs SqlJob's definition, as SqlJob is in thriftserver module, importing its definition in rsc module may not be a good way. Could you please give me some advice?

captainzmc:
Hi @huianyi .  jerryshao  actually means that there are already methods in session.scala: def progressOfStatement(stmtId: Int), which we can use.
And if we need a SqlJob, maybe we can get it through getClass().

captainzmc:
> > > Rethink about the way this ticket doing, I think current way may not be the best way.
> > > 
> > > 1. `jobId` may not exactly match the currently running jobs, there's no issue for beeline, but could be an issue for multi-thread usage. Currently when getting active jobs, there's no direct relation to jobId.
> > > 2. There's already a way to calculate job progress in ReplDriver for statement, we should extend the current way to also apply to job API.
> > > 3. Using LivyServer to store the session/job status may not be fault tolerant. It would better to extend the job API to get job progress synchronously.
> > 
> > 
> > Thanks for your reply, if I have understood your meaning correctly, actually I can use `jobId`(which is spark jobs' group id delivered by livy) in `handleProcessMessage` method of RSCDriver, then get related stages using this `jobId`, to handle these circumstances above. May this be a good way?
> 
> @jerryshao After a deep dive into this implement, I find it is difficult to get job group id from RSCDriver, at first I thought `jobId` might be the spark job group id(because in JobWrapper class, it calls setting JobGroup with `jobId`), actually I'm wrong. Spark JobGroup is set as SqlJob's statementId, and if I want to get this statementId in JobWrapper or RscDriver, I must convert JobWrapper's Job to SqlJob, this needs SqlJob's definition, as SqlJob is in thriftserver module, importing its definition in rsc module may not be a good way. Could you please give me some advice?

@huianyi Reference thrift directly in the rsc will have circular references exception. We can use Java reflection in jobwrap.java, to avoid direct references. Such as:
`Field field = job.getClass().getDeclaredField("statementId");
  String statementId = field.get(job).toString();`

huianyi:
> > > > Rethink about the way this ticket doing, I think current way may not be the best way.
> > > > 
> > > > 1. `jobId` may not exactly match the currently running jobs, there's no issue for beeline, but could be an issue for multi-thread usage. Currently when getting active jobs, there's no direct relation to jobId.
> > > > 2. There's already a way to calculate job progress in ReplDriver for statement, we should extend the current way to also apply to job API.
> > > > 3. Using LivyServer to store the session/job status may not be fault tolerant. It would better to extend the job API to get job progress synchronously.
> > > 
> > > 
> > > Thanks for your reply, if I have understood your meaning correctly, actually I can use `jobId`(which is spark jobs' group id delivered by livy) in `handleProcessMessage` method of RSCDriver, then get related stages using this `jobId`, to handle these circumstances above. May this be a good way?
> > 
> > 
> > @jerryshao After a deep dive into this implement, I find it is difficult to get job group id from RSCDriver, at first I thought `jobId` might be the spark job group id(because in JobWrapper class, it calls setting JobGroup with `jobId`), actually I'm wrong. Spark JobGroup is set as SqlJob's statementId, and if I want to get this statementId in JobWrapper or RscDriver, I must convert JobWrapper's Job to SqlJob, this needs SqlJob's definition, as SqlJob is in thriftserver module, importing its definition in rsc module may not be a good way. Could you please give me some advice?
> 
> @huianyi Reference thrift directly in the rsc will have circular references exception. We can use Java reflection in jobwrap.java, to avoid direct references. Such as:
> `Field field = job.getClass().getDeclaredField("statementId"); String statementId = field.get(job).toString();`

@captainzmc Yes, this is a way to solve the problem, but using reflection is a costly method, especially called multiple times, is there a more decent way to implement this?

huianyi:
Is the new implement reasonable? If yes, I'm gonna change other left problems.

rsc/src/main/java/org/apache/livy/rsc/driver/RSCDriver.java
```diff
@@ -212,6 +212,41 @@ public void onSaslComplete(Rpc client) {
...
+   */
+  void handleProcessMessage(String sqlJobStatementId) {
+    SparkStatusTracker sparkStatusTracker;
+    synchronized (jcLock) {
+      if (jc == null) {
+        return;
+      }
+      sparkStatusTracker = jc.sc().sc().statusTracker();
+    }
+    int[] jobIdsForGroup = sparkStatusTracker.getJobIdsForGroup(sqlJobStatementId);
```
jerryshao: Where did you set this `sqlJobStatementId`?

huianyi: SqlJob will use statementId to set sparkJobGroup, [SqlJob.java#L60](https://github.com/apache/incubator-livy/blob/c7313c5648c63089413464b47f33c9749dbabb20/thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/SqlJob.java#L60) Let Sql job extends abstract class SqlJobStatement, so we can get its statementId.

jerryshao: I see, thanks.

jerryshao:
I would like to hear the comments from others. Here what I saw is that:

1. Each job will start a Timer, which is quite cost if the job is a very small job (metadata job).
2. The current way to get progress is using push based way, I was wondering if it is better to change to pull based way, so that we don't need to store such message in server side. Also we don't have to formalize the message in server side. For example, request: client -> server -> driver, and response: driver->server->client. Like what current repl did.

@mgaido91 can you please help to review this PR? Thanks.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
captainzmc: Currently, the trackingURL is only printed during session initialization. We'd better put this trackingURL LivyExecuteStatementOperation, ensure that every time we execute SQL can print this trackingURL.

huianyi:
@mgaido91 Thanks for your reply, hive can be treated as hadoop client, it just call hadoop api to get job status (Class `Job` of hadoop code), therefore in hive, more specifically, hiveserver, it uses pull based way to get job process and then formalize the message, and put it in logs(like our operationMessage object), the client(beeline client) will pull these messages in a loop. Using pull based way in client side is a reasonable way, I'm gonna change this implement.

I don't think your main concern is a problem, on the one hand, we print detailed process messages of the job(like completed, active, etc), to some entent, these messages, which can be update in a certain intervel, can give end users enough information of their jobs, so at least they know how their jobs are running. On the other hand, I don't think there is a truly trusted percentage in a distribute system, hive may also show wrong percentages if a task is failed or hang there.

huianyi:
> 1. Each job will start a Timer, which is quite cost if the job is a very small job (metadata job).

@jerryshao we can not get detailed information of the job(such as if it is a metadata job) from livy side, so we might need to start a Timer for every job. Is there a more decent way?

huianyi:
> 1. Each job will start a Timer, which is quite cost if the job is a very small job (metadata job).

@jerryshao we can not get detailed information of the job(such as if it is a metadata job) from livy side, so we might need to start a Timer for every job. also, spark use Timer to get job process(`ConsoleProgressBar`), maybe this cost is acceptable?

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
huianyi: This trackingURL normally will not changed, so print it only once is enough?

huianyi:
@jerryshao @mgaido91 Could you please review the new update?

mgaido91:
@huianyi I don't see the previous comments being resolved. You're still keeping logs on server side with a timer instead of giving the progress on client's request.

So my point is: Hive handles this in a different way. Please see https://github.com/apache/hive/blob/209096a71b25a3d75dca1930f825c8c10b99d2b9/service/src/java/org/apache/hive/service/cli/CLIService.java#L482.

I think we should try and handle it that way if possible. If it is not possible, we need to clarify why it is not and why we have no other choices than this one.

huianyi:
@mgaido91 Have I missed something? If we use beeline to connect to hiveserver, hiveserver will catch each job status from AppMaster and save these messages in operation logs, which are files located in the hiveserver‘s machine. Clients will pull these messages in a while loop.

At first, I achieve this by using push based way, RscDriver will push its job status to thriftserver, and In the current implement, our thriftserver will pull these job status from RscDriver and stored it in queue(OperationMessages). Beeline client also will pull these messages in a while loop.

Please correct me if I'm wrong, thanks.

mgaido91:
@huianyi I think what we would like to achieve here is: do not use the operation logs, so that we don't have to keep anything on server side (by this I mean the Livy Server) (and hence avoid to put memory pressure on it).

The ideal solution would be:
 - the beeline client polls for the status;
 - Livy server queries the spark session for the query status:
 - Livy server forwards back to the beeline client the status.

In this way, we need to store nothing on Livy server memory.

This seems to me to be the approach used in Hive. My understanding is the Hive informs the beeline client about the status and progress of the operation with the code I pointed you out in the previous link. So I think we should follow that approach to give feedbacks on the query status/progress.

Hope this is more clear. If not, please ask what it is not. Thanks.

huianyi:
@mgaido91 The previous link you pointed actually is another feature we can implement(process bar), you can check it here, [HiveStatement.java#L381](https://github.com/apache/hive/blob/209096a71b25a3d75dca1930f825c8c10b99d2b9/jdbc/src/java/org/apache/hive/jdbc/HiveStatement.java#L381) , [BeelineInPlaceUpdateStream.java#L39](https://github.com/apache/hive/blob/209096a71b25a3d75dca1930f825c8c10b99d2b9/beeline/src/java/org/apache/hive/beeline/logs/BeelineInPlaceUpdateStream.java#L39)

Like I previous said, I can support this in a new PR.

> > > Beeline from Hive 2.2 has in-place progress bar hive.server2.in.place.progress
> > > https://cwiki.apache.org/confluence/display/Hive/Configuration+Properties#ConfigurationProperties-hive.server2.in.place.progress
> > > (I think it works only with tez - would be nice if this would work with Livy too )
> > 
> > 
> > Would be better to support this if it is not a big effort.
> 
> After reading hive related code, this needs write some new classes to implement this feature, which might cause this PR large, if you don not mind, I'd love to open a new PR to support this.

ProcessBar is great in real-time interaction environment, but it can not work well if we do off-line scheduling jobs, we should print the job's status by individual lines in this circumstances.

> In this way, we need to store nothing on Livy server memory.

By default setting, we can keep each session keep not more than 100 rows, actually under normal circumstances, client will get these message immediately, this might not put much memory pressure on Livy server side.

mgaido91:
@huianyi 

> Like I previous said, I can support this in a new PR.

That sound great, I think you can go ahead, I am happy to review it.

> ProcessBar is great in real-time interaction environment, but it can not work well if we do scheduling jobs, which we need print the job's status in seperate lines.

I don't see the point here. Moreover I think for monitoring such job other method would be more appropriate, like querying the REST API for the status of the job.

My position is to go ahead with the progress bar, but I am not very confident about this feature. Otherwise also other people think that this is useful, I'd go only with the progress bar feature.

Thanks.

huianyi:
@mgaido91 

> I don't see the point here. Moreover I think for monitoring such job other method would be more appropriate, like querying the REST API for the status of the job.

1. For ad hoc querys, we can use process bar to get job status.
2. In most circumstances, we use livy for offline schedule tasks, like using airflow to submit job at a specific time, which needs to get all job process status, process bar can not support this.

Hive support these two implements, it actually calls yarn api to get job status, which is a while loop calling AppMaster(You can check it [here](https://github.com/apache/hadoop/blob/e3496499ecb8d220fba99dc5ed4c99c8f9e33bb1/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/Job.java#L1338) )

huianyi:
@yiheng @jerryshao Could you please also review the new update and give some advice? I think hive also implement it in this way.  Actually this new feature is already in our prodbuction environment, and it works pretty well.

Thanks.

PR #239, opened 2019-09-27
[LIVY-690][Thrift] Exclude curator in thrift server pom to avoid conflict jars

yiheng:
Currently, thrift server has a dependency of curator-client:2.12.0 through the hive service. After the build, a `curator-client-2.12.0.jar` file will be generated in the `jars` folder. It is conflicted with the `curator-client-2.7.1.jar` file, which is used by livy server.

We observed that in some JDK, the `curator-client-2.12.0.jar` is loaded before the `curator-client-2.7.1.jar`, and will crash the recovery enabled livy server.

In this patch, we exclude the `org.apache.curator` modules from the hive dependency. 

## How was this patch tested?
Manual test and existing UT/ITs

### DIFF
--- a/thriftserver/server/pom.xml
+++ b/thriftserver/server/pom.xml
@@ -58,6 +58,10 @@
           <groupId>org.eclipse.jetty</groupId>
           <artifactId>*</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>org.apache.curator</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
       </exclusions>
     </dependency>
 


yiheng:
@mgaido91 Hi Marco. It's a small fix. Can you take a look at? Thanks

mgaido91:
@yiheng thanks for the fix! May you please create a JIRA and reference it in the title of this PR please? Thanks.

jerryshao:
@yiheng would you please fix this? Thanks.

>@yiheng thanks for the fix! May you please create a JIRA and reference it in the title of this PR please? Thanks.

yiheng:
done

yiheng:
Please hold on this patch. I just notice there're other duplicate jars. Let me check...

```bash
-rw-r--r-- 1 root root   966102 Sep 29 14:53 apache-jsp-8.0.33.jar
-rw-r--r-- 1 root root    11012 Sep 29 14:53 apache-jsp-9.3.24.v20180605.jar
...
-rw-r--r-- 1 root root    38015 Sep 29 14:53 commons-logging-1.0.4.jar
-rw-r--r-- 1 root root    61829 Sep 29 14:52 commons-logging-1.2.jar
...
-rw-r--r-- 1 root root   177131 Sep 29 14:52 jetty-util-6.1.26.jar
-rw-r--r-- 1 root root   458642 Sep 29 14:52 jetty-util-9.3.24.v20180605.jar
...
-rw-r--r-- 1 root root    50493 Sep 29 14:53 jsp-api-2.0.jar
-rw-r--r-- 1 root root   100636 Sep 29 14:52 jsp-api-2.1.jar
...
-rw-r--r-- 1 root root     8094 Sep 29 14:52 metrics-servlet-3.1.1.jar
-rw-r--r-- 1 root root    16571 Sep 29 14:52 metrics-servlets-3.1.1.jar
...
-rw-r--r-- 1 root root  1208356 Sep 29 14:53 netty-3.7.0.Final.jar
-rw-r--r-- 1 root root  1330219 Sep 29 14:52 netty-3.9.9.Final.jar
```

yiheng:
None of the above duplications is caused by the thrift server. And so far haven't found any related issue. I think we can fix the curator duplication first. I will open a JIRA to track the jar duplication.

jerryshao:
Thanks @yiheng , merging to master branch.

PR #240, opened 2019-09-30
[LIVY-541][WIP] Generate the unqiue session id when multiple livy servers work at the same time

runzhiwang:

[LIVY-541] Generate the unqiue session id when multiple livy servers work at the same time

1. When generate unique session id with multiple livy servers working. First, get the distributed lock,
  Second, get the session id from the filesystem or zookeeper. Third, increase the sessionid and save it in  the filesystem or zookeeper. Forth, release the distributed lock.

2.  ZooKeeperManager provides the distributed lock to generate the unique session id.

3. If set recovery.mode = off, i.e. generate session id in BlackholeStateStore,  and set multi-active.enabled = true, store the session id in zookeeper.
If set recovery.mode = off, and set multi-active.enabled = false, store the session id in memory, i.e. stateMap

## How was this patch tested?

1. Start multiple livy servers and set multi-active.enabled = true.
2. Set zookeeper.url or state-store.url to save session id.
3. Create session in different livy servers at the same time.
4. Check the session id without duplicating.

### DIFF
--- a/server/src/main/scala/org/apache/livy/server/recovery/StateStore.scala
+++ b/server/src/main/scala/org/apache/livy/server/recovery/StateStore.scala
@@ -22,6 +22,7 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
 
 import org.apache.livy.{LivyConf, Logging}
+import org.apache.livy.server.recovery.ZooKeeperManager
 import org.apache.livy.sessions.SessionKindModule
 import org.apache.livy.sessions.SessionManager._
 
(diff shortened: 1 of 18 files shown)


jerryshao:
Let's wait this until the whole HA design and plan settles down.

server/src/main/scala/org/apache/livy/server/recovery/BlackholeStateStore.scala
```diff
@@ -25,12 +26,42 @@ import org.apache.livy.LivyConf
...
  * Livy will use this when session recovery is disabled.
  */
-class BlackholeStateStore(livyConf: LivyConf) extends StateStore(livyConf) {
-  def set(key: String, value: Object): Unit = {}
+class BlackholeStateStore(livyConf: LivyConf, zkManager: Option[ZooKeeperManager])
+  extends StateStore(livyConf, zkManager) {
+  private val stateMap: mutable.Map[String, Object] = mutable.Map()
 
-  def get[T: ClassTag](key: String): Option[T] = None
+  def set(key: String, value: Object): Unit = {
```
yiheng: why change the set/get behavior?

server/src/main/scala/org/apache/livy/server/recovery/StateStore.scala
```diff
@@ -43,7 +43,8 @@ protected trait JsonMapper {
  *
  * Hardcoded to use JSON serialization for now for easier ops. Will add better serialization later.
  */
-abstract class StateStore(livyConf: LivyConf) extends JsonMapper {
+abstract class StateStore(livyConf: LivyConf, zkManager: Option[ZooKeeperManager])
```
yiheng: should we consider moving the zkManager to SessionStore?

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
```diff
@@ -71,7 +71,6 @@ class SessionManager[S <: Session, R <: RecoveryMetadata : ClassTag](
 
   protected implicit def executor: ExecutionContext = ExecutionContext.global
 
-  protected[this] final val idCounter = new AtomicInteger(0)
```
yiheng: Maybe wrap the logic to the IdCounter class?

PR #241, opened 2019-10-05
[LIVY-692][THRIFT] Use configured principal for Kerberos authentication

wypoon:

https://issues.apache.org/jira/browse/LIVY-692
Currently, if Kerberos authentication is used in the Thrift server (livy.server.thrift.authentication=kerberos), the `TTransportFactory` used in the `ThriftCLIService` uses a `UserGroupInformation` for the service (livy) principal; instead it should use a `UserGroupInformation` for the principal specified in livy.server.auth.kerberos.principal.

## How was this patch tested?

Manual testing with both a Livy Thrift server using Kerberos authentication and one using simple authentication.

### DIFF
--- a/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/auth/AuthBridgeServer.scala
+++ b/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/auth/AuthBridgeServer.scala
@@ -40,12 +40,9 @@ import org.apache.livy.Logging
  * The class is taken from Hive's `HadoopThriftAuthBridge.Server`. It bridges Thrift's SASL
  * transports to Hadoop's SASL callback handlers and authentication classes.
  */
-class AuthBridgeServer(private val secretManager: LivyDelegationTokenSecretManager) {
-  private val ugi = try {
-      UserGroupInformation.getCurrentUser
-    } catch {
-      case ioe: IOException => throw new TTransportException(ioe)
-    }
+class AuthBridgeServer(
+    private val secretManager: LivyDelegationTokenSecretManager,
+    private val ugi: UserGroupInformation) {
 
   /**
    * Create a TTransportFactory that, upon connection of a client socket,
--- a/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/auth/AuthFactory.scala
+++ b/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/auth/AuthFactory.scala
@@ -25,6 +25,7 @@ import javax.security.sasl.{AuthorizeCallback, Sasl}
 
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.CommonConfigurationKeysPublic.HADOOP_SECURITY_AUTHENTICATION
+import org.apache.hadoop.security.{SecurityUtil, UserGroupInformation}
 import org.apache.hadoop.security.SaslRpcServer.AuthMethod
 import org.apache.hive.service.auth.{SaslQOP, TSetIpAddressProcessor}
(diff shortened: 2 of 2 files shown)


mgaido91:
I don't think this patch is correct. I think the current behavior is correct. Please see {{LivyCLIService}} where the two tokens are created. If we'd go ahead with this patch, impersonation would break on many clusters, as it may/should not be enabled on the http UGI.

wypoon:
@mgaido91 binary mode seems to work with this patch. Let me look into http mode.

mgaido91:
@wypoon have you checked also impersonation? Moreover, I think the current behavior is correct and I don't see any benefit introduced by this patch, rather than breaking the existing behavior.

wypoon:
By impersonation, do you mean that the Spark application Livy spawns (to run queries) is run as the user connecting to the Thrift server? I have a user kinit'd, let's say using principal testuser, who connects via beeline to the Thrift server; and the Spark application for the session runs as testuser. If this is what you mean, then impersonation is working.

I tested my patch using both binary and http mode. In binary mode, everything works as I expect. In http mode, everything still works, except that on the server side, there are many exceptions logged, due to
`Caused by: org.apache.hive.service.auth.HttpAuthenticationException: Authorization header received from the client is empty.
`
but they do not cause any failure. I believe these exceptions are an unrelated issue.

Without my patch, I am able to connect when using http mode (the same slew of exceptions about the empty authorization header also occur in the server log), but I am unable to connect when using binary mode, due to the error as described in JIRA.
Thus I fail to see why you say the existing behavior is correct.

This patch fixes the binary mode case, and as far as I can tell, does not break existing behavior.
If you think the existing behavior is correct, and that the failure I experience in binary mode is due to some kind of misconfiguration or user error, I'd very much appreciate knowing what configuration I need to correct or what I am missing in my beeline command. Please comment in JIRA with the information.

mgaido91:
I think the point is that in your JIRA you are using HTTP/... as principal when you connect to beeline. You should use the principal Livy server is using instead.

> I have a user kinit'd, let's say using principal testuser, who connects via beeline to the Thrift server; and the Spark application for the session runs as testuser. If this is what you mean, then impersonation is working.

yes, that is impersonation. Yes, it may not be a problem because in your case you're forcing to use the HTTP principal only for the authentication, but the livy one is used everywhere else, the problem in this case would be token renewal most likely...you are creating the UGI once and never renewing it, so it would not work after 1 week or even 1 day probably.

wypoon:
The reason I use the HTTP principal in my beeline connection is because that is the principal configured in livy.server.auth.kerberos.principal. If I use the livy principal in my beeline connection, the connection still fails with the "GSS initiate failed" error. (This is with the existing code, before I made any changes.)
I understand that my patch may not be correct (not dealing with token renewal), but there is a real issue here.
Concerning renewal, can you point me to where/how the current code takes care of token renewal?

mgaido91:
> If I use the livy principal in my beeline connection, the connection still fails with the "GSS initiate failed" error. (This is with the existing code, before I made any changes.)

this is weird. I'd need to investigate this but I don't have a cluster so I am not sure how to help you.

> Concerning renewal, can you point me to where/how the current code takes care of token renewal?

Please check the `LivyServer` class.

Tagar:
> GSS initiate failed

can you share full exception stack? perhaps it's some sort of kerberos client misconfiguration on 
livy client server, or livy server side.. hard to tell without knowing details.

wypoon:
I updated the JIRA to explain that the problem was due to using an internal version of Livy that had some changes incompatible with the Thrift server; the changes caused the UGI for the livy principal to be obtained from ticket cache instead of keytab; this is what caused the "GSS initiate failed" error when trying to connect from beeline.
Closing this PR.

PR #242, opened 2019-10-06
[LIVY-336][SERVER] Livy should not spawn one thread per job to track the job on Yarn

runzhiwang:

Instead of spawning a monitor thread for every session, create a centralized thread to monitor all apps.

## How was this patch tested?

Existed ITs and UTs.

### DIFF
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -119,8 +119,13 @@
 # For zookeeper, the address to the Zookeeper servers. e.g. host1:port1,host2:port2
 # livy.server.recovery.state-store.url =
 
+# If Livy can't find the yarn app within this time, consider it lost.
+# livy.server.yarn.app-lookup-timeout = 120s
 # If Livy can't find the yarn app within this max times, consider it lost.
-# livy.server.yarn.app-lookup.max-failed.times = 120
+# livy.server.yarn.app-lookup.max-failed.times = 30
+# The size of thread pool to monitor all yarn apps.
+# livy.server.yarn.app-lookup.thread-pool.size = 4
+
 # When the cluster is busy, we may fail to launch yarn app in app-lookup-timeout, then it would
 # cause session leakage, so we need to check session leakage.
 # How long to check livy session leakage.
@@ -128,18 +133,9 @@
 # How often to check livy session leakage.
 # livy.server.yarn.app-leakage.check-interval = 60s
 
-# The size of thread pool to monitor all yarn apps.
-# livy.server.yarn.app-monitor.thread-pool.size = 4
-# How often thread monitor the YARN app.
-# livy.server.yarn.app-monitor.thread-interval = 5s
-# How often to check monitor thread block
-# livy.server.yarn.app-monitor.thread-block.check-interval = 10s
-# If some thread cost more than this config to monitor one app,
-# stop the monitor of the app which considered blocked.
-# livy.server.yarn.app-monitor.thread-block-timeout = 60s
-# If Livy can't monitor the yarn app successfully within this max times, consider the app failed.
-# livy.server.yarn.app-monitor.max-failed.times = 12
-
+# How often Livy polls YARN to refresh YARN app state.
+# livy.server.yarn.poll-interval = 5s
+#
 # Days to keep Livy server request logs.
 # livy.server.request-log-retain.days = 5
 
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -121,20 +121,25 @@
 
 # If Livy can't find the yarn app within this max times, consider it lost.
 # livy.server.yarn.app-lookup.max-failed.times = 120
-# How often to poll yarn for the app id.
-# livy.server.yarn.app-lookup-interval = 1s
 # When the cluster is busy, we may fail to launch yarn app in app-lookup-timeout, then it would
 # cause session leakage, so we need to check session leakage.
 # How long to check livy session leakage.
 # livy.server.yarn.app-leakage.check-timeout = 600s
 # How often to check livy session leakage.
 # livy.server.yarn.app-leakage.check-interval = 60s
+
 # The size of thread pool to monitor all yarn apps.
 # livy.server.yarn.app-monitor.thread-pool.size = 4
+# How often thread monitor the YARN app.
+# livy.server.yarn.app-monitor.thread-interval = 5s
+# How often to check monitor thread block
+# livy.server.yarn.app-monitor.thread-block.check-interval = 10s
+# If some thread cost more than this config to monitor one app,
+# stop the monitor of the app which considered blocked.
+# livy.server.yarn.app-monitor.thread-block-timeout = 60s
+# If Livy can't monitor the yarn app successfully within this max times, consider the app failed.
+# livy.server.yarn.app-monitor.max-failed.times = 12
 
-# How often Livy polls YARN to refresh YARN app state.
-# livy.server.yarn.poll-interval = 5s
-#
 # Days to keep Livy server request logs.
 # livy.server.request-log-retain.days = 5
 
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -122,7 +122,7 @@
 # If Livy can't find the yarn app within this time, consider it lost.
 # livy.server.yarn.app-lookup-timeout = 120s
 # If Livy can't find the yarn app within this max times, consider it lost.
-# livy.server.yarn.app-lookup.max-failed.times = 30
+# livy.server.yarn.app-lookup.max-failed.times = 120
 # The size of thread pool to monitor all yarn apps.
 # livy.server.yarn.app-lookup.thread-pool.size = 4
 
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -133,6 +133,8 @@
 # livy.server.yarn.app-monitor.timeout = 10s
 # If Livy can't monitor the yarn app successfully within this max times, consider the app failed.
 # livy.server.yarn.app-monitor.max-failed.times = 12
+# The size of thread pool to monitor all yarn apps.
+# livy.server.yarn.app-monitor.thread-pool.size = 4
 
 # How often Livy polls YARN to refresh YARN app state.
 # livy.server.yarn.poll-interval = 5s
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -121,15 +121,21 @@
 
 # If Livy can't find the yarn app within this time, consider it lost.
 # livy.server.yarn.app-lookup-timeout = 120s
+# How long to find the yarn app
+# livy.server.yarn.app-lookup-interval = 1s
 # When the cluster is busy, we may fail to launch yarn app in app-lookup-timeout, then it would
 # cause session leakage, so we need to check session leakage.
 # How long to check livy session leakage
 # livy.server.yarn.app-leakage.check-timeout = 600s
 # how often to check livy session leakage
 # livy.server.yarn.app-leakage.check-interval = 60s
+# how long to monitor the yarn app
+# livy.server.yarn.app-monitor.timeout = 10000ms
+# If Livy can't monitor the yarn app successfully within this max times, consider the app failed.
+# livy.server.yarn.app-monitor.max-failed.times = 120
 
 # How often Livy polls YARN to refresh YARN app state.
-# livy.server.yarn.poll-interval = 5s
+# livy.server.yarn.poll-interval = 500ms
 #
 # Days to keep Livy server request logs.
 # livy.server.request-log-retain.days = 5
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -121,18 +121,18 @@
 
 # If Livy can't find the yarn app within this time, consider it lost.
 # livy.server.yarn.app-lookup-timeout = 120s
-# How long to find the yarn app
+# How often to poll yarn for the app id.
 # livy.server.yarn.app-lookup-interval = 1s
 # When the cluster is busy, we may fail to launch yarn app in app-lookup-timeout, then it would
 # cause session leakage, so we need to check session leakage.
-# How long to check livy session leakage
+# How long to check livy session leakage.
 # livy.server.yarn.app-leakage.check-timeout = 600s
-# how often to check livy session leakage
+# How often to check livy session leakage.
 # livy.server.yarn.app-leakage.check-interval = 60s
-# how long to monitor the yarn app
-# livy.server.yarn.app-monitor.timeout = 10000ms
+# If Livy can't monitor yarn app successfully within this time, consider the monitoring failed.
+# livy.server.yarn.app-monitor.timeout = 10s
 # If Livy can't monitor the yarn app successfully within this max times, consider the app failed.
-# livy.server.yarn.app-monitor.max-failed.times = 120
+# livy.server.yarn.app-monitor.max-failed.times = 12
 
 # How often Livy polls YARN to refresh YARN app state.
 # livy.server.yarn.poll-interval = 5s
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -119,8 +119,8 @@
 # For zookeeper, the address to the Zookeeper servers. e.g. host1:port1,host2:port2
 # livy.server.recovery.state-store.url =
 
-# If Livy can't find the yarn app within this time, consider it lost.
-# livy.server.yarn.app-lookup-timeout = 120s
+# If Livy can't find the yarn app within this max times, consider it lost.
+# livy.server.yarn.app-lookup.max-failed.times = 120
 # How often to poll yarn for the app id.
 # livy.server.yarn.app-lookup-interval = 1s
 # When the cluster is busy, we may fail to launch yarn app in app-lookup-timeout, then it would
@@ -129,10 +129,6 @@
 # livy.server.yarn.app-leakage.check-timeout = 600s
 # How often to check livy session leakage.
 # livy.server.yarn.app-leakage.check-interval = 60s
-# If Livy can't monitor yarn app successfully within this time, consider the monitoring failed.
-# livy.server.yarn.app-monitor.timeout = 10s
-# If Livy can't monitor the yarn app successfully within this max times, consider the app failed.
-# livy.server.yarn.app-monitor.max-failed.times = 12
 # The size of thread pool to monitor all yarn apps.
 # livy.server.yarn.app-monitor.thread-pool.size = 4
 
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -135,7 +135,7 @@
 # livy.server.yarn.app-monitor.max-failed.times = 120
 
 # How often Livy polls YARN to refresh YARN app state.
-# livy.server.yarn.poll-interval = 500ms
+# livy.server.yarn.poll-interval = 5s
 #
 # Days to keep Livy server request logs.
 # livy.server.request-log-retain.days = 5
--- a/server/src/main/scala/org/apache/livy/LivyConf.scala
+++ b/server/src/main/scala/org/apache/livy/LivyConf.scala
@@ -202,8 +202,18 @@ object LivyConf {
   // Livy will cache the max no of logs specified. 0 means don't cache the logs.
   val SPARK_LOGS_SIZE = Entry("livy.cache-log.size", 200)
 
+  // If Livy can't find the yarn app within this time, consider it lost.
+  val YARN_APP_LOOKUP_TIMEOUT = Entry("livy.server.yarn.app-lookup-timeout", "120s")
+
   // If Livy can't find the yarn app within this max times, consider it lost.
   val YARN_APP_LOOKUP_MAX_FAILED_TIMES = Entry("livy.server.yarn.app-lookup.max-failed.times", 120)
+
+  // The size of thread pool to monitor all yarn apps.
+  val YARN_APP_LOOKUP_THREAD_POOL_SIZE =
+    Entry("livy.server.yarn.app-lookup.thread-pool.size", 4)
+
+  // How often Livy polls YARN to refresh YARN app state.
+  val YARN_POLL_INTERVAL = Entry("livy.server.yarn.poll-interval", "5s")
 
   // Days to keep Livy server request logs.
   val REQUEST_LOG_RETAIN_DAYS = Entry("livy.server.request-log-retain.days", 5)
@@ -217,22 +227,6 @@ object LivyConf {
   val YARN_APP_LEAKAGE_CHECK_TIMEOUT = Entry("livy.server.yarn.app-leakage.check-timeout", "600s")
   // How often to check livy session leakage.
   val YARN_APP_LEAKAGE_CHECK_INTERVAL = Entry("livy.server.yarn.app-leakage.check-interval", "60s")
-
-  // The size of thread pool to monitor all yarn apps.
-  val YARN_APP_MONITOR_THREAD_POOL_SIZE =
-    Entry("livy.server.yarn.app-monitor.thread-pool.size", 4)
-  // How often thread monitor the YARN app.
-  val YARN_APP_MONITOR_THREAD_INTERVAL = Entry("livy.server.yarn.app-monitor.thread-interval", "5s")
-  // How often to check monitor thread block
-  val YARN_APP_MONITOR_THREAD_BLOCK_CHECK_INTERVAL =
-    Entry("livy.server.yarn.app-monitor.thread-block.check-interval", "10s")
-  // If some thread cost more than this config to monitor one app,
-  // stop the monitor of the app which considered blocked.
-  val YARN_APP_MONITOR_THREAD_BLOCK_TIMEOUT =
-    Entry("livy.server.yarn.app-monitor.thread-block-timeout", "60s")
-  // If Livy can't monitor the yarn app successfully within this max times, consider the app failed.
-  val YARN_APP_MONITOR_MAX_FAILED_TIMES =
-    Entry("livy.server.yarn.app-monitor.max-failed.times", 12)
 
   // Whether session timeout should be checked, by default it will be checked, which means inactive
   // session will be stopped after "livy.server.session.timeout"
--- a/server/src/main/scala/org/apache/livy/LivyConf.scala
+++ b/server/src/main/scala/org/apache/livy/LivyConf.scala
@@ -204,7 +204,7 @@ object LivyConf {
 
   // If Livy can't find the yarn app within this time, consider it lost.
   val YARN_APP_LOOKUP_TIMEOUT = Entry("livy.server.yarn.app-lookup-timeout", "120s")
-  // How long to find the yarn app.
+  // How often to poll yarn for the app id.
   val YARN_APP_LOOKUP_INTERVAL = Entry("livy.server.yarn.app-lookup-interval", "1s")
 
   // How often Livy polls YARN to refresh YARN app state.
@@ -218,15 +218,15 @@ object LivyConf {
   // RSC related jars separated with comma.
   val RSC_JARS = Entry("livy.rsc.jars", null)
 
-  // How long to check livy session leakage
+  // How long to check livy session leakage.
   val YARN_APP_LEAKAGE_CHECK_TIMEOUT = Entry("livy.server.yarn.app-leakage.check-timeout", "600s")
-  // how often to check livy session leakage
+  // How often to check livy session leakage.
   val YARN_APP_LEAKAGE_CHECK_INTERVAL = Entry("livy.server.yarn.app-leakage.check-interval", "60s")
-  // how long to monitor the yarn app
-  val YARN_APP_MONITOR_TIMEOUT = Entry("livy.server.yarn.app-monitor.timeout", "10000ms")
+  // If Livy can't monitor yarn app successfully within this time, consider the monitoring failed.
+  val YARN_APP_MONITOR_TIMEOUT = Entry("livy.server.yarn.app-monitor.timeout", "10s")
   // If Livy can't monitor the yarn app successfully within this max times, consider the app failed.
   val YARN_APP_MONITOR_MAX_FAILED_TIMES =
-    Entry("livy.server.yarn.app-monitor.max-failed.times", "120")
+    Entry("livy.server.yarn.app-monitor.max-failed.times", "12")
 
   // Whether session timeout should be checked, by default it will be checked, which means inactive
   // session will be stopped after "livy.server.session.timeout"
--- a/server/src/main/scala/org/apache/livy/LivyConf.scala
+++ b/server/src/main/scala/org/apache/livy/LivyConf.scala
@@ -228,7 +228,8 @@ object LivyConf {
   val YARN_APP_MONITOR_MAX_FAILED_TIMES =
     Entry("livy.server.yarn.app-monitor.max-failed.times", "12")
   // The size of thread pool to monitor all yarn apps.
-  val YARN_APP_MONITOR_THREAD_POOL_SIZE = Entry("livy.server.yarn.app-monitor.thread-pool.size", "4")
+  val YARN_APP_MONITOR_THREAD_POOL_SIZE =
+    Entry("livy.server.yarn.app-monitor.thread-pool.size", "4")
 
   // Whether session timeout should be checked, by default it will be checked, which means inactive
   // session will be stopped after "livy.server.session.timeout"
--- a/server/src/main/scala/org/apache/livy/LivyConf.scala
+++ b/server/src/main/scala/org/apache/livy/LivyConf.scala
@@ -205,9 +205,6 @@ object LivyConf {
   // If Livy can't find the yarn app within this max times, consider it lost.
   val YARN_APP_LOOKUP_MAX_FAILED_TIMES = Entry("livy.server.yarn.app-lookup.max-failed.times", 120)
 
-  // How often Livy polls YARN to refresh YARN app state.
-  val YARN_POLL_INTERVAL = Entry("livy.server.yarn.poll-interval", "5s")
-
   // Days to keep Livy server request logs.
   val REQUEST_LOG_RETAIN_DAYS = Entry("livy.server.request-log-retain.days", 5)
 
@@ -220,9 +217,22 @@ object LivyConf {
   val YARN_APP_LEAKAGE_CHECK_TIMEOUT = Entry("livy.server.yarn.app-leakage.check-timeout", "600s")
   // How often to check livy session leakage.
   val YARN_APP_LEAKAGE_CHECK_INTERVAL = Entry("livy.server.yarn.app-leakage.check-interval", "60s")
+
   // The size of thread pool to monitor all yarn apps.
   val YARN_APP_MONITOR_THREAD_POOL_SIZE =
     Entry("livy.server.yarn.app-monitor.thread-pool.size", 4)
+  // How often thread monitor the YARN app.
+  val YARN_APP_MONITOR_THREAD_INTERVAL = Entry("livy.server.yarn.app-monitor.thread-interval", "5s")
+  // How often to check monitor thread block
+  val YARN_APP_MONITOR_THREAD_BLOCK_CHECK_INTERVAL =
+    Entry("livy.server.yarn.app-monitor.thread-block.check-interval", "10s")
+  // If some thread cost more than this config to monitor one app,
+  // stop the monitor of the app which considered blocked.
+  val YARN_APP_MONITOR_THREAD_BLOCK_TIMEOUT =
+    Entry("livy.server.yarn.app-monitor.thread-block-timeout", "60s")
+  // If Livy can't monitor the yarn app successfully within this max times, consider the app failed.
+  val YARN_APP_MONITOR_MAX_FAILED_TIMES =
+    Entry("livy.server.yarn.app-monitor.max-failed.times", 12)
 
   // Whether session timeout should be checked, by default it will be checked, which means inactive
   // session will be stopped after "livy.server.session.timeout"
--- a/server/src/main/scala/org/apache/livy/LivyConf.scala
+++ b/server/src/main/scala/org/apache/livy/LivyConf.scala
@@ -204,9 +204,11 @@ object LivyConf {
 
   // If Livy can't find the yarn app within this time, consider it lost.
   val YARN_APP_LOOKUP_TIMEOUT = Entry("livy.server.yarn.app-lookup-timeout", "120s")
+  // How long to find the yarn app.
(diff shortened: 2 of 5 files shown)


conf/livy.conf.template
```diff
@@ -121,15 +121,21 @@
...
 # how often to check livy session leakage
 # livy.server.yarn.app-leakage.check-interval = 60s
+# how long to monitor the yarn app
+# livy.server.yarn.app-monitor.timeout = 10000ms
+# If Livy can't monitor the yarn app successfully within this max times, consider the app failed.
+# livy.server.yarn.app-monitor.max-failed.times = 120
 
 # How often Livy polls YARN to refresh YARN app state.
-# livy.server.yarn.poll-interval = 5s
+# livy.server.yarn.poll-interval = 500ms
```
jerryshao: I'm afraid this will bring overhead to YARN if it is a large cluster.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -33,16 +33,23 @@ import org.apache.hadoop.yarn.conf.YarnConfiguration
...
     sessionLeakageCheckInterval = livyConf.getTimeAsMs(LivyConf.YARN_APP_LEAKAGE_CHECK_INTERVAL)
     sessionLeakageCheckTimeout = livyConf.getTimeAsMs(LivyConf.YARN_APP_LEAKAGE_CHECK_TIMEOUT)
+    yarnPollInterval = livyConf.getTimeAsMs(LivyConf.YARN_POLL_INTERVAL)
+    yarnAppMonitorTimeout = livyConf.getTimeAsMs(LivyConf.YARN_APP_MONITOR_TIMEOUT)
+    yarnAppMonitorMaxFailedTimes = livyConf.getTimeAsMs(LivyConf.YARN_APP_MONITOR_MAX_FAILED_TIMES)
+    yarnAppLookUpInterval = livyConf.getTimeAsMs(LivyConf.YARN_APP_LOOKUP_INTERVAL)
     leakedAppsGCThread.setDaemon(true)
     leakedAppsGCThread.setName("LeakedAppsGCThread")
     leakedAppsGCThread.start()
+    yarnAppMonitorThread.setDaemon(true)
```
jerryshao: Add necessary empty line between different code paragraphs.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -56,17 +63,24 @@ object SparkYarnApp extends Logging {
...
     livyConf.getTimeAsMs(LivyConf.YARN_APP_LOOKUP_TIMEOUT) milliseconds
 
-  private def getYarnPollInterval(livyConf: LivyConf): FiniteDuration =
-    livyConf.getTimeAsMs(LivyConf.YARN_POLL_INTERVAL) milliseconds
-
   private val appType = Set("SPARK").asJava
 
   private val leakedAppTags = new java.util.concurrent.ConcurrentHashMap[String, Long]()
 
+  private[utils] val appList = new ListBuffer[SparkYarnApp]()
```
jerryshao: Should this list support concurrent modification?

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -98,7 +112,33 @@ object SparkYarnApp extends Logging {
     }
   }
 
-
+  private val yarnAppMonitorThread = new Thread() {
+    override def run() : Unit = {
+      val executor = Executors.newSingleThreadExecutor
+      while (true) {
+        for (app <- appList) {
+          val handler = executor.submit(new Runnable {
```
jerryshao: I think this can be handled by Scala `Future`.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -122,31 +162,53 @@ class SparkYarnApp private[utils] (
...
+      return
+    }
+
+    process.foreach(_.destroy())
+
+    if (!appId.isEmpty) {
+      yarnClient.killApplication(appId.get)
+      return
+    } else {
+      Future {
```
jerryshao: What is the purpose of using `Future`?

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -165,40 +227,24 @@ class SparkYarnApp private[utils] (
...
+        case Some(app) => app.getApplicationId
+        case None =>
           throw new IllegalStateException(s"No YARN application is found with tag" +
             s" $appTagLowerCase in ${livyConf.getTimeAsMs(LivyConf.YARN_APP_LOOKUP_TIMEOUT)/1000}" +
             " seconds. This may be because 1) spark-submit fail to submit application to YARN; " +
             "or 2) YARN cluster doesn't have enough resources to start the application in time. " +
             "Please check Livy log and YARN log to know the details.")
-        } else {
-          Clock.sleep(pollInterval.toMillis)
-          getAppIdFromTag(appTagLowerCase, pollInterval, deadline)
```
jerryshao: Why do we remove this code?

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
runzhiwang: Because the code in `Future` may cost a lot of time.  Use `Future` can avoid blocking the main thread.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
runzhiwang: Because this code call `getAppIdFromTag`  recursively until successful or `deadline.isOverdue`, which maybe cost a lot of time, and delay to monitor other app.

conf/livy.conf.template
runzhiwang: Yes, I will revert to `5s`.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -241,50 +291,37 @@ class SparkYarnApp private[utils] (
...
-        appIdOption.map(ConverterUtils.toApplicationId).getOrElse {
-          val pollInterval = getYarnPollInterval(livyConf)
-          val deadline = getYarnTagToAppIdTimeout(livyConf).fromNow
-          getAppIdFromTag(appTag, pollInterval, deadline)
-        }
-      } catch {
-        case e: Exception =>
-          appIdPromise.failure(e)
-          throw e
+      if(appId .isEmpty) {
```
jahstreet:
```suggestion
      if(appId.isEmpty) {
```

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -313,13 +350,19 @@ class SparkYarnApp private[utils] (
...
-        changeState(SparkApp.State.KILLED)
+      // throw InterruptedException when monitor timeout
+      // throw IllegalStateException when getAppId failed
+      case e @ (_: InterruptedException | _: IllegalStateException) =>
+        yarnAppMonitorFailedTimes += 1
+        if (yarnAppMonitorFailedTimes > yarnAppMonitorMaxFailedTimes) {
+          error(s"Exception whiling monitor $appTag", e)
+          yarnDiagnostics = ArrayBuffer(e.getMessage)
+          failToMonitor()
+        }
```
jahstreet: This behaviour seems a bit strange to me. You do not reset `yarnAppMonitorFailedTimes` ever. I would rather set a value of `yarnAppMonitorMaxFailedTimes` in a row and treat it accordingly.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -98,7 +116,35 @@ object SparkYarnApp extends Logging {
...
+        for ((app, appTag) <- appMap) {
+          Future {
+            val handler = executor.submit(new Runnable {
+              override def run(): Unit = {
+                app.monitorSparkYarnApp()
+              }
+            })
+            try {
+              // prevent the rpc block of one app from blocking all apps
+              handler.get(yarnAppMonitorTimeout, MILLISECONDS)
```
jahstreet: Actually you block here for `yarnAppMonitorTimeout` anyway, if Future is not completed

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -165,40 +231,24 @@ class SparkYarnApp private[utils] (
...
-    // FIXME Should not loop thru all YARN applications but YarnClient doesn't offer an API.
-    // Consider calling rmClient in YarnClient directly.
-    yarnClient.getApplications(appType).asScala.find(_.getApplicationTags.contains(appTagLowerCase))
-    match {
-      case Some(app) => app.getApplicationId
-      case None =>
-        if (deadline.isOverdue) {
-          process.foreach(_.destroy())
-          leakedAppTags.put(appTag, System.currentTimeMillis())
+  private def getAppId(): ApplicationId = {
```
jahstreet: I would prefer making that method to return Option[ApplicationId] rather than throwing the exception. The handling of the returning `None` can be moved to the places of calling it in that case. WDYT?

jahstreet (commented):
Nice PR proposal, it should have been done long time ago) Added some comments. I see that improvements that can be done here with rearranging code a bit. Would you mind if I propose some changes on to your branch?
Suggestions: https://github.com/runzhiwang/incubator-livy/pull/1

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -98,7 +116,35 @@ object SparkYarnApp extends Logging {
     }
   }
 
-
+  private val yarnAppMonitorThread = new Thread() {
+    override def run() : Unit = {
+      val executor = Executors.newSingleThreadExecutor
+      while (true) {
+        for ((app, appTag) <- appMap) {
+          Future {
```
jahstreet: What about this Future completion? You create it each `yarnPollInterval`. Can it be better to wait until this future is completed and only then call Thread.sleep, wdyt?

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
jahstreet: Since you submit app.monitor() to executor anyway it doesn't block other apps monitoring.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -134,19 +183,36 @@ class SparkYarnApp private[utils] (
...
-      } finally {
-        process.foreach(_.destroy())
+
+    if (!isRunning) {
+      return
+    }
+
+    process.foreach(_.destroy())
+
+    if (!appId.isEmpty) {
```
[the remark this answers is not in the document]
runzhiwang: Hi, `!isEmpty` equals to `isDefined`.
image

runzhiwang:
> Nice PR proposal, it should have been done long time ago) Added some comments. I see that improvements that can be done here with rearranging code a bit. Would you mind if I propose some changes on to your branch?
> Suggestions: [runzhiwang#1](https://github.com/runzhiwang/incubator-livy/pull/1)

Thank you for your review, and please propose your changes to my branch.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
runzhiwang: ` if (yarnAppMonitorFailedTimes > yarnAppMonitorMaxFailedTimes)`, then the app will be deleted, there is no need to reset `yarnAppMonitorFailedTimes`

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
runzhiwang: I will change it.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -134,19 +183,36 @@ class SparkYarnApp private[utils] (
...
+      yarnClient.killApplication(appId.get)
+      return
```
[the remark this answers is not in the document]
runzhiwang: I will change it.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
runzhiwang: There is no need to call ` getAppIdFromTag` recursively here. Because `Future` was implemented by thread pool, if all the threads in the poll cost a lot of time to  ` getAppIdFromTag`, the monitor of apps which were not monitored by thread pool will also be delayed.

conf/livy.conf.template
```diff
@@ -121,12 +121,18 @@
...
 # livy.server.yarn.app-lookup-timeout = 120s
+# How long to find the yarn app
+# livy.server.yarn.app-lookup-interval = 1s
 # When the cluster is busy, we may fail to launch yarn app in app-lookup-timeout, then it would
 # cause session leakage, so we need to check session leakage.
 # How long to check livy session leakage
 # livy.server.yarn.app-leakage.check-timeout = 600s
 # how often to check livy session leakage
 # livy.server.yarn.app-leakage.check-interval = 60s
+# how long to monitor the yarn app
```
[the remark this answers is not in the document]
runzhiwang: I will change it

conf/livy.conf.template
```diff
@@ -121,12 +121,18 @@
...
+# livy.server.yarn.app-monitor.timeout = 10000ms
```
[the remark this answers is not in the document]
runzhiwang: Is it necessary to change it ?

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
runzhiwang: There are two places of `getAppId`, and the contents of exception is too long. If throw so long exception in two places, the code looks redundant.

runzhiwang: Yes, I will change it.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
runzhiwang: Yes, there is a risk of concurrent calls of monitorSparkYarnApp(). And is it better to use `private def monitorSparkYarnApp(): Unit = synchronized { ... } ` ?

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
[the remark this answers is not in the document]
jahstreet: Yeap! This is the idea indeed) this comment is for readability improvement only.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
jahstreet: My proposal is just to block futures before calling sleep. synchronized may work as well, up to you

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
jahstreet: Yeah, I see now. Depends on how do we want to handle it indeed.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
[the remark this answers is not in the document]
runzhiwang: Yes, you are right.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
runzhiwang: I'm not sure it's better.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
runzhiwang: I have changed.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
runzhiwang: I have changed.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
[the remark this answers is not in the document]
runzhiwang: I have changed.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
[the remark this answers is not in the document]
runzhiwang: I have changed.

server/src/main/scala/org/apache/livy/LivyConf.scala
```diff
@@ -212,6 +214,11 @@ object LivyConf {
   val YARN_APP_LEAKAGE_CHECK_TIMEOUT = Entry("livy.server.yarn.app-leakage.check-timeout", "600s")
   // how often to check livy session leakage
   val YARN_APP_LEAKAGE_CHECK_INTERVAL = Entry("livy.server.yarn.app-leakage.check-interval", "60s")
+  // how long to monitor the yarn app
+  val YARN_APP_MONITOR_TIMEOUT = Entry("livy.server.yarn.app-monitor.timeout", "10000ms")
```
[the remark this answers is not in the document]
runzhiwang: I have changed.

conf/livy.conf.template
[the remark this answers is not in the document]
runzhiwang: I have changed.

conf/livy.conf.template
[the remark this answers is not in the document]
runzhiwang: I have changed.

runzhiwang:
@jahstreet Thank you for your comments. I have updated the PR.

jahstreet:
@runzhiwang Yeah, I see, will go through once again during weekend. Good job man!

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -98,7 +119,37 @@ object SparkYarnApp extends Logging {
...
+                  error(s"monitor app: ${appTag} exception:", e)
+                }
+              }
+              if (!app.isRunning) {
+                appMap -= app
+              }
+            }
+          }
+        }
+        Thread.sleep(yarnPollInterval)
```
jerryshao: We should consider interrupted exception here. It would be better to catch this exception and exit the infinite loop.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -56,17 +67,27 @@ object SparkYarnApp extends Logging {
...
-  private def getYarnPollInterval(livyConf: LivyConf): FiniteDuration =
-    livyConf.getTimeAsMs(LivyConf.YARN_POLL_INTERVAL) milliseconds
+  private def getYarnTagToAppIdMaxFailedTimes(livyConf: LivyConf): Int =
+    livyConf.getTimeAsMs(LivyConf.YARN_APP_LOOKUP_TIMEOUT).toInt / yarnAppLookUpInterval.toInt
 
   private val appType = Set("SPARK").asJava
 
   private val leakedAppTags = new java.util.concurrent.ConcurrentHashMap[String, Long]()
 
+  private[utils] val appMap = new TrieMap[SparkYarnApp, String]()
```
jerryshao: Why don't you choose map which supports concurrent modification?

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
jerryshao: It's a bit complicated to have executor inside the future (thread inside another thread), is it necessary, can it be simplified?

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
jerryshao: `Future` is backed by a thread pool, this will increase the overhead, also changes the sync semantics to async (potentially introduces some threading issues).

jerryshao:
I still need a close look to see the detailed implementations.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
runzhiwang: @jerryshao  `TrieMap ` also supports concurrent modification.
image
image

jerryshao: If this support concurrency, we do we need a lock below?

runzhiwang: If without lock, there maybe exists two threads monitor one app at the same time.

jerryshao: How this could be happened, can you please explain more. What I saw is that one app per future.

runzhiwang: For example, there is only one appA,  and there are 8 threads in the thread pool of future, and the config `yarnAppMonitorTimeout`  is bigger than  `yarnPollInterval `.  In the first `while (true) ` loop, thread1 monitors appA whose cost is longer than `yarnPollInterval`, and after `yarnPollInterval `the second  `while (true) ` loop starts, thread2 begin to monitor appA. Thus thread1 and thread2 monitor appA at the same time. Though I can config `yarnAppMonitorTimeout <  yarnPollInterval ` to avoid this problem by ending the monitor of thread1 before the monitor of thread2, but I think it is not safe in case someone changes the config.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
runzhiwang: I have updated

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -98,7 +118,36 @@ object SparkYarnApp extends Logging {
...
+
+  for (i <- 0 until yarnAppMonitorThreadPoolSize) {
+    yarnAppMonitorThreadPool.execute(new Runnable {
+      override def run(): Unit = {
+        var loop = true
+        try {
+          while (loop) {
+            val app = appQueue.poll()
+            if (app != null) {
+              app.monitorSparkYarnApp()
```
jerryshao: Will this method `monitorSparkYarnApp` be blocked indefinitely?

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -98,7 +118,36 @@ object SparkYarnApp extends Logging {
     }
   }
 
+  val yarnAppMonitorThreadPool: ExecutorService =
+    Executors.newFixedThreadPool(yarnAppMonitorThreadPoolSize)
+
+  for (i <- 0 until yarnAppMonitorThreadPoolSize) {
+    yarnAppMonitorThreadPool.execute(new Runnable {
```
jerryshao: Please create a name class for this `Runnable` logic and use a created object here.

jerryshao:
I'm OK with the current changes, @yiheng would you please also take a look.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -98,7 +118,36 @@ object SparkYarnApp extends Logging {
...
+            }
+
+            if (app.isRunning) {
+              appQueue.add(app)
+            }
+
+            Thread.sleep(yarnPollInterval)
+          }
+        } catch {
+          case _: InterruptedException => loop = false
```
yiheng: Looks like this line is already outside of the while loop. So what's the meaning set loop to false?

runzhiwang:
@yiheng Thanks, you are right. I have not completed the modification, please wait.

jerryshao:
Looks ok from my side, @yiheng please help to review again.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -311,15 +322,22 @@ class SparkYarnApp private[utils] (
...
+
       debug(s"$appId $state ${yarnDiagnostics.mkString(" ")}")
     } catch {
-      case _: InterruptedException =>
-        yarnDiagnostics = ArrayBuffer("Session stopped by user.")
-        changeState(SparkApp.State.KILLED)
+      // throw IllegalStateException when getAppId failed
+      case e: IllegalStateException =>
+        yarnTagToAppIdFailedTimes += 1
+        if (yarnTagToAppIdFailedTimes > getYarnTagToAppIdMaxFailedTimes(livyConf)) {
```
yiheng: why not read the value into a class varaible first?

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
yiheng: I agree with @jahstreet I think throw IllegalStateException and catch it outside is not necessary. Return an Option and handle the None in the outside while loop is more straight forward.

runzhiwang: @yiheng @jahstreet updated.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
runzhiwang: @yiheng updated

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -254,50 +268,45 @@ class SparkYarnApp private[utils] (
...
       // If appId is not known, query YARN by appTag to get it.
-      val appId = try {
-        appIdOption.map(ConverterUtils.toApplicationId).getOrElse {
-          val pollInterval = getYarnPollInterval(livyConf)
-          val deadline = getYarnTagToAppIdTimeout(livyConf).fromNow
-          getAppIdFromTag(appTag, pollInterval, deadline)
+      if (appId.isEmpty) {
+        appId = getAppId()
+        if (appId.isEmpty) {
+          throw new IllegalStateException(s"No YARN application is found with tag " +
```
yiheng: instead of throwing an exception, can we move the exception handle code here?

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -324,15 +333,22 @@ class SparkYarnApp private[utils] (
...
       }
 
+      yarnTagToAppIdFailedTimes = 0
+
       debug(s"$appId $state ${yarnDiagnostics.mkString(" ")}")
     } catch {
-      case _: InterruptedException =>
-        yarnDiagnostics = ArrayBuffer("Session stopped by user.")
-        changeState(SparkApp.State.KILLED)
+      // throw IllegalStateException when getAppId failed
```
yiheng: why remove the interruptedException handle code？

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
runzhiwang: @jerryshao @yiheng I add `checkMonitorAppTimeoutThread ` to check whether `monitorSparkYarnApp ` was blocked.

server/src/main/scala/org/apache/livy/LivyConf.scala
```diff
@@ -216,11 +213,27 @@ object LivyConf {
...
+  // How often to check monitor thread block
+  val YARN_APP_MONITOR_THREAD_BLOCK_CHECK_INTERVAL =
+    Entry("livy.server.yarn.app-monitor.thread-block.check-interval", "10s")
+  // If some thread cost more than this config to monitor one app,
+  // stop the monitor of the app which considered blocked.
+  val YARN_APP_MONITOR_THREAD_BLOCK_TIMEOUT =
+    Entry("livy.server.yarn.app-monitor.thread-block-timeout", "60s")
+  // If Livy can't monitor the yarn app successfully within this max times, consider the app failed.
+  val YARN_APP_MONITOR_MAX_FAILED_TIMES =
+    Entry("livy.server.yarn.app-monitor.max-failed.times", 12)
```
jerryshao: Can you please simplify the configurations, there're some many related configurations, which makes user hard to configure.

server/src/main/scala/org/apache/livy/LivyConf.scala
```diff
@@ -202,11 +202,8 @@ object LivyConf {
   // Livy will cache the max no of logs specified. 0 means don't cache the logs.
   val SPARK_LOGS_SIZE = Entry("livy.cache-log.size", 200)
 
-  // If Livy can't find the yarn app within this time, consider it lost.
-  val YARN_APP_LOOKUP_TIMEOUT = Entry("livy.server.yarn.app-lookup-timeout", "120s")
-
-  // How often Livy polls YARN to refresh YARN app state.
-  val YARN_POLL_INTERVAL = Entry("livy.server.yarn.poll-interval", "5s")
```
jerryshao: The removal of these two configurations will break the backward compatibility.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -56,20 +69,55 @@ object SparkYarnApp extends Logging {
...
+            if (now - updatedTime - yarnAppMonitorThreadInterval >
+              yarnAppMonitorThreadBlockTimeout) {
+              thread.interrupt()
+            }
+          }
+
+          Thread.sleep(yarnAppMonitorThreadBlockCheckInterval)
+        } catch {
+          case e: InterruptedException =>
+            error(s"checkMonitorAppTimeoutThread Exception whiling monitor", e)
```
jerryshao: How do you exit the thread if you catch the `InterruptedException` in the infinite loop?

runzhiwang: @jerryshao updated

server/src/main/scala/org/apache/livy/LivyConf.scala
runzhiwang: @jerryshao updated

server/src/main/scala/org/apache/livy/LivyConf.scala
runzhiwang: @jerryshao updated

server/src/main/scala/org/apache/livy/LivyConf.scala
```diff
@@ -205,6 +205,13 @@ object LivyConf {
   // If Livy can't find the yarn app within this time, consider it lost.
   val YARN_APP_LOOKUP_TIMEOUT = Entry("livy.server.yarn.app-lookup-timeout", "120s")
 
+  // If Livy can't find the yarn app within this max times, consider it lost.
+  val YARN_APP_LOOKUP_MAX_FAILED_TIMES = Entry("livy.server.yarn.app-lookup.max-failed.times", 120)
```
jerryshao: Here the default value is 120, but in the conf file it is just 30.

runzhiwang: @jerryshao updated, 30 times is enough, because sleep 5 seconds each time, so the total time is bigger than 150 seconds.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -254,50 +311,58 @@ class SparkYarnApp private[utils] (
...
+    changeState(SparkApp.State.FAILED)
+    process.foreach(_.destroy())
+    leakedAppTags.put(appTag, System.currentTimeMillis())
+  }
+
+  private def failToGetAppId(): Unit = {
+    yarnTagToAppIdFailedTimes += 1
+    if (yarnTagToAppIdFailedTimes > yarnAppLookupMaxFailedTimes) {
+      val msg = "No YARN application is found with tag " +
+        "$appTagLowerCase. This may be because " +
```
jahstreet:
```suggestion
        s"${appTag.toLowerCase}. This may be because " +
```

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -254,50 +311,58 @@ class SparkYarnApp private[utils] (
...
+        "1) spark-submit fail to submit application to YARN; " +
+        "or 2) YARN cluster doesn't have enough resource to start the application in time. " +
+        "Please check Livy log and YARN log to know the details."
+
+      error(s"Exception whiling monitor $appTag " + msg)
```
jahstreet:
```suggestion
      error(s"Failed monitoring the app $appTag: $msg")
```

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -324,15 +389,22 @@ class SparkYarnApp private[utils] (
...
+      case e: InterruptedException =>
+        yarnAppMonitorFailedTimes += 1
+        if (yarnAppMonitorFailedTimes > yarnAppLookupMaxFailedTimes) {
+          error(s"Exception whiling monitor $appTag", e)
+          yarnDiagnostics = ArrayBuffer(e.getMessage)
+          failToMonitor()
+        }
       case NonFatal(e) =>
-        error(s"Error whiling refreshing YARN state", e)
+        error(s"Error whiling monitor $appTag", e)
```
jahstreet:
```suggestion
        error(s"Failed monitoring the app $appTag.", e)
```

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -324,15 +389,22 @@ class SparkYarnApp private[utils] (
...
+
       debug(s"$appId $state ${yarnDiagnostics.mkString(" ")}")
     } catch {
-      case _: InterruptedException =>
-        yarnDiagnostics = ArrayBuffer("Session stopped by user.")
-        changeState(SparkApp.State.KILLED)
+      case e: InterruptedException =>
+        yarnAppMonitorFailedTimes += 1
+        if (yarnAppMonitorFailedTimes > yarnAppLookupMaxFailedTimes) {
+          error(s"Exception whiling monitor $appTag", e)
```
jahstreet:
```suggestion
          error(s"Monitoring of the app $appTag was interrupted.", e)
```

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -111,7 +152,48 @@ object SparkYarnApp extends Logging {
...
+
+  private def initYarnAppMonitorThreadPool(livyConf: LivyConf): Unit = {
+    val poolSize = livyConf.getInt(LivyConf.YARN_APP_LOOKUP_THREAD_POOL_SIZE)
+    val yarnAppMonitorThreadPool: ExecutorService =
+      Executors.newFixedThreadPool(poolSize)
+
+    val runnable = new YarnAppMonitorRunnable()
+    for (i <- 0 until poolSize) {
+      yarnAppMonitorThreadPool.execute(runnable)
+    }
```
jahstreet: Can you explain your idea here? Why do we need to execute the same runnable in a loop?

jahstreet (commented):
Added some comments, other than that looks nice so far, good job!

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
runzhiwang: @jahstreet Hi,Because we start `poolSize` long running threads to monitor all the app, each execute start one thread.

jahstreet: Oh, I see, checked `YarnAppMonitorRunnable`, my bad)

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -111,7 +152,48 @@ object SparkYarnApp extends Logging {
...
+  }
+
+  private def initYarnAppMonitorThreadPool(livyConf: LivyConf): Unit = {
+    val poolSize = livyConf.getInt(LivyConf.YARN_APP_LOOKUP_THREAD_POOL_SIZE)
+    val yarnAppMonitorThreadPool: ExecutorService =
+      Executors.newFixedThreadPool(poolSize)
+
+    val runnable = new YarnAppMonitorRunnable()
+    for (i <- 0 until poolSize) {
```
jahstreet:
```suggestion
    for (_ <- 0 until poolSize) {
```
1 more nit

runzhiwang:
@jahstreet Thank you very much, I have updated according to your comments.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
runzhiwang: updated

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
runzhiwang: updated

runzhiwang: I have revert it

jahstreet:
@runzhiwang can you test your patch on a Yarn cluster?

runzhiwang:
@jahstreet OK.

runzhiwang:
@jahstreet Hi, I have test the patch, and it works fine.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -56,20 +64,53 @@ object SparkYarnApp extends Logging {
...
+            if (now - updatedTime - yarnPoolInterval >
+              yarnAppLookupTimeout) {
+              thread.interrupt()
+            }
+          }
+
+          Thread.sleep(yarnPoolInterval)
+        } catch {
+          case e: InterruptedException =>
+            loop = false
```
yiheng: Should we keep in the loop? I'm afraid the thread is interrupted while the livy is still running. Then all the app status won't change.

sanchay0:
Hey @runzhiwang what's the status of this PR? Are you going to merge this in? I'm having issues with dangling threads with a lot of jobs running simultaneously and this fix will be great to have!

runzhiwang:
@sanchay0 Thanks for your attention. I have complete the code. And It just need more review.

PR #245, opened 2019-10-09
[LIVY-695] Upgrade JQuery to 3.4.1 and Bootstrap to 3.4.1

peter-toth:

This PR:
- updates JQuery to 3.4.1
- updates Bootstrap to 3.4.1
- switches to minified version of Bootstrap
- removes unused old bootstrap files

The changes are needed to fix CVEs.

## How was this patch tested?

Manual tests:
- Livy Server UI looks ok
- Documentation had a minor issue of logo position after plain upgrade, but fix is included in this commit


mgaido91:
cc @ajbozarth

ajbozarth:
Is there a reason you added 8 new bootstrap files? When downloading bootstrap it allows you to pick and choose which files you need. When we created the UI we made sure to only include the bootstrap files we use in order to minimize the size of the UI static files directory. So if you only included them because they came with the default download you should instead only update the ones we already had. 

The rest of the changes LGTM

peter-toth:
Thanks @ajbozarth for the review. I will remove the unused files then.

peter-toth:
The bootstrap instance in 'docs' looked like a 'full' bootstrap (there were both normal and minified js, included bootstrap-theme files without any reference, ...) that's why I added the full 3.4.1 first.

Now I modified my PR to update necessary new files only and also removed unused old ones.

ajbozarth:
Ah, I see what you mean. The docs should actually be treated as separate from the rest of the Livy server files then. They are designed to be built in conjunction with the livy website and are based on an apache template ( https://github.com/apache/apache-website-template ). I would suggest either leaving the docs alone or updating them in a separate PR tied with a PR on the livy website repo ( https://github.com/apache/incubator-livy-website ) with a matching update. 

What exactly is the CVE reasoning behind the update to begin with? If its worth updating apache website code it should be raised across all the projects which utilize the website template. If it's not then just updating our server version should be fine.

ajbozarth:
A clarification on the doc relation to the website. The themes dir your updating was copied into the main repo from the website repo so users could build and see local copies of the docs without cloning the website repo. Whereas on a release we copy a build on the docs into the website repo, but that copy would then reference the themes dir in the website repo. Therefore we need to keep those dirs completely in sync.

peter-toth:
Thanks @ajbozarth for the clarification, I've reverted all `docs` related changes.
Blackduck reported  CVE-2019-11358 regarding current JQuery and CVE-2018-14040, CVE-2018-20676, CVE-2016-10735, CVE-2019-8331, CVE-2018-14042 regarding current Bootstrap.

peter-toth:
Thanks for the review @ajbozarth. Do you know if https://issues.apache.org/jira/browse/LIVY-695 will be updated automatically? It is still open and the assignee isn't me.

ajbozarth:
thanks for pointing the out @peter-toth I'll fix that, the merge script has been failing the jira updates the last couple times I used it

PR #247, opened 2019-10-21
[LIVY-699][THRIFT] Fix resultSet.getBigDecimal throw java.sql.SQLException: Illegal conversion

runzhiwang:

[LIVY-699][THRIFT] Fix resultSet.getBigDecimal throw java.sql.SQLException: Illegal conversion. 

1. `create table test(id decimal)`.
2. Then `resultSet.getBigDecimal(1)` will throw:` java.sql.SQLException: Illegal conversion`. The reason is `getSchema().getColumnDescriptorAt(columnIndex - 1).getType();` at https://github.com/apache/hive/blob/master/jdbc/src/java/org/apache/hive/jdbc/HiveBaseResultSet.java#L415 return string, so cannot pass the check `val instanceof BigDecimal `at https://github.com/apache/hive/blob/master/jdbc/src/java/org/apache/hive/jdbc/HiveBaseResultSet.java#L133, so throw `java.sql.SQLException: Illegal conversion` at https://github.com/apache/hive/blob/master/jdbc/src/java/org/apache/hive/jdbc/HiveBaseResultSet.java#L137
3. So the root cause is the error return of `getType()`, which should return decimal other than string.
4. Regarding to date and timestamp, though the return type is string, hive-jdbc has done the transformation from string to date and timestamp in the following links. But I think it is necessary to return the right type.
https://github.com/apache/hive/blob/master/jdbc/src/java/org/apache/hive/jdbc/HiveBaseResultSet.java#L255
https://github.com/apache/hive/blob/master/jdbc/src/java/org/apache/hive/jdbc/HiveBaseResultSet.java#L571

Additionally, SparkThrift return decimal type instead of string in the same case, if user use `getBigDecimal` and migrate from SparkThrift to Livy, it will throw exception. So it is necessary to return decimal instead of string in livy. 

## How was this patch tested?

Add `SchemaIT.scala` to test the return of column type.


yiheng:
Please add test cases to cover these scenarios. Others look good to me.

runzhiwang:
@yiheng test cases has been added.

runzhiwang:
@mgaido91 Could you help review this pr ?

mgaido91:
I am not sure about the goal of this PR. May you please describe it? In general these types are supported, but they are serialized and treated as strings... I don't see that being changed: may you please explain which problem you had and why you decided to do this change?

runzhiwang:
@mgaido91  If `create table test(id decimal)`,  and then `resultSet.getMetaData().getColumnTypeName(1) `will print `string` other than `decimal`. The same to `timestamp `and `date`.

runzhiwang:
@mgaido91 Hi, thank you for the suggestions. 
1. I have improve this PR description.
2.This configuration flag maybe is not necessary. Even with this PR, getString() can still get the value of decimal, and the api is compatible.

mgaido91:
@runzhiwang 
what I meant with

> This configuration flag maybe is not necessary. Even with this PR, getString() can still get the value of decimal, and the api is compatible.

was that you might have some code which does something like:

```
resultSet.getMetaData().getColumnTypeName(1) match {
  case "string" => resultSet.getString(1)
  ...
}
```

and it would be broken by this change. So I think we need the flag.

Please in the PR description, you should not just state *this is a problem*, you should say *this is a problem because ....*. Please also format code there. Moreover, please in the title write `[THRIFT]`, all uppercase instead of the current.
Thanks.

runzhiwang:
@mgaido91 Hi, I have updated the PR description. But I do not think user will use` switch case` to get the value. Besides the flag seems wired.

mgaido91:
@runzhiwang regarding the description of the PR: I don0t see how point 3 is relevant for this PR The behavior would be the same in that case, wouldn't it?

> I think the original behavior is kind of bug(e.g. return decimal type as string).

I think this is arguable: what does Livy return for that column? A string. What does that column contain? A decimal. Hence both answers are correct somehow: one refers to what is returned to the client, the other to how data is stored in the table/backend. For this reason, I think having a flag which lets the user customize/choose the more appropriate behavior to him is the best option.

runzhiwang:
@mgaido91 Hi, I have updated the PR. But the flag seems not easy to add, let me have a think.

jerryshao:
Is this a bug or a behavior change?

runzhiwang:
> Is this a bug or a behavior change?

@jerryshao It's a bug.

jerryshao:
Is this BigDecimal an exception to all other types, or we still have other types which have the similar behaviors? I think we should deliver a clear and consistent way to the users.

mgaido91:
> @jerryshao It's a bug.

I don't agree on this. You are stating that the current behavior is wrong but I don't see a clear evidence of this. Please see https://github.com/apache/incubator-livy/pull/247#issuecomment-547007658

runzhiwang:
@mgaido91 Hi, It's a bug because resultSet.getBigDecimal throw java.sql.SQLException: Illegal conversion in livy, which work normally in SparkThrift. And the updated PR description has more details about this.

mgaido91:
ok, now I see, thanks for explaining @runzhiwang. ThenI agree it is a bug. May you please update the UT or create a new one with the case you mentioned? ie. using `getBigDecimal` it works correctly. Thanks.

runzhiwang:
@mgaido91 Hi, I have update the IT to test `getBigDecimal` at https://github.com/apache/incubator-livy/pull/247/commits/047c943e2413cdc40b9f7aaf3c448f3a2fca7ebd#diff-aa5d97e3ed9d52622cd597fa7c140ff0R134.  I also add IT for get value of all other types, such as  `getByte`, `getShort`, to make the IT stronger.

integration-test/src/thriftserver/test/scala/org/apache/livy/test/SchemaIT.scala
```diff
@@ -0,0 +1,151 @@
...
+package org.apache.livy.test
+
+import java.math.BigDecimal
+import java.sql.Timestamp
+import java.text.SimpleDateFormat
+import java.util.Arrays
+
+import org.apache.livy.test.framework.BaseThriftIntegrationTestSuite
+
+class SchemaIT extends BaseThriftIntegrationTestSuite {
```
mgaido91: can't we have this as an UT instead of an IT?

runzhiwang: @mgaido91 Hi, the reason for IT is : 1. the bug has a strong relationship with hive jdbc, actually the exception was threw by hive jdbc, but UT can not cover the relationship. 2. I think it's necessary to write an IT to make sure thrift can work normally, not only for testing the return type.

mgaido91: there are UTs going through JDBC. Please see https://github.com/apache/incubator-livy/blob/master/thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala. As you can see there are also already datatypes tests there. So you might want to integrate them.

runzhiwang: @mgaido91 Hi, thank you for comments. I have integrate ThriftServerSuites.

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
```diff
@@ -34,22 +34,75 @@ trait CommonThriftTests {
...
+
+    assert(resultSet.getDate(14).
+      compareTo(Date.valueOf("2018-08-06")) == 0)
+    assert(rsMetaData.getColumnTypeName(14) == "date")
+
     assert(!resultSet.next())
 
     val resultSetWithNulls = statement.executeQuery("select cast(null as string), " +
-      "cast(null as decimal), cast(null as double), cast(null as date), null")
+      "cast(null as decimal), cast(null as double), cast(null as date), null, cast(null as int)")
```
mgaido91: not really related to this, but for the sake of completeness, may you please add all the datatypes here too, in order to test the null for them all?

runzhiwang:
@mgaido91 Hi, build has passed.

mgaido91:
Thanks, merged to master.

PR #253, opened 2019-11-01
[LIVY-705][THRIFT] Support getting keystore password from Hadoop credential provider

wypoon:

https://issues.apache.org/jira/browse/LIVY-705

LIVY-475 added support for getting the keystore password and key password from a Hadoop credential provider file. The keystore password is also needed for SSL/TLS support in the Thrift server. In this change, we extend the support for getting the keystore password from the Hadoop credential provider to the Thrift server as well.

## How was this patch tested?

Manually tested a Livy Thrift server that has livy.server.thrift.use.SSL=true, using both binary and http mode. Configured keystore password in a Hadoop credential provider file and provided the path to this file in livy.hadoop.security.credential.provider.path.

### DIFF
--- a/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/cli/ThriftCLIService.scala
+++ b/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/cli/ThriftCLIService.scala
@@ -26,6 +26,7 @@ import javax.security.auth.login.LoginException
 import scala.collection.JavaConverters._
 
 import com.google.common.base.Preconditions.checkArgument
+import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.security.UserGroupInformation
 import org.apache.hadoop.security.authentication.util.KerberosName
 import org.apache.hadoop.security.authorize.ProxyUsers
@@ -81,6 +82,22 @@ abstract class ThriftCLIService(val cliService: LivyCLIService, val serviceName:
     minWorkerThreads = livyConf.getInt(LivyConf.THRIFT_MIN_WORKER_THREADS)
     maxWorkerThreads = livyConf.getInt(LivyConf.THRIFT_MAX_WORKER_THREADS)
     super.init(livyConf)
+  }
+
+  protected def getKeyStorePassword(): String = {
+    val credentialProviderPath = livyConf.get(LivyConf.HADOOP_CREDENTIAL_PROVIDER_PATH)
+    val hadoopConf = new Configuration()
+    if (credentialProviderPath != null) {
+      hadoopConf.set("hadoop.security.credential.provider.path", credentialProviderPath)
+    }
+    Option(livyConf.get(LivyConf.SSL_KEYSTORE_PASSWORD))
+      .orElse {
+        Option(hadoopConf.getPassword(LivyConf.SSL_KEYSTORE_PASSWORD.key)).map(_.mkString)
+      }
+      .getOrElse {
+        throw new IllegalArgumentException(
+          "Livy keystore password not configured for SSL connection")
+      }
   }
 
   protected def initServer(): Unit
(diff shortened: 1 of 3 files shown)


wypoon:
@mgaido91 @jerryshao could you please review?

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/cli/ThriftCLIService.scala
```diff
@@ -83,6 +84,22 @@ abstract class ThriftCLIService(val cliService: LivyCLIService, val serviceName:
     super.init(livyConf)
   }
 
+  protected def getKeyStorePassword(): String = {
+    val credentialProviderPath = livyConf.get(LivyConf.HADOOP_CREDENTIAL_PROVIDER_PATH)
```
mgaido91: can't we move this in the `orElse`?

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/cli/ThriftCLIService.scala
```diff
@@ -83,6 +84,22 @@ abstract class ThriftCLIService(val cliService: LivyCLIService, val serviceName:
...
+    val hadoopConf = new Configuration()
+    if (credentialProviderPath != null) {
+      hadoopConf.set("hadoop.security.credential.provider.path", credentialProviderPath)
+    }
+    Option(livyConf.get(LivyConf.SSL_KEYSTORE_PASSWORD))
+      .orElse {
```
mgaido91: nit: on the line above

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/cli/ThriftCLIService.scala
```diff
@@ -83,6 +84,22 @@ abstract class ThriftCLIService(val cliService: LivyCLIService, val serviceName:
...
+        Option(hadoopConf.getPassword(LivyConf.SSL_KEYSTORE_PASSWORD.key)).map(_.mkString)
+      }
+      .getOrElse {
```
mgaido91: nit: on the line above

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/cli/ThriftCLIService.scala
```diff
@@ -83,6 +84,22 @@ abstract class ThriftCLIService(val cliService: LivyCLIService, val serviceName:
...
+        throw new IllegalArgumentException(
```
mgaido91: why don't we return just null as before here?

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/cli/ThriftCLIService.scala
wypoon: Sure, I can do that.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/cli/ThriftCLIService.scala
wypoon: I'm happy to follow the style used in the project. However, isn't it normal in Scala to put the chained method on a new line? I'm pretty sure I've seen that elsewhere in the project.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/cli/ThriftCLIService.scala
wypoon: Before I made the changes here, I found that if the keystore password was not set in the server's livy.conf, as it defaults to null, the null causes an NPE in the Thrift code that gets called subsequently, specifically in https://github.com/apache/thrift/blob/0.9.3/lib/java/src/org/apache/thrift/transport/TSSLTransportFactory.java#L199. I thought it would be better to throw an exception here up front specifying the problem. This is also consistent with throwing an IllegalArgumentException when the keystore path is not configured.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/cli/ThriftCLIService.scala
wypoon: Moved the orElse to the line above.

wypoon:
Thanks @mgaido91.

PR #254, opened 2019-11-06
[LIVY-706] ThriftServer does not actually cancel jobs if we Interrupt querys from beeline

boneanxs:

If we type `Ctrl + c` to interrupt a query from beeline, it does not cancel itself, the job still can be seen running from spark historyServer.

## How was this patch tested?

Tested in my own environment.

### DIFF
--- a/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
+++ b/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
@@ -142,6 +142,13 @@ class LivyExecuteStatementOperation(
     } catch {
       case e: Throwable =>
         val currentState = getStatus.state
+        if ((currentState == OperationState.CANCELED)
+          || (currentState == OperationState.TIMEDOUT)
+          || (currentState == OperationState.CLOSED)
+          || (currentState == OperationState.FINISHED)) {
+          warn("Ignore exception in terminal state", e)
+          return
+        }
         info(s"Error executing query, currentState $currentState, ", e)
         setState(OperationState.ERROR)
         throw new HiveSQLException(e)
@@ -177,7 +184,7 @@ class LivyExecuteStatementOperation(
       val cleaned = rpcClient.cleanupStatement(sessionHandle, statementId).get()
       if (!cleaned) {
         warn(s"Fail to cleanup query $statementId (session = ${sessionHandle.getSessionId}), " +
-          "this message can be ignored if the query failed.")
+          "this message can be ignored if the query failed or canceled.")
       }
     }
     setState(state)
--- a/thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ThriftSessionState.java
+++ b/thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ThriftSessionState.java
@@ -103,12 +103,11 @@ class ThriftSessionState {
 
   boolean cleanupStatement(String statementId) {
     checkNotNull(statementId, "No statement ID.");
+    ctx.sc().cancelJobGroup(statementId);
     if (statements.remove(statementId) == null) {
       return false;
-    } else {
-      ctx.sc().cancelJobGroup(statementId);
-      return true;
     }
+    return true;
   }
 
   void dispose() {


jerryshao:
@yiheng would you please help to review this PR? Thanks.

thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ThriftSessionState.java
```diff
@@ -103,12 +103,11 @@ StatementState findStatement(String statementId) {
 
   boolean cleanupStatement(String statementId) {
     checkNotNull(statementId, "No statement ID.");
+    ctx.sc().cancelJobGroup(statementId);
```
yiheng: Don't quite understand the change here? Why move the cancelJobGroup before the ```if else``` check? What if cancel a canceled job group?

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
```diff
@@ -142,6 +142,13 @@ class LivyExecuteStatementOperation(
     } catch {
       case e: Throwable =>
         val currentState = getStatus.state
+        if ((currentState == OperationState.CANCELED)
```
yiheng: add some test to cover these cases?

thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ThriftSessionState.java
huianyi: Sql job will register statement after it executes finished([SqlJob.java#L87](https://github.com/apache/incubator-livy/blob/9042ff5b4a03cd302a884b5a74280d4b476792b7/thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/SqlJob.java#L87)), so we can not interrput a job when it is not finished.

> What if cancel a canceled job group?

Maybe I should add try catch methods here?

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
huianyi: Sure

jerryshao:
@mgaido91 would you please help to review, thanks.

thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ThriftSessionState.java
yiheng: I see. The `if (statements.remove(statementId) == null)` has two meanings: `it is canceled` or `it is still running the spark sql job`.

I'm not sure if the method will throw an exception when it's double canceled, neither looks like spark doesn't provide a method to check if the job group has been canceled.

I feel there're still some race condition issues in the current solution.

Another solution maybe we make the prepare iterator code quickly and thread-safe. For example, we can write an iterator wrapper to make the really spark SQL execution happened when the first ResultSet.get run. And we synchronize on the statements object.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
```diff
@@ -142,6 +142,13 @@ class LivyExecuteStatementOperation(
...
+          || (currentState == OperationState.TIMEDOUT)
```
mgaido91: please indent this more

thriftserver/session/src/main/java/org/apache/livy/thriftserver/session/ThriftSessionState.java
mgaido91: I agree this is a problem, I see your point doing this, but I don't think this is the right change. What if we add the statement to the session state immediately after the query parsing and we just edit it to add the iterator later? It may happen that there is still no job scheduled and in that case nothing would happen here and the job would keep going on... We might need to introduce more checks on the execution side before the actual start of job execution and after in order to avoid weird issues...

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyExecuteStatementOperation.scala
huianyi: Ok.

PR #259, opened 2019-11-20
[LIVY-714][SERVER] Fix cannot remove the app in leakedAppTags when timeout

runzhiwang:
1.`var isRemoved = false` should be in `while(iter.hasNext),` otherwise if there are two apps, the first app will be killApplication and the second app will timeout in this loop, and after removing the first app,` isRemoved = true`, and the second app cannot pass the` if(!isRemoved)` and only will be deleted in the next loop.

2.`entry.getValue - now` is negative, and never greater than `sessionLeakageCheckTimeout`.

image

## How was this patch tested?

Existed IT and UT.

### DIFF
--- a/server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
+++ b/server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
@@ -37,13 +37,16 @@ import org.apache.livy.{LivyConf, Logging, Utils}
 
 object SparkYarnApp extends Logging {
 
-  def init(livyConf: LivyConf): Unit = {
+  def init(livyConf: LivyConf, client: Option[YarnClient] = None): Unit = {
+    mockYarnClient = client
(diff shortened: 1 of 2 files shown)


server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -73,20 +73,23 @@ object SparkYarnApp extends Logging {
...
               .foreach({ e =>
                 info(s"Kill leaked app ${e.getApplicationId}")
                 yarnClient.killApplication(e.getApplicationId)
                 iter.remove()
                 isRemoved = true
               })
+
             if (!isRemoved) {
-              if ((entry.getValue - now) > sessionLeakageCheckTimeout) {
+              if ((now - entry.getValue) > sessionLeakageCheckTimeout) {
```
yiheng: A unit test for this?

jerryshao:
Is it still WIP? @runzhiwang

runzhiwang:
@jerryshao It's not WIP now. I have change the title.

PR #261, opened 2019-11-20
[LIVY-715][DOC] The configuration in the template is inconsistent with LivyConf.scala

captainzmc:

    When I test livy impersonation found that, in livy.conf.template the value of livy.impersonation.enabled is true. So I thought impersonation was enabled by default.             
    However, impersonation was not turned on when we test. I found that the real configuration in LivyConf. scala is false. 
    This can mislead users.

## How was this patch tested?

no need

### DIFF
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -62,7 +62,7 @@
 # livy.server.session.state-retain.sec = 600s
 
 # If livy should impersonate the requesting users when creating a new session.
-# livy.impersonation.enabled = true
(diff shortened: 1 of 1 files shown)


captainzmc:
@jerryshao Just change the doc to avoid misleading the user. Could please help review it？

yiheng:
Yes. This template configuration is conflicted with this code
https://github.com/apache/incubator-livy/blob/master/server/src/main/scala/org/apache/livy/LivyConf.scala#L72

+1

PR #266, opened 2019-12-18
[LIVY-729] Fix livy recover the killed session

runzhiwang:

1. Set livy.server.recovery.mode=recovery, and create an interactive session: session0 in yarn-cluster
2. kill the yarn application of the session
3. restart livy
4. livy try to recover session0, but application has been killed and driver does not exist, so client can not connect to driver, and exception was thrown as the image.
5. If the ip:port of the driver was reused by session1, client of session0 will try to connect to driver of session1, then driver will throw exception: Unexpected client ID.
6. Both the exception threw by livy and driver will confuse the user, and recover a lot of killed sessions will delay the recover of alive session.
image

How to fix:
When session was killed or dead, remove the session from the store

## How was this patch tested?

Existed IT and UT.


runzhiwang:
@jerryshao @yiheng Could you help to review ?

yiheng:
+1

PR #267, opened 2019-12-20
[LIVY-732][SERVER] A common zookeeper wrapper utility

runzhiwang:

Currently, the utilities of zookeeper mixed with ZooKeeperStateStore. To use the utility of zookeeper, the instance of ZooKeeperStateStore has to be created , which looks weird.

This PR aims to achieve the follow target:

1.  Extract the utilities of zookeeper from ZooKeeperStateStore to support such as distributed lock, service discovery and so on.

## How was this patch tested?

Existed UT and IT.

### DIFF
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -107,6 +107,8 @@
 # Must set livy.server.recovery.state-store and livy.server.recovery.state-store.url to
 # configure the state store.
 # livy.server.recovery.mode = off
+# Zookeeper address used for HA and state store. e.g. host1:port1, host2:port2
+# livy.server.zookeeper.url =
 
 # Where Livy should store state to for recovery. Possible values:
 # <empty>: Default. State store disabled.
@@ -116,7 +118,7 @@
 
 # For filesystem state store, the path of the state store directory. Please don't use a filesystem
 # that doesn't support atomic rename (e.g. S3). e.g. file:///tmp/livy or hdfs:///.
-# For zookeeper, the address to the Zookeeper servers. e.g. host1:port1,host2:port2
+# For zookeeper, please set livy.server.zookeeper.url
 # livy.server.recovery.state-store.url =
 
 # If Livy can't find the yarn app within this time, consider it lost.
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -121,6 +121,19 @@
 # For zookeeper, the address to the Zookeeper servers. e.g. host1:port1,host2:port2
 # livy.server.recovery.state-store.url =
 
+# The policy of curator connecting to zookeeper.
+# For example, m, n means retry m times and the interval of retry is n milliseconds.
+# Please use the new config: livy.server.zk.retry-policy.
+# Keep this config is just for back-compatibility
+# livy.server.recovery.zk-state-store.retry-policy = 5,100
+
+# The policy of curator connecting to zookeeper.
+# For example, m, n means retry m times and the interval of retry is n milliseconds
+# livy.server.zk.retry-policy = 5,100
+
+# The dir in zk to store the data about session.
+# livy.server.recovery.zk-state-store.key-prefix = livy
+
 # If Livy can't find the yarn app within this time, consider it lost.
 # livy.server.yarn.app-lookup-timeout = 120s
 # When the cluster is busy, we may fail to launch yarn app in app-lookup-timeout, then it would
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -134,7 +134,7 @@
 
 # The policy of curator connecting to zookeeper.
 # For example, m, n means retry m times and the interval of retry is n milliseconds
-# livy.server.zk.retry-policy = 5,100
+# livy.server.zk.retry-policy =
 
 # The dir in zk to store the data about session.
 # livy.server.recovery.zk-state-store.key-prefix = livy
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -118,7 +118,7 @@
 
 # For filesystem state store, the path of the state store directory. Please don't use a filesystem
 # that doesn't support atomic rename (e.g. S3). e.g. file:///tmp/livy or hdfs:///.
-# For zookeeper, please set livy.server.zookeeper.url
+# For zookeeper, the address to the Zookeeper servers. e.g. host1:port1,host2:port2
 # livy.server.recovery.state-store.url =
 
 # If Livy can't find the yarn app within this time, consider it lost.
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -119,12 +119,17 @@
 # For filesystem state store, the path of the state store directory. Please don't use a filesystem
 # that doesn't support atomic rename (e.g. S3). e.g. file:///tmp/livy or hdfs:///.
 # For zookeeper, the address to the Zookeeper servers. e.g. host1:port1,host2:port2
+# If livy.server.recovery.state-store is zookeeper, this config is for back-compatibility,
+# so if both this config and livy.server.zookeeper.url exist,
+# livy uses livy.server.zookeeper.url first.
 # livy.server.recovery.state-store.url =
 
 # The policy of curator connecting to zookeeper.
 # For example, m, n means retry m times and the interval of retry is n milliseconds.
 # Please use the new config: livy.server.zk.retry-policy.
-# Keep this config is just for back-compatibility
+# Keep this config for back-compatibility.
+# If both this config and livy.server.zk.retry-policy exist,
+# livy uses livy.server.zk.retry-policy first.
 # livy.server.recovery.zk-state-store.retry-policy = 5,100
 
 # The policy of curator connecting to zookeeper.
--- a/server/src/main/scala/org/apache/livy/LivyConf.scala
+++ b/server/src/main/scala/org/apache/livy/LivyConf.scala
@@ -185,6 +185,10 @@ object LivyConf {
    * configure the state store.
    */
   val RECOVERY_MODE = Entry("livy.server.recovery.mode", "off")
+
+  // Zookeeper address used for HA and state store. e.g. host1:port1, host2:port2
+  val ZOOKEEPER_URL = Entry("livy.server.zookeeper.url", "")
+
   /**
    * Where Livy should store state to for recovery. Possible values:
    * <empty>: Default. State store disabled.
@@ -195,7 +199,7 @@ object LivyConf {
   /**
    * For filesystem state store, the path of the state store directory. Please don't use a
    * filesystem that doesn't support atomic rename (e.g. S3). e.g. file:///tmp/livy or hdfs:///.
-   * For zookeeper, the address to the Zookeeper servers. e.g. host1:port1,host2:port2
+   * For zookeeper, please set livy.server.zookeeper.url
    */
   val RECOVERY_STATE_STORE_URL = Entry("livy.server.recovery.state-store.url", "")
 
--- a/server/src/main/scala/org/apache/livy/LivyConf.scala
+++ b/server/src/main/scala/org/apache/livy/LivyConf.scala
@@ -187,7 +187,7 @@ object LivyConf {
   val RECOVERY_MODE = Entry("livy.server.recovery.mode", "off")
 
   // Zookeeper address used for HA and state store. e.g. host1:port1, host2:port2
-  val ZOOKEEPER_URL = Entry("livy.server.zookeeper.url", "")
+  val ZOOKEEPER_URL = Entry("livy.server.zookeeper.url", null)
 
   /**
    * Where Livy should store state to for recovery. Possible values:
@@ -204,7 +204,7 @@ object LivyConf {
    * so if both this config and livy.server.zookeeper.url exist,
    * livy uses livy.server.zookeeper.url first.
    */
-  val RECOVERY_STATE_STORE_URL = Entry("livy.server.recovery.state-store.url", "")
+  val RECOVERY_STATE_STORE_URL = Entry("livy.server.recovery.state-store.url", null)
 
   /**
     * The policy of curator connecting to zookeeper.
@@ -221,7 +221,7 @@ object LivyConf {
     * The policy of curator connecting to zookeeper.
     * For example, m, n means retry m times and the interval of retry is n milliseconds
    */
-  val ZK_RETRY_POLICY = Entry("livy.server.zk.retry-policy", "5,100")
+  val ZK_RETRY_POLICY = Entry("livy.server.zk.retry-policy", null)
 
   // The dir in zookeeper to store the data about session.
   val RECOVERY_ZK_STATE_STORE_KEY_PREFIX =
--- a/server/src/main/scala/org/apache/livy/LivyConf.scala
+++ b/server/src/main/scala/org/apache/livy/LivyConf.scala
@@ -200,6 +200,9 @@ object LivyConf {
    * For filesystem state store, the path of the state store directory. Please don't use a
    * filesystem that doesn't support atomic rename (e.g. S3). e.g. file:///tmp/livy or hdfs:///.
    * For zookeeper, the address to the Zookeeper servers. e.g. host1:port1,host2:port2
+   * If livy.server.recovery.state-store is zookeeper, this config is for back-compatibility,
+   * so if both this config and livy.server.zookeeper.url exist,
+   * livy uses livy.server.zookeeper.url first.
    */
   val RECOVERY_STATE_STORE_URL = Entry("livy.server.recovery.state-store.url", "")
 
@@ -207,7 +210,9 @@ object LivyConf {
     * The policy of curator connecting to zookeeper.
     * For example, m, n means retry m times and the interval of retry is n milliseconds.
     * Please use the new config: livy.server.zk.retry-policy.
-    * Keep this config is just for  back-compatibility.
+    * Keep this config for back-compatibility.
+    * If both this config and livy.server.zk.retry-policy exist,
+    * livy uses livy.server.zk.retry-policy first.
     */
   val RECOVERY_ZK_STATE_STORE_RETRY_POLICY =
     Entry("livy.server.recovery.zk-state-store.retry-policy", "5,100")
--- a/server/src/main/scala/org/apache/livy/LivyConf.scala
+++ b/server/src/main/scala/org/apache/livy/LivyConf.scala
@@ -203,6 +203,25 @@ object LivyConf {
    */
   val RECOVERY_STATE_STORE_URL = Entry("livy.server.recovery.state-store.url", "")
 
+  /**
+    * The policy of curator connecting to zookeeper.
+    * For example, m, n means retry m times and the interval of retry is n milliseconds.
+    * Please use the new config: livy.server.zk.retry-policy.
+    * Keep this config is just for  back-compatibility.
+    */
+  val RECOVERY_ZK_STATE_STORE_RETRY_POLICY =
+    Entry("livy.server.recovery.zk-state-store.retry-policy", "5,100")
+
+  /**
+    * The policy of curator connecting to zookeeper.
+    * For example, m, n means retry m times and the interval of retry is n milliseconds
+   */
+  val ZK_RETRY_POLICY = Entry("livy.server.zk.retry-policy", "5,100")
+
+  // The dir in zookeeper to store the data about session.
+  val RECOVERY_ZK_STATE_STORE_KEY_PREFIX =
+    Entry("livy.server.recovery.zk-state-store.key-prefix", "livy")
+
   // Livy will cache the max no of logs specified. 0 means don't cache the logs.
   val SPARK_LOGS_SIZE = Entry("livy.cache-log.size", 200)
 
--- a/server/src/main/scala/org/apache/livy/LivyConf.scala
+++ b/server/src/main/scala/org/apache/livy/LivyConf.scala
@@ -199,7 +199,7 @@ object LivyConf {
   /**
    * For filesystem state store, the path of the state store directory. Please don't use a
    * filesystem that doesn't support atomic rename (e.g. S3). e.g. file:///tmp/livy or hdfs:///.
-   * For zookeeper, please set livy.server.zookeeper.url
+   * For zookeeper, the address to the Zookeeper servers. e.g. host1:port1,host2:port2
    */
   val RECOVERY_STATE_STORE_URL = Entry("livy.server.recovery.state-store.url", "")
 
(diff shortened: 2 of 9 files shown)


runzhiwang:
@jerryshao @yiheng Could you help to review ?

conf/livy.conf.template
```diff
@@ -116,7 +118,7 @@
 
 # For filesystem state store, the path of the state store directory. Please don't use a filesystem
 # that doesn't support atomic rename (e.g. S3). e.g. file:///tmp/livy or hdfs:///.
-# For zookeeper, the address to the Zookeeper servers. e.g. host1:port1,host2:port2
+# For zookeeper, please set livy.server.zookeeper.url
```
yiheng: should we consider back-compatibility? For example, user may already use zk for session recovery. Now they need to change the configuration, or it may break their existing work.

I suggest in zk state store, we first look at the `livy.server.recovery.state-store.url`. If it doesn't exist, then we look at the `livy.server.zookeeper.url`

server/src/main/scala/org/apache/livy/LivyConf.scala
```diff
@@ -195,7 +199,7 @@ object LivyConf {
   /**
    * For filesystem state store, the path of the state store directory. Please don't use a
    * filesystem that doesn't support atomic rename (e.g. S3). e.g. file:///tmp/livy or hdfs:///.
-   * For zookeeper, the address to the Zookeeper servers. e.g. host1:port1,host2:port2
+   * For zookeeper, please set livy.server.zookeeper.url
```
yiheng: revert it

server/src/main/scala/org/apache/livy/server/LivyServer.scala
```diff
@@ -146,6 +146,10 @@ class LivyServer extends Logging {
       Future { SparkYarnApp.yarnClient }
     }
 
+    if (livyConf.get(LivyConf.RECOVERY_STATE_STORE) == "zookeeper") {
+      ZooKeeperManager(livyConf)
```
yiheng: maybe
```
ZooKeeperManager.init(livyConf)
```

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
```diff
@@ -0,0 +1,137 @@
...
+import org.apache.curator.framework.CuratorFrameworkFactory
+import org.apache.curator.retry.RetryNTimes
+import org.apache.zookeeper.KeeperException.NoNodeException
+
+import org.apache.livy.LivyConf
+import org.apache.livy.LivyConf.Entry
+import org.apache.livy.Logging
+
+object ZooKeeperManager {
+  val ZK_KEY_PREFIX_CONF = Entry("livy.server.recovery.zk-state-store.key-prefix", "livy")
```
yiheng: this is state-store configuration, I think we should put it in the StateStore class

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
```diff
@@ -0,0 +1,137 @@
...
+  val ZK_RETRY_CONF = Entry("livy.server.recovery.zk-state-store.retry-policy", "5,100")
```
yiheng: see above

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
```diff
@@ -0,0 +1,137 @@
...
+  val DUPLICATE_CREATE_EXCEPTION = "ZooKeeperManager single instance has already been created"
```
yiheng: If we only use the string once, I don't think we can just use it instead of creating a varaible

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  @volatile private var zkManager: ZooKeeperManager = _
+
+  def apply(livyConf: LivyConf,
```
yiheng: apply -> init?

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
```diff
@@ -0,0 +1,137 @@
...
+    mockCuratorClient: Option[CuratorFramework] = None):
+  ZooKeeperManager = synchronized {
```
yiheng: maybe return nothing?

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
```diff
@@ -0,0 +1,137 @@
...
+
+class ZooKeeperManager private(
+    livyConf: LivyConf,
+    mockCuratorClient: Option[CuratorFramework] = None)
+  extends JsonMapper with Logging {
+
+  import ZooKeeperManager._
+
+  private val zkAddress = livyConf.get(LivyConf.ZOOKEEPER_URL)
+  require(!zkAddress.isEmpty, s"Please config ${LivyConf.RECOVERY_STATE_STORE_URL.key}.")
```
yiheng: !zkAddress.isEmpty -> zkAddress.isDefined?

server/src/test/scala/org/apache/livy/server/recovery/ZooKeeperStateStoreSpec.scala
```diff
@@ -34,15 +34,17 @@ class ZooKeeperStateStoreSpec extends FunSpec with LivyBaseUnitTestSuite {
   describe("ZooKeeperStateStore") {
     case class TestFixture(stateStore: ZooKeeperStateStore, curatorClient: CuratorFramework)
     val conf = new LivyConf()
-    conf.set(LivyConf.RECOVERY_STATE_STORE_URL, "host")
+    conf.set(LivyConf.ZOOKEEPER_URL, "host")
```
yiheng: revert it? we may need to keep back-compatible? and add a new test case to cover the zookeeper_url case?

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
runzhiwang: zkAddress is a String.

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
runzhiwang: ZK_RETRY_CONF is a try policy of client connecting to ZooKeeper, and I think it's better to keep it in ZooKeeperManager. And the name i.e. `livy.server.recovery.zk-state-store.retry-policy` should not be changed, in case user config it in livy.conf.

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
runzhiwang: It's not just a state-store configuration, it's a configuration about where to create all zookeeper node, so it's suitable to put it here.

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
runzhiwang: why return nothing ?

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
yiheng: So it may be a little confusing that the key name contains `recovery`? Looks like it's for recovery.

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
yiheng: So it may be a little confusing that the key name contains recovery? Looks like it's for recovery.

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
yiheng: I think the key name is confusing

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
runzhiwang: Yes, but for back-compatibility, it's better to keep the name unchanged.

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
yiheng: This is an init method. Init method has side-effect and changes the context. But it usually doesn't return a value. For example, see (here)[https://github.com/apache/incubator-livy/blob/master/server/src/main/scala/org/apache/livy/server/recovery/StateStore.scala#L81], (here)[https://github.com/apache/incubator-livy/blob/master/server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala#L40] and [here](https://github.com/apache/incubator-livy/blob/master/server/src/main/scala/org/apache/livy/server/AccessFilter.scala#L25)

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
yiheng: So if user doesn't configure it, zkAddress is an empty string?

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
yiheng: Then I suggest you use another key. In zk-state-store, it first looks at the `livy.server.recovery.zk-state-store.key-prefix`, then look at the new key. What do you think?

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    zkManager
+  }
+
+  def get(): ZooKeeperManager = zkManager
+
+  // for test
+  private[recovery] def reset(): Unit = {
+    zkManager = null
```
jerryshao: I personally don't like singleton here, it is hard for testing, and that's why you need to add this method here. Why don't we change to a normal class object?

runzhiwang: If use it as a normal class object, I need to pass the object of ZooKeeperManager as a param to everywhere use it,  and a lot of code has to be changed.

jerryshao: That's fine, and that is what we do currently for other modules. I don't think there're many classes which need to be changed.

runzhiwang: Okay, I will change it to a normal class object.

conf/livy.conf.template
jerryshao: What if state-store is configured to use filesystem?

yiheng:
+1

server/src/main/scala/org/apache/livy/LivyConf.scala
```diff
@@ -185,6 +185,10 @@ object LivyConf {
    * configure the state store.
    */
   val RECOVERY_MODE = Entry("livy.server.recovery.mode", "off")
+
+  // Zookeeper address used for HA and state store. e.g. host1:port1, host2:port2
+  val ZOOKEEPER_URL = Entry("livy.server.zookeeper.url", "")
```
jerryshao: The default value of this can be null, not empty string.

server/src/main/scala/org/apache/livy/LivyConf.scala
```diff
@@ -196,9 +200,33 @@ object LivyConf {
    * For filesystem state store, the path of the state store directory. Please don't use a
    * filesystem that doesn't support atomic rename (e.g. S3). e.g. file:///tmp/livy or hdfs:///.
    * For zookeeper, the address to the Zookeeper servers. e.g. host1:port1,host2:port2
+   * If livy.server.recovery.state-store is zookeeper, this config is for back-compatibility,
+   * so if both this config and livy.server.zookeeper.url exist,
+   * livy uses livy.server.zookeeper.url first.
    */
   val RECOVERY_STATE_STORE_URL = Entry("livy.server.recovery.state-store.url", "")
```
jerryshao: Also for this.

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
```diff
@@ -0,0 +1,125 @@
...
+
+  private val zkAddress = {
+    val zkUrl = livyConf.get(LivyConf.ZOOKEEPER_URL)
+    if (!zkUrl.isEmpty) {
+      zkUrl
+    } else {
+      // for back-compatibility
+      livyConf.get(LivyConf.RECOVERY_STATE_STORE_URL)
+    }
+  }
```
jerryshao: If the default value is null, then this can be simplified to 

```
Option(xxx).getOrElse(xxxx)

require(xxxx != null)
```

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
```diff
@@ -0,0 +1,125 @@
...
+    }
+  }))
+
+  curatorClient.getUnhandledErrorListenable().addListener(new UnhandledErrorListener {
+    def unhandledError(message: String, e: Throwable): Unit = {
+      error(s"Fatal Zookeeper error. Shutting down Livy server.")
+      System.exit(1)
+    }
+  })
+  curatorClient.start()
```
jerryshao: I think we should declare start() and stop() method and call it out of this class (maybe `LiveServer` I think).

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
```diff
@@ -0,0 +1,125 @@
...
+    case _ => throw new IllegalArgumentException(
+      s"contains bad value: $retryValue. " +
+        "Correct format is <max retry count>,<sleep ms between retry>. e.g. 5,100")
+  }
+
+  private val curatorClient = mockCuratorClient.getOrElse {
+    CuratorFrameworkFactory.newClient(zkAddress, retryPolicy)
+  }
+
+  Runtime.getRuntime.addShutdownHook(new Thread(new Runnable {
```
jerryshao: the stop of this class could be added to `LivyServer`'s ShutdownHook.

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
```diff
@@ -0,0 +1,125 @@
...
+    livyConf: LivyConf,
+    mockCuratorClient: Option[CuratorFramework] = None)
+  extends JsonMapper with Logging {
+
+  def this(livyConf: LivyConf) {
+    this(livyConf, None)
+  }
+
+  private val zkAddress = {
+    val zkUrl = livyConf.get(LivyConf.ZOOKEEPER_URL)
```
jerryshao: We should also trim the string to remove unnecessary ws.

runzhiwang: updated

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
runzhiwang: updated

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
runzhiwang: updated

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
runzhiwang: updated

server/src/main/scala/org/apache/livy/LivyConf.scala
runzhiwang: updated

server/src/main/scala/org/apache/livy/LivyConf.scala
runzhiwang: updated

server/src/main/scala/org/apache/livy/server/recovery/FileSystemStateStore.scala
```diff
@@ -44,7 +44,7 @@ class FileSystemStateStore(
 
   private val fsUri = {
     val fsPath = livyConf.get(LivyConf.RECOVERY_STATE_STORE_URL)
-    require(!fsPath.isEmpty, s"Please config ${LivyConf.RECOVERY_STATE_STORE_URL.key}.")
+    require(fsPath != null, s"Please config ${LivyConf.RECOVERY_STATE_STORE_URL.key}.")
```
jerryshao: You should also check if this parameter is empty or not.

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
```diff
@@ -0,0 +1,119 @@
...
+  def this(livyConf: LivyConf) {
+    this(livyConf, None)
+  }
+
+  private val zkAddress = Option(livyConf.get(LivyConf.ZOOKEEPER_URL)).getOrElse {
+    // for back-compatibility
+    val url = livyConf.get(LivyConf.RECOVERY_STATE_STORE_URL)
+    require(url != null, s"Please config ${LivyConf.ZOOKEEPER_URL.key}.")
+    url
+  }.trim
```
jerryshao: `Option(xxx).orElse(yyy).map(_.trim).orNull`

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  curatorClient.getUnhandledErrorListenable().addListener(new UnhandledErrorListener {
+    def unhandledError(message: String, e: Throwable): Unit = {
+      error(s"Fatal Zookeeper error. Shutting down Livy server.")
+      System.exit(1)
+    }
+  })
+
+  def startCuratorClient(): Unit = {
```
jerryshao: Simplify to `start()`.

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
```diff
@@ -0,0 +1,119 @@
...
+    curatorClient.start()
+  }
+
+  def stopCuratorClient(): Unit = {
```
jerryshao: Also here change to `stop`.

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
```diff
@@ -0,0 +1,119 @@
...
+    require(url != null, s"Please config ${LivyConf.ZOOKEEPER_URL.key}.")
+    url
+  }.trim
+
+  private val retryValue = Option(livyConf.get(LivyConf.ZK_RETRY_POLICY)).getOrElse {
+    // for back-compatibility
+    val policy = livyConf.get(LivyConf.RECOVERY_ZK_STATE_STORE_RETRY_POLICY)
+    require(policy != null, s"Please config ${LivyConf.ZK_RETRY_POLICY.key}.")
+    policy
+  }.trim
```
jerryshao: Also here

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  private val curatorClient = mockCuratorClient.getOrElse {
+    CuratorFrameworkFactory.newClient(zkAddress, retryPolicy)
+  }
+
+  curatorClient.getUnhandledErrorListenable().addListener(new UnhandledErrorListener {
+    def unhandledError(message: String, e: Throwable): Unit = {
+      error(s"Fatal Zookeeper error. Shutting down Livy server.")
+      System.exit(1)
```
jerryshao: This is a legacy code, but I think here `System.exit(1)` is too overkill, maybe we should have a better way to handle this issue.

runzhiwang: updated

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
runzhiwang: updated

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
runzhiwang: updated

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
runzhiwang: updated

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
runzhiwang: updated

server/src/main/scala/org/apache/livy/server/recovery/FileSystemStateStore.scala
runzhiwang: updated

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
```diff
@@ -0,0 +1,118 @@
...
+    curatorClient.start()
+  }
+
+  def stop(): Unit = {
+    curatorClient.close()
+  }
+
+  // TODO Make sure ZK path has proper secure permissions so that other users cannot read its
+  // contents.
+
```
jerryshao: Please remove this blank line.

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
```diff
@@ -0,0 +1,118 @@
...
+        "Correct format is <max retry count>,<sleep ms between retry>. e.g. 5,100")
+  }
+
+  private val curatorClient = mockCuratorClient.getOrElse {
+    CuratorFrameworkFactory.newClient(zkAddress, retryPolicy)
+  }
+
+  curatorClient.getUnhandledErrorListenable().addListener(new UnhandledErrorListener {
+    def unhandledError(message: String, e: Throwable): Unit = {
+      error(s"Fatal Zookeeper error: ${message}.", e)
```
jerryshao: I would suggest to add code to throw uncaught exception, so that upstream could handle it, or just throw exception and finish the program.

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
runzhiwang: updated

PR #274, opened 2020-01-10
[LIVY-721][SERVER] Distributed session id generation

runzhiwang:

1. When generate unique session id with multiple-active HA mode. First, get the distributed lock,
Second, get the session id from filesystem or zookeeper. Third, increase the session id and save it in the filesystem or zookeeper. Forth, release the distributed lock.

2. ZooKeeperManager provides the distributed lock to generate the distributed session id.

3. `LocalSessionIdGenerator` used when only one livy work. and `DistributedSessionIdGenerator` used when serveral livy work at the same time, such as in multi-active HA mode. Maybe the name `Local `of  `LocalSessionIdGenerator` is confused, because it also use hdfs or zookeeper. But I do not have a better name, maybe rename `LocalSessionIdGenerator` to `StandaloneSessionIdGenerator`  ？
  
## How was this patch tested?

Existed UT and IT.


runzhiwang:
@jerryshao @yiheng Could you help review this PR ?

jerryshao:
Seems like the JIRA number is not the correct one. LIVY-718 is an umbrella jira.

runzhiwang:
@jerryshao Yes, it's a wrong jira number. I have changed it from LIVY-718 to LIVY-721.

conf/livy.conf.template
```diff
@@ -109,6 +109,8 @@
 # livy.server.recovery.mode = off
 # Zookeeper address used for HA and state store. e.g. host1:port1, host2:port2
 # livy.server.zookeeper.url =
+# Whether to enable HA with multi-active mode, by default it is false.
+# livy.server.ha.multi-active.enabled = false
```
yiheng: Should we change it to something like?
```
livy.servver.ha.mode=multi-active
```

conf/livy.conf.template
```diff
@@ -138,6 +140,8 @@
 
 # The dir in zk to store the data about session.
 # livy.server.recovery.zk-state-store.key-prefix = livy
+# The dir related to zookeeper utility, such as: distributed lock, service discovery.
+# livy.server.zk.utility.dir-prefix = livy/zk-utility
```
yiheng: Should we change it to below?
```
livy.server.zk.prefix = livy/zk
```

server/src/main/scala/org/apache/livy/sessions/DistributedSessionIdGenerator.scala
```diff
@@ -0,0 +1,40 @@
...
+import org.apache.livy.server.recovery.BlackholeStateStore
+import org.apache.livy.server.recovery.SessionStore
+import org.apache.livy.server.recovery.ZooKeeperManager
+
+class DistributedSessionIdGenerator(
+    sessionType: String,
+    sessionStore: SessionStore,
+    zkManager: ZooKeeperManager) extends SessionIdGenerator {
+
+  require(sessionStore.getStore.isInstanceOf[BlackholeStateStore] == false)
```
yiheng: I don't think this check is safe. Say in the future a LocalFileSystem state store is added, the check will not alarm

conf/livy.conf.template
yiheng: Will this be only used in distributed lock and service discovery? If so, I suggest using configurations like
```
# The dir in zk to store distributed lock
livy.server.zk.lock = livy/zk/lock
# The dir in zk to store service information
livy.server.zk.services = livy/zk/services

File: server/src/main/scala/org/apache/livy/sessions/DistributedSessionIdGenerator.scala [hunk 4]
```diff
@@ -0,0 +1,40 @@
+/*
```
yiheng: I think you can merge DistributedSessionIdGenerator, LocalSessionIdGenerator and SessionIdGenerator to one file

server/src/main/scala/org/apache/livy/sessions/SessionIdGenerator.scala
```diff
@@ -0,0 +1,33 @@
...
+abstract class SessionIdGenerator {
+  /**
+    * Get next session id, then increase next session id and save it in store.
+    */
+  def getNextSessionId(): Int
+
+  /**
+    * recover next session id from store.
+    */
+  def recover(): Unit
```
yiheng: Do we really need the method to be called explicit? Can we hide it in the constructor?

server/src/main/scala/org/apache/livy/sessions/LocalSessionIdGenerator.scala
```diff
@@ -0,0 +1,39 @@
...
+  private val idCounter = new AtomicInteger(0)
+
+  def getNextSessionId(): Int = {
+    val result = idCounter.getAndIncrement()
+    sessionStore.saveNextSessionId(sessionType, result + 1)
+    result
+  }
+
+  def recover(): Unit = {
+    idCounter.set(sessionStore.getNextSessionId(sessionType))
```
yiheng: can we move it to constructor?

server/src/main/scala/org/apache/livy/sessions/SessionIdGenerator.scala
```diff
@@ -0,0 +1,33 @@
...
+package org.apache.livy.sessions
+
+/**
+  * Interface of session id generator.
+  */
+abstract class SessionIdGenerator {
+  /**
+    * Get next session id, then increase next session id and save it in store.
+    */
+  def getNextSessionId(): Int
```
yiheng: getNextSessionId -> getAndIncrementSessionId ?

server/src/main/scala/org/apache/livy/sessions/SessionIdGenerator.scala
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package org.apache.livy.sessions
+
+/**
+  * Interface of session id generator.
+  */
+abstract class SessionIdGenerator {
+  /**
+    * Get next session id, then increase next session id and save it in store.
```
yiheng: remove `and save it in store` ?

yiheng: I think the `save it in store` is not guaranteed here. If we require the developer to follow the pattern, I suggest we break the getNextSessionId to two methods:

```scala
/**
 * Get and increase the next session-id in an atomic operation. The session id will be persisted and recoverable.
 */
def getNextSessionId: Int = {
    val nextSessionId = getAndIncreaseId()
    persist()
    return nextSessionId;
}

private def getAndIncreaseId: Int

private def persist: Unit
```

yiheng:
Do we have some test cases to cover the sessionIdGenerator?

server/src/main/scala/org/apache/livy/sessions/SessionIdGenerator.scala
runzhiwang: @yiheng Good point.  Actually  it does not need `recover ` any more. For `LocalSessionIdGenerator`, it store session id in memory, so can not recover. For `DistributedSessionIdGenerator`,  it will get session id from store every time `getNextSessionId`, so it does not need recover when restart. I will remove it.

conf/livy.conf.template
runzhiwang: In this PR, it only used for  distributed lock. And I will change it to `livy.server.zk.lock = livy/zk/lock`. `livy.server.zk.services = livy/zk/services` will be added in the PR used it.

server/src/main/scala/org/apache/livy/LivyConf.scala
```diff
@@ -227,6 +242,9 @@ object LivyConf {
   val RECOVERY_ZK_STATE_STORE_KEY_PREFIX =
     Entry("livy.server.recovery.zk-state-store.key-prefix", "livy")
 
+  // The dir in zk to store distributed lock.
+  val ZK_LOCK_DIR = Entry("livy.server.zk.lock", "livy/zk/lock")
```
jerryshao: Is it better to change to `livy.server.ha.zk-lock`? Since it is HA related.

server/src/main/scala/org/apache/livy/server/LivyServer.scala
```diff
@@ -417,13 +417,20 @@ class LivyServer extends Logging {
   private[livy] def testRecovery(livyConf: LivyConf): Unit = {
     if (!livyConf.isRunningOnYarn()) {
       // If recovery is turned on but we are not running on YARN, quit.
-      require(livyConf.get(LivyConf.RECOVERY_MODE) == SESSION_RECOVERY_MODE_OFF,
+      val haMode = Option(livyConf.get(LivyConf.HA_MODE)).
+        orElse(Option(livyConf.get(LivyConf.RECOVERY_MODE))).
+        map(_.trim).orNull
```
jerryshao: nit: put `.` in the beginning of the line.

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
```diff
@@ -84,13 +87,19 @@ class SessionManager[S <: Session, R <: RecoveryMetadata : ClassTag](
   private[this] final val sessionStateRetainedInSec =
     TimeUnit.MILLISECONDS.toNanos(livyConf.getTimeAsMs(LivyConf.SESSION_STATE_RETAIN_TIME))
 
+  private final val sessionIdGenerator = {
+    if (livyConf.get(LivyConf.HA_MODE) == LivyServer.HA_MODE_MULTI_ACTIVE) {
+      new DistributedSessionIdGenerator(sessionType, sessionStore, zkManager.get)
```
jerryshao: What if here `zkManager` is `None`?

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
```diff
@@ -145,7 +154,11 @@ class SessionManager[S <: Session, R <: RecoveryMetadata : ClassTag](
   }
 
   def shutdown(): Unit = {
-    val recoveryEnabled = livyConf.get(LivyConf.RECOVERY_MODE) != SESSION_RECOVERY_MODE_OFF
+    val haMode = Option(livyConf.get(LivyConf.HA_MODE)).
+      orElse(Option(livyConf.get(LivyConf.RECOVERY_MODE))).
+      map(_.trim).orNull
```
jerryshao: Nit: put `.` in the beginning of the line.

jerryshao (commented):
Do we need to add some UTs to cover the code changes?

server/src/main/scala/org/apache/livy/server/LivyServer.scala
```diff
@@ -417,13 +417,20 @@ class LivyServer extends Logging {
...
+
+      require(haMode == LivyServer.HA_MODE_OFF,
         "Session recovery requires YARN.")
     }
   }
 }
 
 object LivyServer {
+  val HA_MODE_OFF = "off"
```
yiheng: move these to LivyConf ?

server/src/main/scala/org/apache/livy/LivyConf.scala
```diff
@@ -186,6 +189,18 @@ object LivyConf {
...
+    * HA mode of Livy. Possible values:
+    * 1. null: Default. livy will use the value of livy.server.recovery.mode
+    * 2. off: Turn off recovery. Every time Livy shuts down, it stops and forgets all sessions.
+    * 3. recovery: Livy persists session info to the state store. When Livy restarts, it recovers
+    *              previous sessions from the state store.
+    *              Must set livy.server.recovery.state-store and
+    *              livy.server.recovery.state-store.url to configure the state store.
+    * 4. multi-active: HA with multi-active mode.
+    */
+  val HA_MODE = Entry("livy.server.ha.mode", null)
```
yiheng: default value is off?

conf/livy.conf.template
```diff
@@ -138,6 +151,8 @@
 
 # The dir in zk to store the data about session.
 # livy.server.recovery.zk-state-store.key-prefix = livy
+# The dir in zk to store distributed lock.
+# livy.server.zk.lock = livy/zk/lock
```
yiheng: should it be `/livy/zk/lock`

yiheng: Do we really need a lock dir? I think the lock should be put on some node where the concurrent operation happens. For example, the session id file when generating a unique distributed session id.

If we use a single lock dir, there will be some problem when the lock object is different. For example, generate session-id should not be blocked by update some session map.

server/src/main/scala/org/apache/livy/LivyConf.scala
runzhiwang: If default value is off, Livy will use `livy.server.ha.mode` rather than` livy.server.recovery.mode` even though  user does not config livy.server.ha.mode, which break back-compatibility.

wypoon:
See https://github.com/apache/incubator-livy/pull/275/commits/29114951c48d3d1829d4a0fdaaa338983e193caf for my fix to the python-api CI failure.

runzhiwang:
@jerryshao @yiheng UTs has been added.

server/src/main/scala/org/apache/livy/server/recovery/StateStore.scala
```diff
@@ -104,9 +108,13 @@ object StateStore extends Logging {
   }
 
   private[recovery] def pickStateStore(livyConf: LivyConf): Class[_] = {
-    livyConf.get(LivyConf.RECOVERY_MODE) match {
-      case SESSION_RECOVERY_MODE_OFF => classOf[BlackholeStateStore]
-      case SESSION_RECOVERY_MODE_RECOVERY =>
+    val haMode = Option(livyConf.get(LivyConf.HA_MODE)).
+      orElse(Option(livyConf.get(LivyConf.RECOVERY_MODE))).
+      map(_.trim).orNull
```
jerryshao: Move `.` to the beginning of the line. Also there's other duplicated code in `testRecovery`, can we merge them together?

server/src/main/scala/org/apache/livy/sessions/SessionIdGenerator.scala
```diff
@@ -0,0 +1,87 @@
...
+  def getNextSessionId(): Int
+
+  protected def getAndIncreaseId(): Int
+
+  protected def persist(id: Int): Unit = {
+    sessionStore.saveNextSessionId(sessionType, id)
+  }
+}
+
+class LocalSessionIdGenerator(
```
jerryshao: Would be better to rename to `InMemorySessionIdGenerator`.

server/src/main/scala/org/apache/livy/sessions/SessionIdGenerator.scala
```diff
@@ -0,0 +1,87 @@
...
+    "Choose a distributed store such as hdfs or zookeeper")
+
+  val distributedLock = mockLock.getOrElse(
+    zkManager.createLock(SessionStore.sessionIdLockPath(sessionType)))
+
+  override def getNextSessionId(): Int = {
+    distributedLock.acquire()
+    val result = getAndIncreaseId()
+    persist(result + 1)
+    distributedLock.release()
```
jerryshao: What if there's an exception thrown in the above code, are we going to lose the lock?

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
```diff
@@ -145,7 +157,11 @@ class SessionManager[S <: Session, R <: RecoveryMetadata : ClassTag](
   }
 
   def shutdown(): Unit = {
-    val recoveryEnabled = livyConf.get(LivyConf.RECOVERY_MODE) != SESSION_RECOVERY_MODE_OFF
+    val haMode = Option(livyConf.get(LivyConf.HA_MODE))
+      .orElse(Option(livyConf.get(LivyConf.RECOVERY_MODE)))
+      .map(_.trim).orNull
```
jerryshao: Also here, duplicated code.

server/src/test/scala/org/apache/livy/sessions/SessionManagerSpec.scala
```diff
@@ -271,17 +272,61 @@ class SessionManagerSpec extends FunSpec with Matchers with LivyBaseUnitTestSuit
...
 
-      val sm = new BatchSessionManager(conf, sessionStore, Some(Seq(session)))
+      val sm = new BatchSessionManager(conf, sessionStore, None, Some(Seq(session)))
       sm.get(sessionId) shouldBe defined
       sm.shutdown()
 
       verify(session, never).stop()
     }
+
+    it("should crate instance of LocalSessionIdGenerator if livy.server.ha.mode is off") {
```
jerryshao: Typo: "create".

server/src/main/scala/org/apache/livy/sessions/SessionIdGenerator.scala
```diff
@@ -0,0 +1,93 @@
...
+    zkManager.createLock(SessionStore.sessionIdLockPath(sessionType)))
+
+  override def getNextSessionId(): Int = {
+    distributedLock.acquire()
+    try {
+      val result = getAndIncreaseId()
+      persist(result + 1)
+      result
+    } catch {
+      case e: Exception => throw new LivyUncaughtException(e.getMessage)
```
jerryshao: What will be happened here when occurring an exception？

server/src/main/scala/org/apache/livy/sessions/SessionIdGenerator.scala
```diff
@@ -0,0 +1,93 @@
...
+    } finally {
+      distributedLock.release()
```
jerryshao: Will here throw an exception again?

andrasbeni:
@runzhiwang do you plan to continue working on this? If not, I am happy to address present concerns and move forward the feature.

PR #268, opened 2019-12-20
[LIVY-735][RSC] Fix rpc channel closed when multi clients connect to one driver

runzhiwang:
## What changes were proposed in this pull request?

Currently, the driver tries to support communicating with multi-clients, by registering each client at https://github.com/apache/incubator-livy/blob/master/rsc/src/main/java/org/apache/livy/rsc/driver/RSCDriver.java#L220.

But actually, if multi-clients connect to one driver, the rpc channel will close, the reason are as follows.

1.  In every communication, client sends two packages to driver: header{type, id}, and payload at https://github.com/apache/incubator-livy/blob/master/rsc/src/main/java/org/apache/livy/rsc/rpc/RpcDispatcher.java#L144. 

2. If client1 sends header1, payload1, and client2 sends header2, payload2 at the same time. 
  The driver receives the package in the order: header1, header2, payload1, payload2.

3. When driver receives header1, driver assigns lastHeader at https://github.com/apache/incubator-livy/blob/master/rsc/src/main/java/org/apache/livy/rsc/rpc/RpcDispatcher.java#L73.

4. Then driver receives header2, driver process it as a payload at https://github.com/apache/incubator-livy/blob/master/rsc/src/main/java/org/apache/livy/rsc/rpc/RpcDispatcher.java#L78 which cause exception and rpc channel closed.

In the muti-active HA mode, the design doc is at: https://docs.google.com/document/d/1bD3qYZpw14_NuCcSGUOfqQ0pqvSbCQsOLFuZp26Ohjc/edit?usp=sharing, the session is allocated among servers by consistent hashing. If a new livy joins, some session will be migrated from old livy to new livy. If the session client in new livy connect to driver before stoping session client in old livy, then two session clients will both connect to driver, and rpc channel close.  In this case, it's hard to ensure only one client connect to one driver at any time. So it's better to support multi-clients connect to one driver, which has no side effects.

How to fix:
1. Move the code of processing client message from `RpcDispatcher` to each `Rpc`.
2. Each `Rpc` registers itself to `channelRpc` in RpcDispatcher.
3. `RpcDispatcher` dispatches each message to `Rpc` according to  `ctx.channel()`.

## How was this patch tested?

Existed UT and IT

PR #275, opened 2020-01-17
[LIVY-745] Ensure that a single RSCClientFactory gets loaded.

wypoon:

In LivyClientBuilder, a ServiceLoader is used to load configured LivyClientFactory providers (HttpClientFactory and RSCClientFactory). Correct behavior of RSCClientFactory implicitly depends on there being a single instance of it in the Livy server.
Instead of instantiating a new ServiceLoader every time the build method is called on a LivyClientBuilder, keep a single ServiceLoader in a static field. ServiceLoader instances are not safe for use by multiple concurrent threads, so have the ServiceLoader load the LivyClientFactory implementations into a static List. Then LivyClientBuilder#build will use the single instance of either HttpClientFactory or RSCClientFactory in this List to create a LivyClient.

## How was this patch tested?

Tested by having 20 clients connect to the Livy Thrift server simultaneously. Before the change, the behavior described in https://issues.apache.org/jira/browse/LIVY-745 was encountered. With the change, the problem is not encountered.
Also added a unit test that fails before this change to LivyClientBuilder, and passes with the change.

### DIFF
--- a/api/src/main/java/org/apache/livy/LivyClientBuilder.java
+++ b/api/src/main/java/org/apache/livy/LivyClientBuilder.java
@@ -23,9 +23,11 @@ import java.io.Reader;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.ServiceLoader;
+import java.util.concurrent.CopyOnWriteArrayList;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 /**
@@ -38,18 +40,15 @@ public final class LivyClientBuilder {
   private static final ServiceLoader<LivyClientFactory> CLIENT_FACTORY_LOADER =
     ServiceLoader.load(LivyClientFactory.class, classLoader());
 
-  private static LivyClientFactory getLivyClientFactory() {
-    LivyClientFactory factory = null;
+  private static List<LivyClientFactory> getLivyClientFactories() {
+    List<LivyClientFactory> factories = new CopyOnWriteArrayList<>();
     for (LivyClientFactory f : CLIENT_FACTORY_LOADER) {
-      factory = f;
-      if (factory != null) {
-        break;
-      }
+      factories.add(f);
     }
-    return factory;
+    return factories;
   }
 
-  private static final LivyClientFactory CLIENT_FACTORY = getLivyClientFactory();
+  private static final List<LivyClientFactory> CLIENT_FACTORIES = getLivyClientFactories();
 
   private final Properties config;
 
@@ -134,16 +133,22 @@ public final class LivyClientBuilder {
     }
 
     LivyClient client = null;
-    if (CLIENT_FACTORY == null) {
+    if (CLIENT_FACTORIES.isEmpty()) {
       throw new IllegalStateException("No LivyClientFactory implementation was found.");
     }
-    try {
-      client = CLIENT_FACTORY.createClient(uri, config);
-    } catch (Exception e) {
-      if (!(e instanceof RuntimeException)) {
-        e = new RuntimeException(e);
+
+    for (LivyClientFactory factory : CLIENT_FACTORIES) {
+      try {
+        client = factory.createClient(uri, config);
+      } catch (Exception e) {
+        if (!(e instanceof RuntimeException)) {
+          e = new RuntimeException(e);
+        }
+        throw (RuntimeException) e;
       }
-      throw (RuntimeException) e;
+      if (client != null) {
+        break;
+      }
     }
 
     if (client == null) {
--- a/api/src/main/java/org/apache/livy/LivyClientBuilder.java
+++ b/api/src/main/java/org/apache/livy/LivyClientBuilder.java
@@ -38,6 +38,19 @@ public final class LivyClientBuilder {
   private static final ServiceLoader<LivyClientFactory> CLIENT_FACTORY_LOADER =
     ServiceLoader.load(LivyClientFactory.class, classLoader());
 
+  private static LivyClientFactory getLivyClientFactory() {
+    LivyClientFactory factory = null;
+    for (LivyClientFactory f : CLIENT_FACTORY_LOADER) {
+      factory = f;
+      if (factory != null) {
+        break;
+      }
+    }
+    return factory;
+  }
+
+  private static final LivyClientFactory CLIENT_FACTORY = getLivyClientFactory();
+
   private final Properties config;
 
   /**
@@ -121,29 +134,18 @@ public final class LivyClientBuilder {
     }
 
     LivyClient client = null;
-    boolean loaded = false;
-    // ServiceLoader instances are not safe for use by multiple concurrent threads.
-    // Ensure that the ServiceLoader's iterator is called by only one thread at a time.
-    synchronized (CLIENT_FACTORY_LOADER) {
-      for (LivyClientFactory factory : CLIENT_FACTORY_LOADER) {
-        loaded = true;
-        try {
-          client = factory.createClient(uri, config);
-        } catch (Exception e) {
-          if (!(e instanceof RuntimeException)) {
-            e = new RuntimeException(e);
-          }
-          throw (RuntimeException) e;
-        }
-        if (client != null) {
-          break;
-        }
-      }
-    }
-
-    if (!loaded) {
+    if (CLIENT_FACTORY == null) {
       throw new IllegalStateException("No LivyClientFactory implementation was found.");
     }
+    try {
+      client = CLIENT_FACTORY.createClient(uri, config);
+    } catch (Exception e) {
+      if (!(e instanceof RuntimeException)) {
+        e = new RuntimeException(e);
+      }
+      throw (RuntimeException) e;
+    }
+
     if (client == null) {
       // Redact any user information from the URI when throwing user-visible exceptions that might
       // be logged.
--- a/api/src/main/java/org/apache/livy/LivyClientBuilder.java
+++ b/api/src/main/java/org/apache/livy/LivyClientBuilder.java
@@ -35,6 +35,9 @@ public final class LivyClientBuilder {
 
   public static final String LIVY_URI_KEY = "livy.uri";
 
+  private static final ServiceLoader<LivyClientFactory> CLIENT_FACTORY_LOADER =
+    ServiceLoader.load(LivyClientFactory.class, classLoader());
+
   private final Properties config;
 
   /**
@@ -118,24 +121,21 @@ public final class LivyClientBuilder {
     }
 
     LivyClient client = null;
-    ServiceLoader<LivyClientFactory> loader = ServiceLoader.load(LivyClientFactory.class,
-      classLoader());
-    if (!loader.iterator().hasNext()) {
-      throw new IllegalStateException("No LivyClientFactory implementation was found.");
-    }
-
-    Exception error = null;
-    for (LivyClientFactory factory : loader) {
-      try {
-        client = factory.createClient(uri, config);
-      } catch (Exception e) {
-        if (!(e instanceof RuntimeException)) {
-          e = new RuntimeException(e);
+    // ServiceLoader instances are not safe for use by multiple concurrent threads.
+    // Ensure that the ServiceLoader's iterator is called by only one thread at a time.
+    synchronized (LivyClientBuilder.class) {
+      for (LivyClientFactory factory : CLIENT_FACTORY_LOADER) {
+        try {
+          client = factory.createClient(uri, config);
+        } catch (Exception e) {
+          if (!(e instanceof RuntimeException)) {
+            e = new RuntimeException(e);
+          }
+          throw (RuntimeException) e;
         }
-        throw (RuntimeException) e;
-      }
-      if (client != null) {
-        break;
+        if (client != null) {
+          break;
+        }
       }
     }
 
@@ -158,10 +158,10 @@ public final class LivyClientBuilder {
     return client;
   }
 
-  private ClassLoader classLoader() {
+  private static ClassLoader classLoader() {
     ClassLoader cl = Thread.currentThread().getContextClassLoader();
     if (cl == null) {
-      cl = getClass().getClassLoader();
+      cl = LivyClientBuilder.class.getClassLoader();
     }
     return cl;
   }
--- a/api/src/main/java/org/apache/livy/LivyClientBuilder.java
+++ b/api/src/main/java/org/apache/livy/LivyClientBuilder.java
@@ -23,11 +23,11 @@ import java.io.Reader;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.ServiceLoader;
-import java.util.concurrent.CopyOnWriteArrayList;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 /**
@@ -41,7 +41,7 @@ public final class LivyClientBuilder {
     ServiceLoader.load(LivyClientFactory.class, classLoader());
 
   private static List<LivyClientFactory> getLivyClientFactories() {
-    List<LivyClientFactory> factories = new CopyOnWriteArrayList<>();
+    List<LivyClientFactory> factories = new ArrayList<>();
     for (LivyClientFactory f : CLIENT_FACTORY_LOADER) {
       factories.add(f);
     }
--- a/api/src/main/java/org/apache/livy/LivyClientBuilder.java
+++ b/api/src/main/java/org/apache/livy/LivyClientBuilder.java
@@ -121,10 +121,12 @@ public final class LivyClientBuilder {
     }
 
     LivyClient client = null;
+    boolean loaded = false;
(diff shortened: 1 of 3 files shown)


wypoon:
@jerryshao @mgaido91 can you please take a look?

api/src/main/java/org/apache/livy/LivyClientBuilder.java
```diff
@@ -35,6 +35,9 @@
 
   public static final String LIVY_URI_KEY = "livy.uri";
 
+  private static final ServiceLoader<LivyClientFactory> CLIENT_FACTORY_LOADER =
```
mgaido91: why do we need to move it here?

api/src/main/java/org/apache/livy/LivyClientBuilder.java
```diff
@@ -118,24 +121,21 @@ public LivyClient build() {
     }
 
     LivyClient client = null;
-    ServiceLoader<LivyClientFactory> loader = ServiceLoader.load(LivyClientFactory.class,
-      classLoader());
-    if (!loader.iterator().hasNext()) {
-      throw new IllegalStateException("No LivyClientFactory implementation was found.");
```
mgaido91: now we are missing this check, aren't we?

api/src/main/java/org/apache/livy/LivyClientBuilder.java
```diff
@@ -118,24 +121,21 @@ public LivyClient build() {
...
-    Exception error = null;
-    for (LivyClientFactory factory : loader) {
-      try {
-        client = factory.createClient(uri, config);
-      } catch (Exception e) {
-        if (!(e instanceof RuntimeException)) {
-          e = new RuntimeException(e);
+    // ServiceLoader instances are not safe for use by multiple concurrent threads.
+    // Ensure that the ServiceLoader's iterator is called by only one thread at a time.
+    synchronized (LivyClientBuilder.class) {
```
mgaido91: shall we also put a check for the client being null here in order to avoid multiple instantiations?

api/src/main/java/org/apache/livy/LivyClientBuilder.java
wypoon: I think it is ok to not perform this check. If the iterator is empty, then after the for loop below, client will be null, and an exception (in this case, an IllegalArgumentException) will be thrown. So there will still be an error. I am trying to minimize the code that gets called in the synchronized block. I don't feel this check adds a lot of value, because this IllegalStateException will only happen if the services file is not in the build, which would be a build issue. If you feel very strongly about having this check, I can put it back in.
The other reason I took it out was that ideally this check should be done just once. The normal way to do that would be in a static initializer, but you normally should not throw an exception in those (you can throw a RuntimeException, but I think that is still bad form).

api/src/main/java/org/apache/livy/LivyClientBuilder.java
wypoon: It doesn't really matter. We are in the LivyClientBuilder class so we know it has been loaded. It is enough to synchronize on that.

wypoon:
> would be great to have a UT....

It would. The bug itself seems (to me) a difficult thing to test in a UT. Do you have a suggestion?

api/src/main/java/org/apache/livy/LivyClientBuilder.java
bersprockets: You could also simply set a flag in the new for loop that iterates over factories. Immediately after the loop, if the flag is not set, you can throw an exception ("No LivyClientFactory implementation was found."). That *might* be cheaper than first creating an iterator and calling hasNext... but I don't know that.

api/src/main/java/org/apache/livy/LivyClientBuilder.java
```diff
@@ -118,24 +121,21 @@ public LivyClient build() {
     }
 
     LivyClient client = null;
-    ServiceLoader<LivyClientFactory> loader = ServiceLoader.load(LivyClientFactory.class,
```
bersprockets: It's still slightly mysterious to me why this was not a problem even with connections coming in slowly, since a new ServiceLoader is created each time... unless the cache contains ServiceLoader<LivyClientFactory> instances, which would obviously be a static cache since they are created from a static method, instead of LivyClientFactory instances cached within a ServiceLoader instance.

mgaido91:
I think this can be tested easily by defining your class which keeps a counter incremented in its constructor. We need to ensure only one `LivyClientFactory` is created.

api/src/main/java/org/apache/livy/LivyClientBuilder.java
mgaido91: I think we can also do this only once when the `ServiceLoader` is created. But I do see value in keeping the behavior consistent so that a user who faced this issue once, knows what to do facing it again.

wypoon: I'll restore a check for whether any LivyClientFactory implementation was found.
@bersprockets thanks for the suggestion.

wypoon:
> I think this can be tested easily by defining your class which keeps a counter incremented in its constructor. We need to ensure only one `LivyClientFactory` is created.

My initial, instinctive reaction is that I don't like this idea. You're proposing to add something to the LivyClientFactory that is not necessary for its functionality, simply in order to construct a test. This is different from exposing a private field or a method in a class (needed for its functioning) for testing purposes.
Anyway, let me think more about it.

mgaido91:
> My initial, instinctive reaction is that I don't like this idea. You're proposing to add something to the LivyClientFactory that is not necessary for its functionality, simply in order to construct a test. This is different from exposing a private field or a method in a class (needed for its functioning) for testing purposes.

I am not saying that. I am saying you can mock that class for instance.

wypoon:
> > My initial, instinctive reaction is that I don't like this idea. You're proposing to add something to the LivyClientFactory that is not necessary for its functionality, simply in order to construct a test. This is different from exposing a private field or a method in a class (needed for its functioning) for testing purposes.
> 
> I am not saying that. I am saying you can mock that class for instance.

Sorry I misunderstood.
The trick then is getting a ServiceLoader for LivyClientFactory that loads the mock LivyClientFactory. Let me look into it.

wypoon:
I found that there was already a test implementation of LivyClientFactory along with the services file to load it, so I modified this TestClientFactory and added a testcase to TestLivyClientBuilder.
Before the change to LivyClientBuilder, running
`mvn -pl api -Dtest=TestLivyClientBuilder#testSinglenessOfClientFactory test`
resulted in
`Failed tests: 
  TestLivyClientBuilder.testSinglenessOfClientFactory:97 expected:<1> but was:<2>`
With the change, the test passes.

api/src/main/java/org/apache/livy/LivyClientBuilder.java
mgaido91: Yes, we know it has been loaded, the point is: do we sync on the class or on its field? Since we want to protect `CLIENT_FACTORY_LOADER` from concurrency issues, it is logical to use that, so if in the future we will sync on `LivyClientBuilder.class` we do not create useless locks.

api/src/main/java/org/apache/livy/LivyClientBuilder.java
```diff
@@ -118,27 +121,29 @@ public LivyClient build() {
...
+    boolean loaded = false;
+    // ServiceLoader instances are not safe for use by multiple concurrent threads.
+    // Ensure that the ServiceLoader's iterator is called by only one thread at a time.
+    synchronized (LivyClientBuilder.class) {
+      for (LivyClientFactory factory : CLIENT_FACTORY_LOADER) {
+        loaded = true;
+        try {
+          client = factory.createClient(uri, config);
+        } catch (Exception e) {
+          if (!(e instanceof RuntimeException)) {
```
mgaido91: why do we need this?

api/src/test/java/org/apache/livy/TestLivyClientBuilder.java
```diff
@@ -78,4 +78,22 @@ public void testRedaction() throws Exception {
     }
   }
 
+  @Test
+  public void testSinglenessOfClientFactory() throws Exception {
+    Runnable r = () -> {
```
mgaido91: does it need to be concurrent? don't we have the same problem without concurrency?

api/src/main/java/org/apache/livy/LivyClientBuilder.java
wypoon: That's existing code. (Everything in the for loop is the same as before, just indented one level because the for loop is wrapped in the synchronized.)
The reason for the catch block here is that `LivyClientBuilder#build` does not throw any checked exception.

api/src/test/java/org/apache/livy/TestLivyClientBuilder.java
wypoon: You are right that even without concurrency, the `TestClientFactory.getInstanceCount()` would be greater than 1 before.
I was using concurrent calls to `LivyClientBuilder#build` simply because that was how the problem was discovered.

mgaido91: I think we should Keep It Simple Stupid, if we can. So I'd prefer a non-concurrent version.

wypoon: Simplified the test.

wypoon:
@mgaido91 thanks again for reviewing. I have made the changes you suggested.

api/src/main/java/org/apache/livy/LivyClientBuilder.java
```diff
@@ -118,27 +121,29 @@ public LivyClient build() {
...
-    for (LivyClientFactory factory : loader) {
-      try {
-        client = factory.createClient(uri, config);
-      } catch (Exception e) {
-        if (!(e instanceof RuntimeException)) {
-          e = new RuntimeException(e);
+    boolean loaded = false;
+    // ServiceLoader instances are not safe for use by multiple concurrent threads.
+    // Ensure that the ServiceLoader's iterator is called by only one thread at a time.
+    synchronized (CLIENT_FACTORY_LOADER) {
```
mgaido91: the change seems ok to me now. I just have one concern: this synchronization over the creation of a client may become a performance issue. Since in the createClient we establish connections, it is good practice to avoid sync these things. Hence, I am wondering if we can create a thread safe wrapper over the `ServiceLoader` or avoid `ServiceLoader`'s usage and switch to something which doesn't have this issue so we can avoid this synchronization. cc @vanzin @jerryshao

wypoon: The performance is actually much better now (with this change) than before. With this change, I can create 50 concurrent connections to Livy "all at once". Before this change, I could barely create two dozen, and only by using retries, since more than 11 could not be created "at once"; and that took much longer.
Note that without the synchronization, when hitting Livy with multiple concurrent connections, some connection could fail with
`"java.lang.IllegalArgumentException: URI 'rsc:/' is not supported by any registered client factories."`
and in the server log, this is preceded by an ERROR
```
Caused by: java.util.NoSuchElementException
        at sun.misc.CompoundEnumeration.nextElement(CompoundEnumeration.java:5)
        at java.util.ServiceLoader$LazyIterator.hasNextService(ServiceLoader.java:357)
        at java.util.ServiceLoader$LazyIterator.hasNext(ServiceLoader.java:393)
        at java.util.ServiceLoader$1.hasNext(ServiceLoader.java:474)
        at org.apache.livy.LivyClientBuilder.build(LivyClientBuilder.java:124)
```
(Line number in LivyClientBuilder is different now.)
Now that we have this hugely better performance, I understand you're trying to see if it can be better (but that is on a different order of magnitude).
I don't know why service loading was used in the first place, not having been involved in Livy. If we don't need it, sure, let's do it a different way. But if that's more involved, let's do it as a separate PR.

wypoon: Actually, there is something simple we can do, which is what I'd originally thought to do.
In the for loop, we're trying to create a client using each client factory loaded by the `ServiceLoader`, exiting once we get a client. In reality, there is only one client factory in the actual system, `RSCClientFactory`. If we agree that we will only ever use one client factory (at any one time) in a Livy server, then there is no need to have this for loop here. We would simply load a client factory once (statically). And every time we need a client (call `LivyClientBuilder#build`), we would use that client factory to `createClient`.
The ostensible reason for iterating through client factories this way here is that we could have more than one, and one `createClient` call could work with one and another `createClient` call might not but could work with a different one. In reality that does not happen. If we agree that any client factory we load (the first one wins) should always work for any `createClient` call, then we really don't need this for loop here. And we don't need to synchronize.

wypoon: Implemented this and updated the PR description.

wypoon:
I am unfamiliar with how the integration tests are set up. I had deployed the change to a real cluster, and did not find any issues with creating sessions. In the integration tests, creating sessions is failing with
```
JobApiIT:
- create a new session and upload test jar *** FAILED *** (2 milliseconds)
  java.lang.IllegalArgumentException: URI 'http://localhost:8998' is not supported by any registered client factories.
  at org.apache.livy.LivyClientBuilder.build(LivyClientBuilder.java:162)
  at org.apache.livy.test.JobApiIT.org$apache$livy$test$JobApiIT$$createClient(JobApiIT.scala:312)
```
Any insight would be appreciated.

api/src/main/java/org/apache/livy/LivyClientBuilder.java
wypoon: I see that I was mistaken. There are two client factories in use, `HttpClientFactory` and `RSCClientFactory`.
I will revert my last change.

mgaido91: Yes, I think we would rather need a class which extends `ServiceLoad` and makes it thread safe

wypoon: What I decided to do is to load the service providers into a static List that has a thread-safe iterator. Then LivyClientBuilder#build can iterate through the client factories in this List.

wypoon:
@mgaido91 I have updated the PR with a new implementaion and updated the PR description accordingly.

api/src/main/java/org/apache/livy/LivyClientBuilder.java
```diff
@@ -35,6 +37,19 @@
 
   public static final String LIVY_URI_KEY = "livy.uri";
 
+  private static final ServiceLoader<LivyClientFactory> CLIENT_FACTORY_LOADER =
+    ServiceLoader.load(LivyClientFactory.class, classLoader());
+
+  private static List<LivyClientFactory> getLivyClientFactories() {
+    List<LivyClientFactory> factories = new CopyOnWriteArrayList<>();
```
squito: why does this need to be a copyonwriteArrayList?  here, you're touching it only from a single thread.  And it's not modified outside of this method.

wypoon: It is possible that a CopyOnWriteArrayList is overkill. Even though I do not modify the List after it is populated, and perhaps in practice, an ArrayList might do, from what I have read, the behavior of List.iterator() is not specified in general. The behavior of CopyOnWriteArrayList.iterator() is, however, specified. "The returned iterator provides a snapshot of the state of the list when the iterator was constructed. No synchronization is needed while traversing the iterator." It is guaranteed to be thread-safe.

jerryshao:
Is this PR #268 related to this issue, do they fix the same issue?

CC @runzhiwang .

runzhiwang:
@jerryshao They do not fix the same issue. This pr fix the problem of create client concurrently,  and #268 fix the problem of multi client connect to one driver.

api/src/main/java/org/apache/livy/LivyClientBuilder.java
mgaido91: Since you are never modifying the list, there is no need for it to be thread safe. It is safe to read the same List from different threads if it is never modified. So I think an `ArrayList` would be a better option here.

wypoon: Switched to use an ArrayList.

wypoon:
@mgaido91 I hope this can be merged now.

api/src/main/java/org/apache/livy/LivyClientBuilder.java
```diff
@@ -35,6 +37,19 @@
 
   public static final String LIVY_URI_KEY = "livy.uri";
 
+  private static final ServiceLoader<LivyClientFactory> CLIENT_FACTORY_LOADER =
+    ServiceLoader.load(LivyClientFactory.class, classLoader());
+
+  private static List<LivyClientFactory> getLivyClientFactories() {
```
mgaido91: nit: I'd prefer using this method in `build`, with `CLIENT_FACTORY_LOADER` being `null` at the beginning and initialized on the first invocation of this method returning it when initialized (as it is commonly done in singleton). This would enforce that the factories are created only when they needed and they are not created at loading time, which is a very delicate moment in which I prefer to avoid doing things as much as possible.

mgaido91:
Thanks, merging to master.

artparks:
I built Livy from master to try and include this fix, but I still get the same problem of exhausted ports.

PR #283, opened 2020-03-02
[MINOR] Add description of POST /batches

marblejenka:

Just added a description of POST /batches in the api-doc.

## How was this patch tested?

Since the change is quite small, I didn't test the patch. If I have an instruction, I will follow it.


jerryshao:
Thanks for the fix, merging to master branch.

PR #284, opened 2020-03-02
[LIVY-752][THRIFT] Fix implementation of limits on connections.

wypoon:

`LivyThriftSessionManager` keeps a `ConcurrentHashMap[String, AtomicLong]` named `connectionsCount` to track the number of connections per user, etc. The `incrementConnectionsCount` and `decrementConnectionsCount` methods in `LivyThriftSessionManager` check that `connectionsCount` does not contain a key (instead of contains the key) before getting the value and incrementing or decrementing the count (leading to a `NullPointerException`). Even accounting for the incorrect condition, they do not use the `ConcurrentHashMap` correctly. There is a race -- a thread can get a count, find that it's within a limit, create a new session and then increment the count, while in the meantime, another thread could have incremented the count and so the limit is now actually exceeded.

We increment all relevant counts optimistically before creating a new session, check if any limits are violated, and if so, decrement all incremented counts.

## How was this patch tested?

Tested by deploying the change on a cluster and setting livy.server.thrift.limit.connections.per.user. Verified that the number of connections reaches but does not exceed the limit.
Also, added basic unit tests that connection limits are enforced.

### DIFF
--- a/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
+++ b/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
@@ -413,20 +413,6 @@ class LivyThriftSessionManager(val server: LivyThriftServer, val livyConf: LivyC
     }
   }
 
-  // Taken from Hive
-  @throws[HiveSQLException]
-  private def incrementConnections(
-      username: String,
-      ipAddress: String,
-      forwardedAddresses: util.List[String]): Unit = {
-    val clientIpAddress: String = getOriginClientIpAddress(ipAddress, forwardedAddresses)
-    val violation = anyViolations(username, clientIpAddress)
-    if (!violation.isEmpty) {
-      error(violation.get)
-      throw new HiveSQLException(violation.get)
-    }
-  }
-
   private def incrementConnectionsCount(key: String): Long = {
     // Increment the count, returning the new value.
     val count = connectionsCount.putIfAbsent(key, new AtomicLong(1L))
@@ -457,12 +443,17 @@ class LivyThriftSessionManager(val server: LivyThriftServer, val livyConf: LivyC
     }
   }
 
-  private def anyViolations(username: String, ipAddress: String): Option[String] = {
-    // As a side effect, if there are no violations, the connection counts are incremented.
-    val userAndAddress = username + ":" + ipAddress
+  // Visible for testing
+  @throws[HiveSQLException]
+  private[thriftserver] def incrementConnections(
+      username: String,
+      ipAddress: String,
+      forwardedAddresses: util.List[String]): Unit = {
+    val clientIpAddress: String = getOriginClientIpAddress(ipAddress, forwardedAddresses)
+    val userAndAddress = username + ":" + clientIpAddress
     val trackUser = trackConnectionsPerUser(username)
-    val trackIpAddress = trackConnectionsPerIpAddress(ipAddress)
-    val trackUserIpAddress = trackConnectionsPerUserIpAddress(username, ipAddress)
+    val trackIpAddress = trackConnectionsPerIpAddress(clientIpAddress)
+    val trackUserIpAddress = trackConnectionsPerUserIpAddress(username, clientIpAddress)
     var userLimitExceeded = false
     var ipAddressLimitExceeded = false
     var userIpAddressLimitExceeded = false
@@ -473,7 +464,7 @@ class LivyThriftSessionManager(val server: LivyThriftServer, val livyConf: LivyC
       if (userCount > userLimit) userLimitExceeded = true
     }
     if (trackIpAddress) {
-      val ipAddressCount = incrementConnectionsCount(ipAddress)
+      val ipAddressCount = incrementConnectionsCount(clientIpAddress)
       if (ipAddressCount > ipAddressLimit) ipAddressLimitExceeded = true
     }
     if (trackUserIpAddress) {
@@ -485,21 +476,25 @@ class LivyThriftSessionManager(val server: LivyThriftServer, val livyConf: LivyC
     // so decrement all counts that have been incremented.
     if (userLimitExceeded || ipAddressLimitExceeded || userIpAddressLimitExceeded) {
       if (trackUser) decrementConnectionsCount(username)
-      if (trackIpAddress) decrementConnectionsCount(ipAddress)
+      if (trackIpAddress) decrementConnectionsCount(clientIpAddress)
       if (trackUserIpAddress) decrementConnectionsCount(userAndAddress)
     }
 
-    // the violation to return
-    if (userLimitExceeded) {
-      Some(s"Connection limit per user reached (user: $username limit: $userLimit)")
-    } else if (ipAddressLimitExceeded) {
-      Some(s"Connection limit per ipaddress reached (ipaddress: $ipAddress limit: " +
-        s"$ipAddressLimit)")
-    } else if (userIpAddressLimitExceeded) {
-      Some(s"Connection limit per user:ipaddress reached (user:ipaddress: $userAndAddress " +
-        s"limit: $userIpAddressLimit)")
-    } else {
-      None
+    val violation =
+      if (userLimitExceeded) {
+        Some(s"Connection limit per user reached (user: $username limit: $userLimit)")
+      } else if (ipAddressLimitExceeded) {
+        Some(s"Connection limit per ipaddress reached (ipaddress: $clientIpAddress limit: " +
+          s"$ipAddressLimit)")
+      } else if (userIpAddressLimitExceeded) {
+        Some(s"Connection limit per user:ipaddress reached (user:ipaddress: $userAndAddress " +
+          s"limit: $userIpAddressLimit)")
+      } else {
+        None
+      }
+    if (!violation.isEmpty) {
+      error(violation.get)
+      throw new HiveSQLException(violation.get)
     }
   }
 
@@ -517,11 +512,6 @@ class LivyThriftSessionManager(val server: LivyThriftServer, val livyConf: LivyC
   // Taken from Hive
   private def trackConnectionsPerUser(username: String): Boolean = {
     userLimit > 0 && username != null && !username.isEmpty
-  }
-
-  // Taken from Hive
-  private def withinLimits(track: String, limit: Int): Boolean = {
-    !(connectionsCount.containsKey(track) && connectionsCount.get(track).intValue >= limit)
   }
 
   private def decrementConnections(sessionInfo: SessionInfo): Unit = {
--- a/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
+++ b/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
@@ -416,8 +416,11 @@ class LivyThriftSessionManager(val server: LivyThriftServer, val livyConf: LivyC
   private def incrementConnectionsCount(key: String): Long = {
     // Increment the count, returning the new value.
     val count = connectionsCount.putIfAbsent(key, new AtomicLong(1L))
-    if (count != null) count.incrementAndGet()
-    else 1L
+    if (count != null) {
+      count.incrementAndGet()
+    } else {
+      1L
+    }
   }
 
   private def decrementConnectionsCount(key: String): Unit = {
@@ -426,8 +429,11 @@ class LivyThriftSessionManager(val server: LivyThriftServer, val livyConf: LivyC
       override def apply(unused: String, count: AtomicLong): AtomicLong = {
         val countValue = count.decrementAndGet()
         debug(s"Connection count for $key is $countValue")
-        if (countValue == 0L) null
-        else count
+        if (countValue == 0L) {
+          null
+        } else {
+          count
+        }
       }
     }
     connectionsCount.computeIfPresent(key, decrementCount)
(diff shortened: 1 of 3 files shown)


thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
```diff
@@ -21,6 +21,7 @@ import java.lang.reflect.UndeclaredThrowableException
 import java.security.PrivilegedExceptionAction
 import java.util
 import java.util.{Date, Map => JMap, UUID}
+import java.util.function.BiFunction
```
andrasbeni: You will need to move this line after `java.util.concurrent...` imports to maintain alphabetic ordering and satisfy scalastyle.

wypoon: Thanks. Somehow I'd missed that.

squito:
I think the helper `withinLimits()` is no longer used.

without testing for concurrency, could you at least add a basic unit test that the limits are enforced, say on `anyViolations()`?  (given that the logic was entirely backwards before, a unit test seems like it would help here.)

squito:
those things aside, the change itself lgtm

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
```diff
@@ -455,17 +457,45 @@ class LivyThriftSessionManager(val server: LivyThriftServer, val livyConf: LivyC
     }
   }
 
-  // Taken from Hive
   private def anyViolations(username: String, ipAddress: String): Option[String] = {
```
bersprockets: This now has a funny name, because it's responsible for incrementing (and possibly decrementing) counts as part of testing if there are any violations. So it has that side effect.

wypoon:
> I think the helper `withinLimits()` is no longer used.
> 
Thanks for pointing that out. I'll remove it.

> without testing for concurrency, could you at least add a basic unit test that the limits are enforced, say on `anyViolations()`? (given that the logic was entirely backwards before, a unit test seems like it would help here.)

Yes, it would be helpful to have a basic unit test for that.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
bersprockets: Is there a reason not to rename it incrementConnections and move the remaining few lines of the old incrementConnections into here?

I don't feel super strongly about it. Just thought that it might make it more clear.

wypoon: Yes, I suppose I could do that.

wypoon:
@mgaido91 @jerryshao can you please review?

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/TestLivyThriftSessionManager.scala
```diff
@@ -0,0 +1,127 @@
...
+    thriftSessionMgr.incrementConnections(user, "10.20.30.42", forwardedAddresses)
+    thriftSessionMgr.incrementConnections(user, "10.20.30.43", forwardedAddresses)
+
+    thriftSessionMgr.incrementConnections(user, ipAddress, forwardedAddresses)
+    thriftSessionMgr.incrementConnections(user, ipAddress, forwardedAddresses)
+    val msg =
+      s"Connection limit per user:ipaddress reached (user:ipaddress: $userAndAddress limit: 3)"
+    testLimit(thriftSessionMgr, user, ipAddress, forwardedAddresses, msg)
+  }
+
```
mgaido91: nit: would be good to have a test also for the case when there are multiple violations in order to enforce the error message thrown in that condiotion

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
```diff
@@ -412,37 +413,24 @@ class LivyThriftSessionManager(val server: LivyThriftServer, val livyConf: LivyC
...
-        incrementConnectionsCount(username + ":" + clientIpAddress)
-      }
-    } else {
-      error(violation.get)
-      throw new HiveSQLException(violation.get)
-    }
+  private def incrementConnectionsCount(key: String): Long = {
+    // Increment the count, returning the new value.
+    val count = connectionsCount.putIfAbsent(key, new AtomicLong(1L))
+    if (count != null) count.incrementAndGet()
```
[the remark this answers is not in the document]
wypoon: I'm fine to put the braces in for if-else. I didn't because the code this replaces has exactly this kind of if-else (no braces, each clause being a single line), so I thought that was the project style!

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
```diff
@@ -455,21 +443,58 @@ class LivyThriftSessionManager(val server: LivyThriftServer, val livyConf: LivyC
...
+    }
+    if (trackUserIpAddress) {
+      val userIpAddressCount = incrementConnectionsCount(userAndAddress)
+      if (userIpAddressCount > userIpAddressLimit) userIpAddressLimitExceeded = true
+    }
+
+    // If any limit has been exceeded, we won't be going ahead with the connection,
+    // so decrement all counts that have been incremented.
+    if (userLimitExceeded || ipAddressLimitExceeded || userIpAddressLimitExceeded) {
+      if (trackUser) decrementConnectionsCount(username)
```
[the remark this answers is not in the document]
wypoon: I think this is clean and easy to reason about. If I understand you correctly, you are suggesting not incrementing calling incrementConnectionsCount on a key if we already have a limit exceeded so that we don't have to call decrementConnectionsCount on that same key. That leads to more complex logic and is harder to reason about.

E.g.,
```
    // Optimistically increment the counts while getting them to check for violations.
    // If any limit has been exceeded, we won't be going ahead with the connection,
    // so decrement all counts that have been incremented.
    if (trackUser) {
      val userCount = incrementConnectionsCount(username)
      if (userCount > userLimit) {
        userLimitExceeded = true
        decrementConnectionsCount(username)
      }
    }
    if (trackIpAddress && !userLimitExceeded) {
      val ipAddressCount = incrementConnectionsCount(clientIpAddress)
      if (ipAddressCount > ipAddressLimit) {
        ipAddressLimitExceeded = true
        if (trackUser) decrementConnectionsCount(username)
        decrementConnectionsCount(clientIpAddress)
      }
    }
    if (trackUserIpAddress && !userLimitExceeded && !ipAddressLimitExceeded) {
      val userIpAddressCount = incrementConnectionsCount(userAndAddress)
      if (userIpAddressCount > userIpAddressLimit) {
        userIpAddressLimitExceeded = true
        if (trackUser) decrementConnectionsCount(username)
        if (trackIpAddress) decrementConnectionsCount(clientIpAddress)
        decrementConnectionsCount(userAndAddress)
      }
    }
```
This logic is, imo, more complex and more difficult to reason about (and prove is correct!).
Therefore I prefer to stick with the existing version.

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/TestLivyThriftSessionManager.scala
wypoon: I can add a case with multiple violations (the error message will the first violation since that is how the logic goes), but there are a number of permutations, and I don't intend these unit tests to be comprehensive. There were no tests on this area at all, so these ones, even though basic, are already an improvement.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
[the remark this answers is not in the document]
squito: I think marco is suggesting that you include the throw in there as well, which then decreases the complexity of the conditions, and lets you skip the `userLimitExceeded` etc vals and the entire `val violation` section below.  (though you still repeat the `decrementCounts` part

```scala
def errorAndThrow(msg: String): Unit = {
  error(msg)
  throw new HiveSQLException(msg)
}


    if (trackUser) {
      val userCount = incrementConnectionsCount(username)
      if (userCount > userLimit) {
        decrementConnectionsCount(username)
        errorAndThrow(s"Connection limit per user reached (user: $username limit: $userLimit)")        
      }
    }
    if (trackIpAddress) {
      val ipAddressCount = incrementConnectionsCount(clientIpAddress)
      if (ipAddressCount > ipAddressLimit) {
        if (trackUser) decrementConnectionsCount(username)
        decrementConnectionsCount(clientIpAddress)
        errorAndThrow(s"Connection limit per ipaddress reached (ipaddress: $clientIpAddress limit: " +
          s"$ipAddressLimit)")
      }
    }
    if (trackUserIpAddress) {
      val userIpAddressCount = incrementConnectionsCount(userAndAddress)
      if (userIpAddressCount > userIpAddressLimit) {
        if (trackUser) decrementConnectionsCount(username)
        if (trackIpAddress) decrementConnectionsCount(clientIpAddress)
        decrementConnectionsCount(userAndAddress)
        errorAndThrow(s"Connection limit per user:ipaddress reached (user:ipaddress: $userAndAddress " +
          s"limit: $userIpAddressLimit)")
      }
    }
```

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/TestLivyThriftSessionManager.scala
squito: it might also be nice to add a test for closing connections, and opening new ones, to ensure that does *not* trigger violations.

(I agree with your sentiment that you are strictly improving coverage here.  this could be done later, but sometimes its easier to just do it while its fresh in your mind, if its not too hard.)

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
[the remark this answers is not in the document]
wypoon: @squito
Thanks for the suggestion, which lets us skip the violation portion (as well as remove the need for the vars). That leaves us still with the core that I'd outlined, and the logic of that core still seems harder to reason about (prove is correct) than the previous approach. You have repeated this core, so I assume you agree that it is correct:
```
    if (trackUser) {
      // 1
    }
    if (trackIpAddress) {
      // 2
    }
    if (trackUserIpAddress) {
      // 3
    }
```
If we got to 2, we know that we may have incremented the count for user, so if there's a violation now, we may need to decrement the count for user as well as for clientIpAddress.
If we got to 3, we know that we may have incremented the count for user and for clientIpAddress, so if there's a violation now, we may need to decrement the count for those as well as for userAndAddress.

That's the reasoning that is more complex.
But if you agree that the logic is correct, I'm ok to do it this way.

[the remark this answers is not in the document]
squito: honestly I see it both ways -- I was mostly just trying to clarify marco's comment a bit to keep things moving.   I'm fine either way.   its @mgaido91 's call.

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/TestLivyThriftSessionManager.scala
mgaido91: yes, thanks @squito , I agree the current status of this PR is already an improvement, my comment was a nit, but I think it is a nice to have. Since I don't think it is much effort to add it, I think it would be great to have it.

thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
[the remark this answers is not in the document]
mgaido91: yes, thanks for the clarification @squito , that was what I meant. Honestly, I don't see a big difference in terms of complexity/readability and in one case we avoid some sync operations, so it is more efficient: hence I prefer that option. This part is not critical anyway IMHO, so it is not a big deal, but if you're not strongly against it, I'd go this way. Thanks.

wypoon:
I adopted Imran's suggestion for rewriting `incrementConnections`.
I also added a test case for multiple connection limits being violated; the error message is for the first limit that we check.
I am passing on additional nice-to-have tests. There is always room for improvement for the tests. This can be an area for further contributions, possibly from new contributors!

thriftserver/server/src/test/scala/org/apache/livy/thriftserver/TestLivyThriftSessionManager.scala
```diff
@@ -0,0 +1,145 @@
...
+class TestLivyThriftSessionManager {
+
+  import ConnectionLimitType._
+
+  private def createThriftSessionManager(
+      limitTypes: ConnectionLimitType*): LivyThriftSessionManager = {
+    val conf = new LivyConf()
+    conf.set(LivyConf.LIVY_SPARK_VERSION, sys.env("LIVY_SPARK_VERSION"))
+    val limit = 3
+    for (limitType <- limitTypes) {
```
mgaido91: nit:
```suggestion
   limitTypes.foreach { limitType =>
```

wypoon:
@mgaido91 can you merge this now? Thanks.

mgaido91:
Merged to master. Thanks.

PR #289, opened 2020-04-03
[LIVY-756] Add Support Spark 3

tprelle:

Add spark 3.0 support to livy with scala 2.12 and Python 3
Fix also https://issues.apache.org/jira/browse/LIVY-423 

## How was this patch tested?
Scala, Python 2 and 3 test work
I'm not able yet to make sparkR work for spark 3 yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,12 @@ matrix:
     env: MVN_FLAG='-Pspark-2.4 -Pthriftserver -DskipITs'
   - name: "Spark 2.4 ITs"
     env: MVN_FLAG='-Pspark-2.4 -Pthriftserver -DskipTests'
+  - name: "Spark 3.0 Unit Tests"
+    env: MVN_FLAG='-Pthriftserver -Pspark-3.0 -DskipITs'
+  - name: "Spark 3.0 ITs"
+    env: 
+      - MVN_FLAG='-Pthriftserver -Pspark-3.0 -DskipTests'
+      - PYSPARK_ROW_FIELD_SORTING_ENABLED=true
 
 jdk:
   - oraclejdk8
--- a/assembly/pom.xml
+++ b/assembly/pom.xml
@@ -29,7 +29,7 @@
   <packaging>pom</packaging>
 
   <properties>
-    <assembly.name>apache-livy-${project.version}-bin</assembly.name>
+    <assembly.name>apache-livy-${project.version}-bin_${scala.binary.version}</assembly.name>
     <assembly.format>zip</assembly.format>
     <skipDeploy>true</skipDeploy>
   </properties>
@@ -43,7 +43,7 @@
 
     <dependency>
       <groupId>${project.groupId}</groupId>
-      <artifactId>livy-repl_2.11</artifactId>
+      <artifactId>livy-repl_${scala.binary.version}</artifactId>
       <version>${project.version}</version>
     </dependency>
 
@@ -51,6 +51,79 @@
       <groupId>${project.groupId}</groupId>
       <artifactId>livy-server</artifactId>
       <version>${project.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>${project.groupId}</groupId>
+          <artifactId>livy-core_2.11</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>${project.groupId}</groupId>
+          <artifactId>livy-core_2.12</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.json4s</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.scalatra</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>com.fasterxml.jackson.module</groupId>
+          <artifactId>jackson-module-scala_2.11</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>com.fasterxml.jackson.module</groupId>
+          <artifactId>jackson-module-scala_2.12</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>livy-core_${scala.binary.version}</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>org.json4s</groupId>
+      <artifactId>json4s-ast_${scala.binary.version}</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.json4s</groupId>
+      <artifactId>json4s-core_${scala.binary.version}</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.json4s</groupId>
+      <artifactId>json4s-jackson_${scala.binary.version}</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.scalatra</groupId>
+      <artifactId>scalatra_${scala.binary.version}</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.scalatra</groupId>
+      <artifactId>scalatra-json_${scala.binary.version}</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.scalatra</groupId>
+      <artifactId>scalatra-metrics_${scala.binary.version}</artifactId>
+      <version>${scalatra.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>com.typesafe.akka</groupId>
+          <artifactId>akka-actor_${scala.binary.version}</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+
+    <dependency>
+      <groupId>com.fasterxml.jackson.module</groupId>
+      <artifactId>jackson-module-scala_${scala.binary.version}</artifactId>
     </dependency>
   </dependencies>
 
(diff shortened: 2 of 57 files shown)


api/src/main/java/org/apache/livy/JobContext.java
```diff
@@ -41,12 +40,12 @@
...
    * @return The shared SQLContext instance.
    */
-  SQLContext sqlctx();
+  SQLContext sqlctx() throws Exception ;
 
   /**
-   * @return The shared HiveContext instance.
+   * @return The shared hive SQLContext instance.
    */
-  HiveContext hivectx();
```
andrasbeni, line 49: This will break compatibility to those who keep using Livy with Spark 2.4.
I'd recommend to maintain two versions of this file and adding the correct version to compile sources based on profiles.

tprelle: Hi, @andrasbeni thanks for the review, I do the code in december with spark 3.0.0-preview2, I just check on spark and they push back they class in https://issues.apache.org/jira/browse/SPARK-31088, I will make the change back when they will released spark 3.0.0

Minutis:
Spark 3 was released. https://spark.apache.org/downloads.html, please, continue :+1:

assembly/pom.xml
```diff
@@ -29,7 +29,7 @@
   <packaging>pom</packaging>
 
   <properties>
-    <assembly.name>apache-livy-${project.version}-bin</assembly.name>
+    <assembly.name>apache-livy-${project.version}-bin-${scala.binary.version}</assembly.name>
```
zjffdu, line 32: to  `_${scala.binary.version}` ?

tprelle:
> Spark 3 was released. https://spark.apache.org/downloads.html, please, continue 👍

Move to spark 3.0.0.
Do not have time yet to test spark R

jerryshao:
@tprelle is this PR ready for review?

jerryshao:
I'm going to try this PR locally to see if the current code is complete or not.

tprelle:
@jerryshao the code it's working for spark scala and python but I do not managed to make it work for sparkR.

I can submit the PR and open a ticket for sparkR support for spark3 if it's ok for the community

jerryshao:
Yes, I think it is fine to leave R support to another PR. Please make sure this PR is OK for Scala and Python part.

tprelle:
All unit and intregration scala and Python test are working.
I'm currently redeploy it to make more integration test because of https://issues.apache.org/jira/browse/SPARK-29748 (the biggest change I spot for livy between 3.0.0-preview2 version en 3.0.0 version.

jerryshao:
I just did a quick review about the overall implementation. Seems like we can only support one Scala version for one build here  in this PR. But Livy could actually support different versions of Scala in one build (we used to support Scala 2.10 and 2.11 for one build). So I think we should have both Scala 2.12 and 2.11 support, and let Livy to choose jars automatically based on the Spark.

client-http/pom.xml
```diff
@@ -109,58 +109,6 @@
 
   <build>
     <plugins>
-      <plugin>
```
jerryshao, line 112: What's the reason to remove this?

.travis.yml
```diff
@@ -34,6 +34,12 @@ matrix:
     env: MVN_FLAG='-Pspark-2.4 -Pthriftserver -DskipITs'
   - name: "Spark 2.4 ITs"
     env: MVN_FLAG='-Pspark-2.4 -Pthriftserver -DskipTests'
+  - name: "Spark 3.0 Unit Tests"
+    env: MVN_FLAG='-Pthriftserver -Pspark-3.0 -DskipITs'
+  - name: "Spark 3.0 ITs"
+    env: 
+      - MVN_FLAG='-Pthriftserver -Pspark-3.0 -DskipTests'
+      - PYSPARK_ROW_FIELD_SORTING_ENABLED=true
```
jerryshao, line 42: What is the purpose of setting this?

jerryshao: Got it, seems an incompatible change in Spark 3.0

pom.xml
```diff
@@ -86,24 +87,25 @@
...
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <py4j.version>0.10.7</py4j.version>
     <scala-2.11.version>2.11.12</scala-2.11.version>
+    <scala-2.12.version>2.12.10</scala-2.12.version>
     <scala.binary.version>2.11</scala.binary.version>
     <scala.version>${scala-2.11.version}</scala.version>
-    <scalatest.version>2.2.4</scalatest.version>
-    <scalatra.version>2.3.0</scalatra.version>
+    <scalatest.version>3.0.8</scalatest.version>
+    <scalatra.version>2.6.5</scalatra.version>
```
jerryshao, line 108: Are this artifacts upgrading necessary for Spark 3.0 support?

andrasbeni: Scalatra added scala 2.12 support in version 2.5.0 and scalatest in 3.0.0. So the version currently in use are incompatibel with Spark 3.0's Scala 2.12.

api/src/main/java/org/apache/livy/JobContext.java
```diff
@@ -41,12 +41,12 @@
   /**
    * @return The shared SQLContext instance.
    */
-  SQLContext sqlctx();
+  SQLContext sqlctx() throws Exception ;
```
andrasbeni, line 44: I believe adding throws clauses to these methods could break Java client applications. I suggest to wrap exceptions that are not RuntimeException in a RuntimeException instead.

pom.xml
jerryshao: Scalatra is only used by Livy Server, we don't have to upgrade Livy Server's Scala version to 2.12, we could leave the Livy Server Scala support to another JIRA.

api/src/main/java/org/apache/livy/JobContext.java
jerryshao: Yes, I have the same concern about changing the signature of this interface.

jerryshao:
I'm doing some local improvements based on @tprelle 's current PR, I can submit a PR when everything is ready.

api/src/main/java/org/apache/livy/JobContext.java
tprelle: It was before https://issues.apache.org/jira/browse/SPARK-31088, when they put back HiveContext class back to spark.
When I revert the change I forgot to revert also the method signature.

client-http/pom.xml
tprelle: It was a mistake, i will put it back

tprelle:
> I just did a quick review about the overall implementation. Seems like we can only support one Scala version for one build here in this PR. But Livy could actually support different versions of Scala in one build (we used to support Scala 2.10 and 2.11 for one build). So I think we should have both Scala 2.12 and 2.11 support, and let Livy to choose jars automatically based on the Spark.

I like the idea, I try it, but I was not able to make it work.

pom.xml
tprelle: I try to limit the number of version change, but because of scala 2.12 like @andrasbeni says I need to use scalatest 3.0.8 who are some breaking change like  org.scalatest.Suite =­>org.scalatest.TestSuite, so to do not copy the class I need to change also the global version of scalatest. So I need to change the version of scalatra because of the incompatibility between scalatest 3.0.8 and scalatra-scalatest 2.3.0. 
As I was working and easier to move forward with these upgrade of version I choose to submit like this.

mmigdiso:
Hey @tprelle , will @jerryshao comments be a blocker for that PR?

scala-api/src/main/scala/org/apache/livy/scalaapi/ScalaJobHandle.scala
```diff
@@ -190,6 +190,19 @@ class ScalaJobHandle[T] private[livy] (jobHandle: JobHandle[T]) extends Future[T
     getJavaFutureResult(jobHandle, atMost)
     this
   }
+
+  // These two methods must be implemented in Scala 2.12. They're implemented as a no-op here
```
jerryshao, line 194: Would you please elaborate why the below two methods are needed for Scala 2.12?

andrasbeni: In Scala 2.12 these two new methods ([see source](https://github.com/scala/scala/blob/v2.12.10/src/library/scala/concurrent/Future.scala#L248-L268 )) were added to the trait `Future`.  As a result, `ScalaJobHandle`, which extends Future needs to implement them.

At the same time I am not sure what @tprelle means by "the two subclasses below". AFAICT `ScalaJobHandle` does not have any subclasses. 
Also, I wouldn't say throwing an exception is a no-op. I suggest "They are implemented as unsupported operations here" if they are really not going to be [implemented](https://github.com/scala/scala/blob/v2.12.10/src/library/scala/concurrent/impl/Promise.scala#L31-L48).

jerryshao: Yes, the comments seems confusing.

jerryshao:
Thanks @tprelle for your work, I've already merged this PR, all credits to you.

stczwd:
Glad to see this patch merged. I have test this path, it works well with spark3.0.

giftkugel:
Is there any timeline to add this patch into a release? Maybe Livy 0.8?

jerryshao:
There still have some minor issues in JDBC side, we prepare a release when all the issues are done.

hongweijia:
Hi @tprelle
Has livy already supported Spark 3 with Python 3.6?
Thanks!

tprelle:
Hi @hongweijia, never test it with Python 3.6 but it's working well with 3.7,

Regards

PR #290, opened 2020-04-16
[LIVY-758] Document how to attach to an existing session from Java client

tmnd1991:

Add an explicit way to set the session when connecting from the Java client.
https://issues.apache.org/jira/browse/LIVY-758

## How was this patch tested?

Unit tested through `HttpClientSpec`.

### DIFF
--- a/api/src/main/java/org/apache/livy/LivyClientBuilder.java
+++ b/api/src/main/java/org/apache/livy/LivyClientBuilder.java
@@ -36,6 +36,7 @@ import static java.nio.charset.StandardCharsets.UTF_8;
 public final class LivyClientBuilder {
 
   public static final String LIVY_URI_KEY = "livy.uri";
+  public static final String LIVY_SESSION_ID_KEY = "livy.sessionId";
 
   private static final ServiceLoader<LivyClientFactory> CLIENT_FACTORY_LOADER =
     ServiceLoader.load(LivyClientFactory.class, classLoader());
@@ -96,8 +97,30 @@ public final class LivyClientBuilder {
     }
   }
 
+  /**
+   * Set the URI of the Livy server the client will connect to.
+   * If the URI contains <pre>sessions/{sessionId}</pre>,
+   * the client will connect to the specified existing session,
+   * otherwise it will create a new session.
+   *
+   * @param uri The URI of Livy server.
+   * @return The builder itself.
+   */
   public LivyClientBuilder setURI(URI uri) {
     config.setProperty(LIVY_URI_KEY, uri.toString());
+    return this;
+  }
+
+  /**
+   * Sets the sessionId the client will connect to. If sessionId is set,
+   * all Spark configurations will be ignored and the original session ones will be used.
+   * If not set, a new session will be created when the client is built.
+   *
+   * @param sessionId The ID of the session to attach to.
+   * @return the builder itself.
+   */
+  public LivyClientBuilder setSessionId(int sessionId) {
+    config.setProperty(LIVY_SESSION_ID_KEY, String.valueOf(sessionId));
     return this;
   }
 
--- a/api/src/main/java/org/apache/livy/LivyClientBuilder.java
+++ b/api/src/main/java/org/apache/livy/LivyClientBuilder.java
@@ -98,9 +98,8 @@ public final class LivyClientBuilder {
   }
 
   /**
-   * Sets the URI of the Livy server the client will connect to.
-   * If the URI contains <pre>sessions/{sessionId}</pre>,
-   * the client will connect to the specified existing session;
+   * Sets the URI of the Livy server the client will connect to. If the URI contains
+   * <pre>sessions/{sessionId}</pre>, the client will connect to the specified existing session;
    * otherwise it will create a new session.
    *
    * @param uri The URI of Livy server.
@@ -112,9 +111,9 @@ public final class LivyClientBuilder {
   }
 
   /**
-   * Sets the session ID the client will connect to. If a session ID is set,
-   * the chosen session will be used with its own configurations, so Spark configurations will be ignored.
-   * If not set, a new session will be created when the client is built.
+   * Sets the session ID the client will connect to. If a session ID is set, the chosen session
+   * will be used with its own configurations, so Spark configurations will be ignored. If not set,
+   * a new session will be created when the client is built.
    *
    * @param sessionId The ID of the session to attach to.
    * @return the builder itself.
--- a/api/src/main/java/org/apache/livy/LivyClientBuilder.java
+++ b/api/src/main/java/org/apache/livy/LivyClientBuilder.java
@@ -100,7 +100,7 @@ public final class LivyClientBuilder {
   /**
    * Set the URI of the Livy server the client will connect to.
    * If the URI contains <pre>sessions/{sessionId}</pre>,
-   * the client will connect to the specified existing session,
+   * the client will connect to the specified existing session;
    * otherwise it will create a new session.
    *
    * @param uri The URI of Livy server.
--- a/api/src/main/java/org/apache/livy/LivyClientBuilder.java
+++ b/api/src/main/java/org/apache/livy/LivyClientBuilder.java
@@ -112,7 +112,7 @@ public final class LivyClientBuilder {
   }
 
   /**
-   * Sets the sessionId the client will connect to. If sessionId is set,
+   * Sets the session ID the client will connect to. If a session ID is set,
    * the chosen session will be used with its own configurations, so Spark configurations will be ignored.
    * If not set, a new session will be created when the client is built.
    *
--- a/api/src/main/java/org/apache/livy/LivyClientBuilder.java
+++ b/api/src/main/java/org/apache/livy/LivyClientBuilder.java
@@ -98,7 +98,7 @@ public final class LivyClientBuilder {
   }
 
   /**
-   * Set the URI of the Livy server the client will connect to.
+   * Sets the URI of the Livy server the client will connect to.
    * If the URI contains <pre>sessions/{sessionId}</pre>,
    * the client will connect to the specified existing session;
    * otherwise it will create a new session.
--- a/api/src/main/java/org/apache/livy/LivyClientBuilder.java
+++ b/api/src/main/java/org/apache/livy/LivyClientBuilder.java
@@ -113,7 +113,7 @@ public final class LivyClientBuilder {
 
   /**
    * Sets the sessionId the client will connect to. If sessionId is set,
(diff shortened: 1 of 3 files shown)


tmnd1991:
I could not assign myself the relate jira, how can I do it? @mgaido91 @vanzin @jerryshao

mgaido91:
@tmnd1991 you can't, JIRAs are assigned when related patches are merged and they are resolved

api/src/main/java/org/apache/livy/LivyClientBuilder.java
```diff
@@ -96,11 +97,30 @@ public LivyClientBuilder(boolean loadDefaults) throws IOException {
     }
   }
 
+  /**
+   *
+   * @param uri the uri of the livy server,
+   *            if the uri contains <pre>sessions/{sessionId}</pre>
+   *            the client will connect to the existing session,
+   *            otherwise it will create a new session.
```
mgaido91, line 105:
```suggestion
   * @param uri The URI of Livy server. If the URI contains <pre>sessions/{sessionId}</pre>,
   *            the client will connect to the specified existing session, otherwise it will create a
   *            new session.
```

api/src/main/java/org/apache/livy/LivyClientBuilder.java
```diff
@@ -96,11 +97,30 @@ public LivyClientBuilder(boolean loadDefaults) throws IOException {
...
+   * @return the builder itself.
```
mgaido91, line 106:
```suggestion
   * @return The builder itself.
```

api/src/main/java/org/apache/livy/LivyClientBuilder.java
```diff
@@ -96,11 +97,30 @@ public LivyClientBuilder(boolean loadDefaults) throws IOException {
...
+   */
   public LivyClientBuilder setURI(URI uri) {
     config.setProperty(LIVY_URI_KEY, uri.toString());
     return this;
   }
 
+  /**
+   *
```
mgaido91, line 114: ditto

api/src/main/java/org/apache/livy/LivyClientBuilder.java
```diff
@@ -96,11 +97,30 @@ public LivyClientBuilder(boolean loadDefaults) throws IOException {
...
+   * @param sessionId the id of the session to attach to,
+   *                  if not set a new session will be created when the client is built.
```
mgaido91, line 116:
```suggestion
   * @param sessionId The ID of the session to attach to. If not set, a new
   *                  session will be created when the client is built.
```

api/src/main/java/org/apache/livy/LivyClientBuilder.java
```diff
@@ -96,11 +97,30 @@ public LivyClientBuilder(boolean loadDefaults) throws IOException {
...
+   * @return the builder itself.
+   */
+  public LivyClientBuilder setSessionId(int sessionId) {
+    config.setProperty(LIVY_SESSION_ID_KEY, "" + sessionId);
```
mgaido91, line 120:
```suggestion
    config.setProperty(LIVY_SESSION_ID_KEY, String.valueOf(sessionId));
```

client-http/src/main/java/org/apache/livy/client/http/HttpClient.java
```diff
@@ -54,17 +55,18 @@
     this.config = httpConf;
     this.stopped = false;
 
-    // If the given URI looks like it refers to an existing session, then try to connect to
-    // an existing session. Note this means that any Spark configuration in httpConf will be
-    // unused.
-    Matcher m = Pattern.compile("(.*)" + LivyConnection.SESSIONS_URI + "/([0-9]+)")
-      .matcher(uri.getPath());
-
+    Matcher m = Pattern.compile("(.*)" + LivyConnection.SESSIONS_URI + "/([0-9]+)").matcher(uri.getPath());
```
mgaido91, line 58: why did you remove the comment above? moreover, here I see no changes, so please revert the change to keep the history clean

client-http/src/main/java/org/apache/livy/client/http/HttpClient.java
```diff
@@ -54,17 +55,18 @@
...
+        throw new IllegalArgumentException("Cannot set existing session both from URI and configuration");
+      } else if (sessionIdFromConf != null) {
+        this.conn = new LivyConnection(uri, httpConf);
+        this.sessionId = Integer.parseInt(sessionIdFromConf);
+        conn.post(null, SessionInfo.class, "/%d/connect", sessionId);
+      } else if (m.matches()) {
         URI base = new URI(uri.getScheme(), uri.getUserInfo(), uri.getHost(), uri.getPort(),
-          m.group(1), uri.getQuery(), uri.getFragment());
-
+            m.group(1), uri.getQuery(), uri.getFragment());
```
mgaido91, line 69: please revert unneeded change

tmnd1991:
> @tmnd1991 you can't, JIRAs are assigned when related patches are merged and they are resolved

Then I think [contributing section](https://livy.apache.org/community/) is wrong :)

client-http/src/main/java/org/apache/livy/client/http/HttpClient.java
tmnd1991: I did it because the in-line comment was the *only* doc about connecting to an existing session. Now that is documented in the builder I don't think the inline comment is still needed, anyway no problem at all at reverting that change.

tmnd1991:
I changed what you suggested @mgaido91 , thanks for the review.
Tagging @ajbozarth as he asked on Jira :)

ajbozarth:
@mgaido91 I think your mixing up how Spark handles JIRA assignments with Livy's policy. We assign JIRAs once a PR is open for the issue. For some reason though, I can't assign issues to @tmnd1991 in the Livy JIRA at the moment. We had this issue before but I can't remember how we solved it.

@tmnd1991 A heads up, when you force push updates to commits it makes it difficult to track your changes as you address review (eg. I can't tell which of @mgaido91 comments you've addressed because there's no history to compare to). Since we squash all commits in a PR before merging the quantity of commits in a PR isn't an issue.

As for your code, I read through it and I don't see any issues with the current code, but I don't have to time check it out and more thoroughly test it so I will defer to @mgaido91 on final approval for this PR.

tmnd1991:
> @tmnd1991 A heads up, when you force push updates to commits it makes it difficult to track your changes as you address review (eg. I can't tell which of @mgaido91 comments you've addressed because there's no history to compare to). Since we squash all commits in a PR before merging the quantity of commits in a PR isn't an issue.

Good, I did not know if the squash before merge was a "thing" in Livy repo, I won't do the same mistake again :)

> As for your code, I read through it and I don't see any issues with the current code, but I don't have to time check it out and more thoroughly test it so I will defer to @mgaido91 on final approval for this PR.

👍

mgaido91:
> @mgaido91 I think your mixing up how Spark handles JIRA assignments with Livy's policy. We assign JIRAs once a PR is open for the issue. For some reason though, I can't assign issues to @tmnd1991 in the Livy JIRA at the moment. We had this issue before but I can't remember how we solved it.

Sorry, I saw the same pattern being followed in Livy too in practice so I assumed it was the same. I think the point is that @tmnd1991 has to be added to the contributors in JIRA. I'll do that when resolving the JIRA.

api/src/main/java/org/apache/livy/LivyClientBuilder.java
```diff
@@ -96,11 +97,33 @@ public LivyClientBuilder(boolean loadDefaults) throws IOException {
     }
   }
 
+  /**
+   * Set the URI of the Livy server the client will connect to.
```
mgaido91, line 101:
```suggestion
   * Sets the URI of the Livy server the client will connect to.
```

api/src/main/java/org/apache/livy/LivyClientBuilder.java
```diff
@@ -96,11 +97,33 @@ public LivyClientBuilder(boolean loadDefaults) throws IOException {
...
+   * If the URI contains <pre>sessions/{sessionId}</pre>,
+   * the client will connect to the specified existing session,
```
mgaido91, line 103:
```suggestion
   * the client will connect to the specified existing session;
```

api/src/main/java/org/apache/livy/LivyClientBuilder.java
```diff
@@ -96,11 +97,33 @@ public LivyClientBuilder(boolean loadDefaults) throws IOException {
...
+   * @param uri The URI of Livy server.
+   * @return The builder itself.
+   */
   public LivyClientBuilder setURI(URI uri) {
     config.setProperty(LIVY_URI_KEY, uri.toString());
     return this;
   }
 
+  /**
+   * Sets the sessionId the client will connect to. If sessionId is set,
```
mgaido91, line 115:
```suggestion
   * Sets the session ID the client will connect to. If a session ID is set,
```

api/src/main/java/org/apache/livy/LivyClientBuilder.java
```diff
@@ -96,11 +97,33 @@ public LivyClientBuilder(boolean loadDefaults) throws IOException {
...
+   * all Spark configurations will be ignored and the original session ones will be used.
```
mgaido91, line 116:
```suggestion
   * the chosen session will be used with its own configurations, so Spark configurations will be ignored.
```

mgaido91 (commented):
LGTM apart from some wordings.

api/src/main/java/org/apache/livy/LivyClientBuilder.java
```diff
@@ -96,11 +97,32 @@ public LivyClientBuilder(boolean loadDefaults) throws IOException {
     }
   }
 
+  /**
+   * Sets the URI of the Livy server the client will connect to. If the URI contains
+   * <pre>sessions/{sessionId}</pre>, the client will connect to the specified existing session;
```
mgaido91, line 102: one last question: @ajbozarth @jerryshao @tmnd1991  I think that by adding this API, we might want to deprecate this "hacky" approach and eventually remove the support to it in next major release. Any thought on this?

tmnd1991: I 100% agree on removing the hacky (and until now not even documented) approach. I left it to make the PR backward compatible

mgaido91: yes, we definitely need to leave it for now, but we can start deprecating it. Let's wait for others opinion .

PR #292, opened 2020-04-24
[LIVY-764] Give code block a filename and cache it in linecache

mzhang-code:
[LIVY-764] Give code block a filename and cache it in linecache

Today python shell (`fake_shell.py`) uses 'stdin' as file name to compile code blocks and never caches them in linecache. This breaks the user code that depends on `inspect.findsource`, like `torch.jit.script`

To enable code block caching at the backend, we can use the way how ipython does: https://github.com/ipython/ipython/blob/44bd47d5ac27805a297798d39ad7e31f63181468/IPython/core/compilerop.py#L54

## How was this patch tested?

E2E test with local livy server

### DIFF
--- a/repl/src/main/resources/fake_shell.py
+++ b/repl/src/main/resources/fake_shell.py
@@ -20,8 +20,10 @@ import ast
 from collections import OrderedDict
 import datetime
 import decimal
+import hashlib
 import io
 import json
+import linecache
 import logging
 import sys
 import traceback
@@ -30,6 +32,7 @@ import os
 import re
 import threading
 import tempfile
+import time
 import shutil
 import pickle
 import textwrap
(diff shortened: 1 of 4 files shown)


mzhang-code:
This change enables `inspect.findsource` and `torch.jit.script`

image

image

repl/src/main/resources/fake_shell.py
```diff
@@ -212,20 +215,35 @@ def __init__(self, exc_info):
...
     def __init__(self, code):
-        self.code = compile(code, '<stdin>', 'exec', ast.PyCF_ONLY_AST, 1)
+        self.cell_name = self.code_name(code)
+        linecache_entry = (len(code), time.time(), [line+'\n' for line in code.splitlines()], self.cell_name)
+        linecache.cache[self.cell_name] = linecache_entry
+        linecache._livy_cache[self.cell_name] = linecache_entry
+
+        self.code = compile(code, self.cell_name, 'exec', ast.PyCF_ONLY_AST, 1)
+
+    def code_name(self, code, number=0):
```
vkuzmin-uber, line 225: nit: It seems like this doesn't need self and util method that should be static and maybe private ?

```
@staticmethod
def _code_name(code, number=0):
```

mzhang-code: Makes sense. Will fix it together with unit test fix

vkuzmin-uber (approved):

PR #299, opened 2020-06-23
LIVY-776 - Update Maven Shade Plugin to 3.2.1

coheigea:
In order to apply the PR to update Jackson to 2.10.1 (https://github.com/apache/incubator-livy/pull/258/files), it is also necessary to update the Maven Shade Plugin to avoid an ASM IllegalArgumentException error in the client-http module.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -661,7 +661,7 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-shade-plugin</artifactId>
-          <version>2.4.2</version>
+          <version>3.2.1</version>
         </plugin>
 
         <plugin>


tprelle:
Hi @coheigea  I faced the same issue (shaded plugin and jackson version) on https://github.com/apache/incubator-livy/pull/289 for scala 2.12 and spark 3 support. I need also to change this. I do not know how deal with multiple PR.

coheigea:
@tprelle It's up to the maintainers of the project, either it could all be done in one go (as in your PR), or maybe they might prefer to just update Jackson separately. If your PR is merged first, then I'll just close this PR. Either way it would be good from my POV to get us updated to Jackson 2.10.x ;-)

jerryshao:
Fixed in #300

PR #302, opened 2020-08-11
Add html escape to session name

mgaido91:

The PR adds HTML escaping to session names.

## How was this patch tested?

Manual test.


mgaido91:
cc @ajbozarth

PR #303, opened 2020-08-14
Add app state to AppInfo

AFFogarty:

This PR adds app state to `AppInfo`.  This change gives clients more power to understand the state of their application and to debug any failures.

For example, there is currently no way to differentiate between interactive sessions that have gone into YARN state `FAILED` or YARN state `FINISHED`. Livy reports both of these as `dead`.  With this new change, a client could inspect the `appState` to see the true state of the YARN application.

## How was this patch tested?

This change includes unit tests.

CC: @rapoth @imback82

### DIFF
--- a/server/src/main/scala/org/apache/livy/utils/SparkApp.scala
+++ b/server/src/main/scala/org/apache/livy/utils/SparkApp.scala
@@ -19,7 +19,7 @@ package org.apache.livy.utils
 
 import scala.collection.JavaConverters._
 
-import com.fasterxml.jackson.annotation.JsonProperty
+import com.fasterxml.jackson.module.scala.JsonScalaEnumeration
 
 import org.apache.livy.LivyConf
 
@@ -32,7 +32,7 @@ object AppInfo {
 case class AppInfo(
     var driverLogUrl: Option[String] = None,
     var sparkUiUrl: Option[String] = None,
-    @JsonProperty
+    @JsonScalaEnumeration(classOf[SparkApp.State])
     var appState: Option[SparkApp.State] = None) {
   import AppInfo._
   def asJavaMap: java.util.Map[String, String] =
--- a/server/src/main/scala/org/apache/livy/utils/SparkApp.scala
+++ b/server/src/main/scala/org/apache/livy/utils/SparkApp.scala
@@ -30,7 +30,7 @@ object AppInfo {
 case class AppInfo(
     var driverLogUrl: Option[String] = None,
     var sparkUiUrl: Option[String] = None,
-    var appState: Option[SparkApp.State]) {
+    var appState: Option[SparkApp.State] = None) {
   import AppInfo._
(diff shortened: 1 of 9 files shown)


server/src/main/scala/org/apache/livy/utils/SparkApp.scala
```diff
@@ -19,17 +19,28 @@ package org.apache.livy.utils
...
+    var driverLogUrl: Option[String] = None,
+    var sparkUiUrl: Option[String] = None,
+    @JsonScalaEnumeration(classOf[StateTypeReference])
+    var appState: Option[SparkApp.State] = None) {
   import AppInfo._
   def asJavaMap: java.util.Map[String, String] =
-    Map(DRIVER_LOG_URL_NAME -> driverLogUrl.orNull, SPARK_UI_URL_NAME -> sparkUiUrl.orNull).asJava
+    Map(DRIVER_LOG_URL_NAME -> driverLogUrl.orNull,
+      SPARK_UI_URL_NAME -> sparkUiUrl.orNull,
+      APP_STATE_NAME -> appState.map(s => s.toString).orNull).asJava
```
imback82, line 43: nit: `appState.map(_.toString)`

server/src/test/scala/org/apache/livy/server/interactive/InteractiveSessionServletSpec.scala
```diff
@@ -168,7 +168,10 @@ class InteractiveSessionServletSpec extends BaseInteractiveServletSpec {
     val proxyUser = "proxyUser"
     val state = SessionState.Running
     val kind = Spark
-    val appInfo = AppInfo(Some("DRIVER LOG URL"), Some("SPARK UI URL"))
+    val appInfo = AppInfo(
+      Some("DRIVER LOG URL"),
+      Some("SPARK UI URL"),
+      Some(SparkApp.State.RUNNING))
```
imback82, line 174: can we also test `None` case?

AFFogarty: Added verification that `appState=None` gets serialized to `"appState": null`.

imback82:
@jerryshao Could you help reviewing this PR? Thanks in advance!

server/src/main/scala/org/apache/livy/utils/SparkApp.scala
AFFogarty: Fixed

andrasbeni:
@AFFogarty, I'm curious what you think of the following alternatives:
- Mapping SparkApp.State.FAILED to SessionState.ERROR()
- Introducing SessionState.FAILED()
I'm not too familiar with the state transitions, so above ideas may be dumb. In which case I'm happy to learn why.

rapoth:
@jerryshao Any chance you could review this PR? Thanks a lot for your time in advance!

PR #307, opened 2020-09-10
[LIVY-786] Fix cancellation of driver side jobs

namanmishra91:
Currently, cancelling a statement on an interactive session only cancels the jobs running on Spark executors but anything running on the Driver keeps running and keeps the session busy. This can be seen by running a sleep in code and running something immediately after cancelling it. This change fixes the cancellation of such tasks for Scala and Python interpreters.

https://issues.apache.org/jira/projects/LIVY/issues/LIVY-786

## How was this patch tested?
Tested manually on local setup. Sequence of steps followed:
For Python:
1. curl localhost:8998/sessions/0/statements -X POST -H 'Content-Type: application/json' -d '{"code":"import time\ntime.sleep(60)"}' 
2. curl -X POST localhost:8998/sessions/0/statements/0/cancel
3. curl localhost:8998/sessions/0/statements -X POST -H 'Content-Type: application/json' -d '{"code":"print(\"hello\")"}'

The result of second statement was available immediately, meaning that the repl was free.

For Scala:
1. curl localhost:8998/sessions/0/statements -X POST -H 'Content-Type: application/json' -d '{"code":"Thread.sleep(60000)", "kind":"scala"}'
2. curl -X POST localhost:8998/sessions/0/statements/2/cancel
3. curl localhost:8998/sessions/0/statements -X POST -H 'Content-Type: application/json' -d '{"code":"val a=1", "kind":"scala"}'

Here also the sleep was interrupted.

### DIFF
--- a/repl/src/main/scala/org/apache/livy/repl/AbstractSparkInterpreter.scala
+++ b/repl/src/main/scala/org/apache/livy/repl/AbstractSparkInterpreter.scala
@@ -61,6 +61,8 @@ abstract class AbstractSparkInterpreter extends Interpreter with Logging {
 
   protected def conf: SparkConf
 
+  private var currentThread: Thread = _
+
   protected def postStart(): Unit = {
     entries = new SparkEntries(conf)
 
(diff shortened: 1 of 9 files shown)


repl/src/main/scala/org/apache/livy/repl/AbstractSparkInterpreter.scala
```diff
@@ -349,4 +355,10 @@ abstract class AbstractSparkInterpreter extends Interpreter with Logging {
 
     output
   }
+
+  override def cancel(): Unit = {
+    if (currentThread != null) {
+      currentThread.interrupt()
```
heylch, line 361: interrupt cannot terminate the running task. This can resolve the problem of `Thread.sleep(xxx)`, but it won't help for long running task like `while(true): xxx`.

PR #312, opened 2020-10-03
[LIVY-794] Add audit log for interactive session statements submitted through Livy REST API

sergeymazin:
We should add audit log for interactive session statements submitted through Livy REST API.
JIRA: https://issues.apache.org/jira/browse/LIVY-794

## How was this patch tested?
manual tests

Audit log example that get's printed out:
```
20/10/03 12:10:40.886 DEBUG InteractiveSessionAudit$: Executing statement in interactive session 0, appId: APPID, owner: null, proxyUser: null, queue: null, kind: pyspark, code: 1 + 2
```


bin/livy-server
```diff
@@ -102,6 +102,10 @@ start_livy_server() {
     LIVY_CLASSPATH="$LIVY_CLASSPATH:$YARN_CONF_DIR"
   fi
 
+  if [ -n "$LIVY_LOG_DIR" ]; then
+    LIVY_SERVER_JAVA_OPTS="-DLIVY_LOG_DIR=$LIVY_LOG_DIR $LIVY_SERVER_JAVA_OPTS"
```
andrasbeni, line 106: As I can see, the value of $LIVY_LOG_DIR is used only in the example in log4j template. As such, I don't think it's worth adding it here. Especially, because users can add the log dir definition to LIVY_SERVER_JAVA_OPTS right where they export it.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSessionAudit.scala
```diff
@@ -0,0 +1,35 @@
...
+object InteractiveSessionAudit extends Logging {
+  def audit(
+             id: Int,
+             appId: String,
+             owner: String,
+             proxyUser: String,
+             queue: String,
+             kind: String,
+             code: String): Unit = {
+    info(s"Executing statement in interactive session $id, appId: $appId, " +
```
andrasbeni, line 31: Logging the statements on INFO level is unfortunate because by default these logs will end up in the same log file (unless users modify their log4j config when upgrading). DEBUG level is more appropriate (and is shown in the commit message).
I'd like this even more if there were a LivyConf entry to switch off audit logging.

andrasbeni (commented):
@sergeymazin thanks for the pull request. I have a few comments below. I'm curious what you think about them.

bin/livy-server
sergeymazin: Yep, you are right.

But this is added because if you define $LIVY_LOG_DIR in `livy-env.sh` file it won't be passed to JVM as system var and you won't be able to use it in log4j out-of-the-box, additionally, you have to set the same variable second time in $LIVY_SERVER_JAVA_OPTS also. I think this is duplication and end user doesn't have to deal with it.

It took me also some time to figure out why it doesn't work, so decided to make it better for others also.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSessionAudit.scala
sergeymazin: So what do you think, if I add new LivyConf entry:
* should level still change to DEBUG or it can stay INFO if, by default, audit log will be switched off
* any good proposed name for LivyConf entry name?

sergeymazin: Added `livy.server.audit-log.enabled` option and set default to `false`

sergeymazin:
@andrasbeni could you please review this change again when you have time

blv-madhavacarrillo:
I'm interested on this feature too. Thanks for the effort @sergeymazin . Any plans on merging this @andrasbeni ?

andrasbeni:
@blv-madhavacarrillo, I'm not a maintainer of this project. As I can see @RajshekharMuchandi , @askhatri and @dacort are able to merge PRs (and probably others too).

idzikovsky:
I think it'd also make sense to add Audit logging on creation of Batch Sessions.
Because for now this PR covers only one of the two available ways of submitting code to Livy.
Anyway, this feature is nice to have, and I hope to see it merged.

PR #320, opened 2021-04-22
Expose livy session id to allow the session to be shared

jimenefe:
See RD-4563

This allows us to create a Livy client instance, and fetch the Session ID from it. From that point on, we can reuse the session by attaching the session ID to the URL when creating another client instance


jimenefe:
@Ilya Cassina

PR #322, opened 2021-06-21
[LIVY-863] Missing JVM class imports for Spark3

moritzmeister:

Description of the problem: https://issues.apache.org/jira/browse/LIVY-863

The proposed fix consists of adding the missing imports that upstream Spark has when it's initiating the Java Gateway.
See the imports of Spark here: https://github.com/apache/spark/blob/87bf6b0ea4ca0618c8604895d05037edce8b7cb0/python/pyspark/java_gateway.py#L153

As far as I am aware, the `java_import()` does not fail or fails silently if the imported class does not exist.
But I might need to add some code to account for different Spark versions, looking for some guidance on this.

## How was this patch tested?

If you could please provide some guidance where to add tests for this, I am happy to add them.

### DIFF
--- a/repl/src/main/resources/fake_shell.py
+++ b/repl/src/main/resources/fake_shell.py
@@ -581,8 +581,11 @@ def main():
             java_import(gateway.jvm, "org.apache.spark.SparkConf")
             java_import(gateway.jvm, "org.apache.spark.api.java.*")
             java_import(gateway.jvm, "org.apache.spark.api.python.*")
+            java_import(gateway.jvm, "org.apache.spark.ml.python.*")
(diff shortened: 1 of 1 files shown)


yantzu:
Retrigger the test.

lmccay:
@moritzmeister - did you try adding a bogus import to see whether your assumption about it failing silently is true?

moritzmeister:
No I have never done that, I kind of moved on to new work. I see someone in the original Jira issue tested it though: https://issues.apache.org/jira/browse/LIVY-863

nileshrathi345:
looks good to me...
Even tested for few fake imports which has been suggested in https://issues.apache.org/jira/browse/LIVY-863
All test pass

PR #328, opened 2021-09-15
[LIVY-757] Add support for Active Directory through LDAP

idzikovsky:

In this PR I implemented and ability to configure custom LDAP user DN pattern as and alternative to base-dn option, for cases when username attribute isn't `uid` (which is now hardcoded in Livy).

So now instead of specifying baseDN you also can specify user DN pattern by setting the following property in `livy.conf`:
```
livy.server.auth.ldap.user-dn-pattern = sAMAccountName=%s,ou=People,dc=example,dc=com
```

I decided to provide an ability to specify pattern rather than just add an option to configure user attribute, because it seems to be more common way to do this:
* https://github.com/cloudera/hue/blob/18cd309/desktop/conf.dist/hue.ini#L485
* https://github.com/apache/zeppelin/blob/5dd5a65/conf/shiro.ini.template#L46
* https://cwiki.apache.org/confluence/display/Hive/User+and+Group+Filter+Support+with+LDAP+Atn+Provider+in+HiveServer2#UserandGroupFilterSupportwithLDAPAtnProviderinHiveServer2-hive.server2.authentication.ldap.userDNPattern

On the other hand, as Livy already has an option to set baseDN, it would be more logical just to add an option to set custom user attribute.
So I opened another PR with an alternative fix for this issue: https://github.com/apache/incubator-livy/pull/330

## How was this patch tested?

Manual, since unit tests seems to be broken.

### DIFF
--- a/server/src/main/scala/org/apache/livy/LivyConf.scala
+++ b/server/src/main/scala/org/apache/livy/LivyConf.scala
@@ -90,6 +90,7 @@ object LivyConf {
   // Ldap configurations
   val AUTH_LDAP_URL = Entry("livy.server.auth.ldap.url", null)
   val AUTH_LDAP_BASE_DN = Entry("livy.server.auth.ldap.base-dn", null)
+  val AUTH_LDAP_USER_DN_PATTERN = Entry("livy.server.auth.ldap.user-dn-pattern", null)
   val AUTH_LDAP_USERNAME_DOMAIN = Entry("livy.server.auth.ldap.username-domain", null)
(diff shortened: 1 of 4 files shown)


idzikovsky:
I've realized that the current LdapAuthenticationHandler in Livy is exact copy of LdapAuthenticationHandler from Hadoop which also does not support different ways to use LDAP.
So my fix would help only with my case but it's not enough to properly implement different LDAP configurations so it's not worth merging.

PR #330, opened 2021-09-16
[LIVY-757] Add support of Active Directory through LDAP by providing an option to set user attribute

idzikovsky:

In this PR I added an option to set custom username attribute instead of `uid`.
It can be configured in the following way in the `livy.conf`:
```
livy.server.auth.ldap.user-attribute = sAMAccountName
```

This is an alternative implementation of https://github.com/apache/incubator-livy/pull/328

## How was this patch tested?

Manual, since unit tests seems to be broken.


idzikovsky:
I've realized that the current LdapAuthenticationHandler in Livy is exact copy of LdapAuthenticationHandler from Hadoop which also does not support different ways to use LDAP.
So my fix would help only with my case but it's not enough to properly implement different LDAP configurations so it's not worth merging.

PR #332, opened 2021-10-22
[LIVY-871] Add client config to keep fields of table magic in query order

dohongdayi:

Currently the fields in table magic are sorted by field name by default, but in many scenarios, we want to keep fields of table magic in query order, for example:

`val result = spark.sql("select b, a from table")`
`%table result`

currently Livy responses json with fields in the order of "a", "b", but we want them in the order of "b", "a" which is consistent with the SQL

JIRA: https://issues.apache.org/jira/browse/LIVY-871

fix #871

## How was this patch tested?

existing tests & new tests in ScalaInterpreterSpec.


yantzu:
Retrigger the test.

PR #333, opened 2021-11-08
[LIVY-873] Fix pagination issue on Livy UI

tprelle:
The PR add pagination on call to retrieve batches or sessions inside
javascript because when livy reach 100 sessions or batches ui do not
show all sessions or batches

FIX : https://issues.apache.org/jira/browse/LIVY-873

## How was this patch tested?

Manual test

### DIFF
--- a/server/src/main/resources/org/apache/livy/server/ui/static/js/all-sessions.js
+++ b/server/src/main/resources/org/apache/livy/server/ui/static/js/all-sessions.js
@@ -52,35 +52,66 @@ function loadBatchesTable(sessions) {
   });
 }
 
+function getWithPagination(path, from, size){
+    return new Promise((resolve, reject) =>{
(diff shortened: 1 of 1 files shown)


lukasmeirelles:
I tested this changes in my environment and it looks good after the comments I added on the 2 LOC above

yantzu:
@tprelle could you attach the screen capture of UI ,both before and after your change

jianzhenwu:
@tprelle When the page is refreshed, it needs to load all the sessions, doesn't it? Can it do paging search and only load the session of the current page.

jshmchenxi:
> @tprelle When the page is refreshed, it needs to load all the sessions, doesn't it? Can it do paging search and only load the session of the current page？

I don't think it can as the page number would be lost after the refresh. Also the start/end sessions change over time and the content of _current page_ could change when you refresh.

PR #337, opened 2022-03-16
compatiable with spark3.2.1 and scala 2.12.15

dongbixibiX:
fix scala version match from spark-submit --version

compatiable with spark-3.2 and scala-2.12.15
fix scala version match from spark-submit --version

## How was this patch tested?
UT


dongbixibiX:
test

PR #341, opened 2022-05-03
[LIVY-888] AddJar or AddFile call on a duplicate file should not result in failure for Livy session

praneetsharma:

When a duplicate file is attempted to be added using LivyClient, duplicate file's addition is skipped and an appropriate warning is logged instead of the old behavior of throwing fatal IOException. This is now consistent with SparkContext's addJar and addFile APIs behavior. 

https://issues.apache.org/jira/browse/LIVY-888

## How was this patch tested?

Modified the unit test TestSparkClient#testAddJarsAndFiles to attempt adding a duplicate file, and the test did not fail.


jianzhenwu:
Hi @praneetsharma , I think statement failure is acceptable and easy to be perceived by the user, not just a warning.

lmccay:
>

lmccay (approved):
LGTM and rechecks have passed.

+1

PR #344, opened 2022-06-30
[LIVY-889] Adding jars to Spark Interpreter classpath on an already running Livy Session

praneetsharma:

* addJar invoked for a running Livy session will add that jar resource to SparkInterpreter classpath in addition to Spark driver classpath.
* addJarOrPyFile signature changed to return the local path to which it was copied. This local path is required to add the same jar resource to SparkInterpreter classpath.
* addFile now goes through appropriate Interpreter based on the interpreter instead of default PythonInterpreter.

https://issues.apache.org/jira/browse/LIVY-889

## How was this patch tested?

This patch was tested by adding a Jar on a running Livy session and then trying to load a class from that jar using spark interpreter.

### DIFF
--- a/repl/scala-2.11/src/main/scala/org/apache/livy/repl/SparkInterpreter.scala
+++ b/repl/scala-2.11/src/main/scala/org/apache/livy/repl/SparkInterpreter.scala
@@ -18,7 +18,7 @@
 package org.apache.livy.repl
 
 import java.io.File
-import java.net.URLClassLoader
+import java.net.{URL, URLClassLoader}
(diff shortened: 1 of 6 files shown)


jianzhenwu:
Hi @praneetsharma , can you provide the facts on which this PR is based? If it's a problem, can you provide steps to reproduce it?

lmccay:
@jianzhenwu - I believe the description in LIVY-889 is a decent summary of the facts, no?

lmccay (approved):
LGTM and rechecks passed.
+1

PR #348, opened 2022-07-14
[LIVY-891] Expose the sessionID via the client APIs [WIP]

haripriya-b:
https://issues.apache.org/jira/browse/LIVY-891

Currently getSessionID() is exposed in HttpClient only for testing purposes. Since the internally created session ID is useful to doing operations such as re-connecting to the session, killing the session etc., it will be useful to expose it as a public api in HttpClient and LivyClient. (Include a link to the associated JIRA and make sure to add a link to this pr on the JIRA as well)

## How was this patch tested?

### DIFF
--- a/client-http/src/main/java/org/apache/livy/client/http/HttpClient.java
+++ b/client-http/src/main/java/org/apache/livy/client/http/HttpClient.java
@@ -17,23 +17,20 @@
 
 package org.apache.livy.client.http;
 
+import org.apache.livy.Job;
+import org.apache.livy.JobHandle;
+import org.apache.livy.LivyClient;
+import org.apache.livy.client.common.Serializer;
+
 import java.io.File;
 import java.net.URI;
 import java.nio.ByteBuffer;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.concurrent.Callable;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.*;
 import java.util.regex.Matcher;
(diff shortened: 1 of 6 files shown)


client-http/src/main/java/org/apache/livy/client/http/HttpClient.java
```diff
@@ -17,23 +17,20 @@
...
 import java.net.URI;
 import java.nio.ByteBuffer;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.concurrent.Callable;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.*;
```
lmccay, line 30: Please do not use wildcard imports to pull in everything in a package.

lmccay (changes requested):
One minor change requested.

client-http/src/main/java/org/apache/livy/client/http/HttpClient.java
haripriya-b: Thank you, @lmccay. I've updated the files. Can you please review the changes.

lmccay (commented):
Left a comment that questions the appropriateness of this as an extension to LivyClient given that RSCClient is throwing an UnsupportedOperationException. Given that it represents half of the implementations of LivyClient, it seems like consumers of the interface shouldn't encounter a runtime exception from a couple new getters. It also strikes me as odd to need to add a new dependency on livy-api to rsc/pom.xml in order to throw and UnsupportedOperationException. There is code across the project using expecting to us the LivyClient interface - maybe I am missing something there.

PR #367, opened 2022-12-07
Add a Dockerfile for the build

dacort:

In [LIVY-784](https://issues.apache.org/jira/browse/LIVY-784) it was mentioned that the provided `mvn package` resulted in a maven error. This PR adds a Dockerfile that can be used to successfully run maven against the Livy Project.

## How was this patch tested?

By running a combination of `mvn test`, `mvn package` and `mvn -Pthriftserver verify`.

## TODO

- [x] The `print unicode correctly` test in `Python2InterpreterSpec` is still failing due to unicode handling differences in Python2/Python3.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,0 @@
-# Build output
-**/target/
-
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+# Build output
+**/target/
+
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,13 @@
 FROM ubuntu:xenial
 
+# configure locale
+RUN apt update -qq > /dev/null && apt install -qq --yes --no-install-recommends \
+    locales && \
+    locale-gen en_US.UTF-8
+ENV LANG="en_US.UTF-8" \
+    LANGUAGE="en_US.UTF-8" \
+    LC_ALL="en_US.UTF-8"
+
 # Install necessary dependencies for build/test
 RUN apt update && \
     apt install -y \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,34 @@
+FROM ubuntu:xenial
+
+# Install necessary dependencies for build/test
+RUN apt update && \
+    apt install -y \
+    apt-transport-https \
+    libkrb5-dev \
+    maven \
+    openjdk-8-jdk \
+    python-dev \
+    python-pip \
+    python3-pip \
+    software-properties-common
+
+# R 3.x install - ensure to add the signing key per https://cran.r-project.org/bin/linux/ubuntu/olderreleasesREADME.html
+RUN add-apt-repository 'deb https://cloud.r-project.org/bin/linux/ubuntu xenial-cran35/' && \
+    apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E298A3A825C0D65DFD57CBB651716619E084DAB9 && \
+    apt-get update && \
+    apt-get -y install r-base
+
+# Add build dependencies for python2
+# - First we upgrade pip because that makes a lot of things better
+# - Then we remove the provided version of setuptools and install a different version
+# - Then we install additional dependencies
+RUN python -m pip install -U "pip < 21.0"
+RUN apt-get remove -y python-setuptools
+RUN python -m pip install "setuptools < 36"
+RUN python -m pip install "requests >= 2.10.0" "responses >= 0.5.1" "futures>=3.0.5" "future>=0.15.2" pytest pytest-runner flaky flake8 requests-kerberos install codecov cloudpickle
+
+# Now do the same for python3
+RUN python3 -m pip install -U pip
+
+WORKDIR /workspace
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+# Add build dependencies for python2
+# - First we upgrade pip because that makes a lot of things better
+# - Then we remove the provided version of setuptools and install a different version
+# - Then we install additional dependencies
+RUN python -m pip install -U "pip < 21.0"
+RUN apt-get remove -y python-setuptools
+RUN python -m pip install "setuptools < 36"
+RUN python -m pip install "requests >= 2.10.0" "responses >= 0.5.1" "futures>=3.0.5" "future>=0.15.2" pytest pytest-runner flaky flake8 requests-kerberos install codecov cloudpickle
+
```
ksumit, line 29: should we consider merging these into a single layer, i.e. just like what is done in line#16-19

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+# Now do the same for python3
+RUN python3 -m pip install -U pip
+
+WORKDIR /workspace
```
ksumit, line 33: It might be useful to map this workspace to the real workspace on the host?

Dockerfile
dacort: I can double-check with setuptools specifically, but I know with the pip upgrade, it needs to run as two separate commands for the pip upgrade ot take effect. I _think_ `setuptools` is the same, but I should actually be able to at least squash lines#26-27

Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM ubuntu:xenial
+
+# Install necessary dependencies for build/test
+RUN apt update && \
```
ksumit, line 4: should we add other tools that are useful for dev/debugging, few thinks that come to mind - `git vim tmux curl wget lsb-release telnet netcat iputils-ping sudo`

Dockerfile
dacort: As in a `COPY . /workspace`? Or do you mean something else?

Dockerfile
dacort: git and vim, definitely. The others could definitely be useful too.

re: `sudo` Right now the image is running - since it's largely used for local dev/test not sure if we should also create a `dev` user or similar.

Dockerfile
dacort: Got it, yep. In the readme, I mention the mapping for the workspace, but not for `.m2`. 

I don't think anything needs to be changed in the Dockerfile to support that, though?

Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM ubuntu:xenial
```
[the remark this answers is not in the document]
dacort: *nod* For sure. Today, Travis is used (see `.travis.yml`), but would nice to be able to use the same container image.

Dockerfile
dacort: Cool. Just pushed up a new readme that adds the `.m2` part. Leaving as `root` for now.

Dockerfile
dacort: Added `git vim curl wget`

.dockerignore
```diff
@@ -0,0 +1,3 @@
+# Build output
```
ksumit, line 1: I tried doing a build and noticed that it was sending all the artifacts from `.` to the docker daemon. Should we consider creating a separate folder so we don't do this? Here is a snapshot i noticed:
```
$ docker build -t livy .
^Cnding build context to Docker daemon  4.556GB
```

Pls ignore the size (4.556GB) you see there, that was because i put Dockerfile in a folder with lots of content.

dacort: I believe the `target` directories made up the majority of the artifacts getting sent to the docker daemon. 

When using that, the build context is much smaller.

```
Sending build context to Docker daemon  7.879MB
```

I'm unsure what typical practices are here. [Many other apache projects](https://cs.github.com/?scopeName=All+repos&scope=&q=org%3Aapache+language%3ADockerfile) seem to place them in subdirs.

ksumit: I was wondering if we move Dockerfile to a separate folder, we won't have to keep this `.dockerignore` file at all. This size will grow over time, so that's a suggestion. I agree that 8MB may not be significant.

[Here](https://github.com/apache/spark/blob/master/dev/infra/Dockerfile) is another example from spark project to consider

dacort: Seems like it's fairly common to place Dockerfiles in something like `dev/infra` or `support/` directories. Seems to make sense.

dacort: Moved Dockerfile into `dev/docker` and updated README appropriately.

lmccay:
@dacort - is the Spark 2.4 unit test failures related to the commented our spark line at the end of the docker file?
Is it unrelated to this change and just a flaky test?
Maybe retry the build?

dacort:
Looks like a flake, re-running.

lmccay (approved):
+1

Thanks, @dacort !!!

PR #373, opened 2022-12-16
compatiable with spark3.2 and scala 2.12.15

dongbixibiX:
fix scala version match from spark-submit --version

compatiable with spark-3.2 and scala-2.12.15
fix scala version match from spark-submit --version

## How was this patch tested?

by UT

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,20 +16,20 @@
 #
 
 sudo: required
-dist: xenial
+dist: bonic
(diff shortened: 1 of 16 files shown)


lmccay (changes requested):
Hi @dongbixibiX - I notice a bunch of exclusions being removed here.
Can you describe these changes and why it is okay to remove them?

PR #377, opened 2022-12-18
[Livy 899] The state of interactive session is always idle when using thrift protocol.

jianzhenwu:
1. Should broadcast `ReplState` to the `RSCClient` in `RSCDriver` or the session state is always idle when using thrift protocol.
2. Add unit test `testSessionState` in `ThriftSessionTest` to verify that the state change is as expected.

https://issues.apache.org/jira/browse/LIVY-899

## How was this patch tested?
This patch is tested by unit tests.  The the state change is recorded in `replStateChangedHistoryInTest ` when testing.


thriftserver/session/src/test/java/org/apache/livy/thriftserver/session/ThriftSessionTest.java
```diff
@@ -192,6 +192,27 @@ public void testThriftSession() throws Exception {
     waitFor(new UnregisterSessionJob(s3));
   }
 
+  @Test
+  public void testSessionState() throws Exception {
```
mgaido91, line 196: why are we testing the RSCDriver in thriftserver? All changes are outside the thriftserver, so the UTs should be in rsc as well. This will also avoid adding dependencies to the thriftserver module.

mgaido91 (commented):
I am not sure about the change, as this does not involve only the thriftserver module, but all interactive sessions. If the problem involves all interactive sessions, then it is ok, otherwise the fix is probably not the correct one. Can you check if the same issue affects all interactive jobs? If so, I think we can go ahead with this PR, but I would recommend a different reviewer more familiar with the code in rsc, otherwise, the fix needs to be done only on the thriftserver part and we need more investigation to understand why normal interactive sessions do not have the issue while thriftserver has. Thanks.

rsc/src/main/java/org/apache/livy/rsc/driver/RSCDriver.java
```diff
@@ -392,6 +393,13 @@ <T> void jobFinished(String jobId, T result, Throwable error) {
 
   void jobStarted(String jobId) {
     broadcast(new JobStarted(jobId));
+    broadcast(new ReplState(SessionState.Busy$.MODULE$.state()));
+  }
+
+  public void idleIfActiveJobsEmpty(){
```
leesf, line 399: `public void idleIfActiveJobsEmpty(){` -> `public void idleIfActiveJobsEmpty() {`

thriftserver/session/src/test/java/org/apache/livy/thriftserver/session/ThriftSessionTest.java
```diff
@@ -19,21 +19,21 @@
 
 import java.net.URI;
 import java.nio.file.Files;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Properties;
-import java.util.concurrent.TimeUnit;
+import java.util.*;
```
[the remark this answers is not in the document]
leesf: +1

thriftserver/session/src/test/java/org/apache/livy/thriftserver/session/ThriftSessionTest.java
```diff
@@ -19,21 +19,21 @@
...
 import java.util.concurrent.TimeoutException;
+import static java.util.concurrent.TimeUnit.*;
```
leesf, line 24: ditto

jianzhenwu:
> I am not sure about the change, as this does not involve only the thriftserver module, but all interactive sessions. If the problem involves all interactive sessions, then it is ok, otherwise the fix is probably not the correct one. Can you check if the same issue affects all interactive jobs? If so, I think we can go ahead with this PR, but I would recommend a different reviewer more familiar with the code in rsc, otherwise, the fix needs to be done only on the thriftserver part and we need more investigation to understand why normal interactive sessions do not have the issue while thriftserver has. Thanks.

The same issue doesn't affect interactive jobs which created by REST API. The is because the handle functions are different in the driver. The REST API uses `handle(ctx: ChannelHandlerContext, msg: BaseProtocol.ReplJobRequest)` in the ReplDriver while thrift API uses `handle(ChannelHandlerContext ctx, JobRequest<?> msg)` in the RSCDRiver.  
ReplDriver changes state when execute code by invoking stateChangedCallback which broadcast the state to all RPC clients. 
RSCDriver does not broadcast  state to RPC clients. 
I think this patch would not affect other interactive jobs except using thrift API.

thriftserver/session/src/test/java/org/apache/livy/thriftserver/session/ThriftSessionTest.java
jianzhenwu: Because I need to verify that the state is correct at the end.

mgaido91: at the end of what? Here we are not going through the thriftserver, so this test should be moved to rsc

mgaido91:
This fix is causing many changes. The `submit` method is used for many purposes, such as adding jar, files, so I do not think is a good idea to have a fix that has such a large impact on other cases that should not be affected. The current fix would also set to busy when fetching the results, and executing schema operations, and even while unregistering a session...

I think a better place for the fix would be `SqlJob`, although it is not a perfect solution either. So I would be very happy if someone comes with a better alternative.

jianzhenwu:
> 



> This fix is causing many changes. The `submit` method is used for many purposes, such as adding jar, files, so I do not think is a good idea to have a fix that has such a large impact on other cases that should not be affected. The current fix would also set to busy when fetching the results, and executing schema operations, and even while unregistering a session...
> 
> I think a better place for the fix would be `SqlJob`, although it is not a perfect solution either. So I would be very happy if someone comes with a better alternative.

sorry for the late reply. I have migrated the changes to SqlJob, can you @mgaido91  review the code again?

PR #384, opened 2023-01-23
[LIVY-968][SERVER] Provide ttl field for a livy session

askhatri:

A session should have a field 'ttl' (Time to live) which once exceeded should get the session killed.This field can be a parameter in the POST /sessions request and then be a part of session object. Livy has a config `livy.server.session.timeout` but this is a global config and cannot be configured for a session. A 'ttl' field can help set this for a session level.

JIRA: https://issues.apache.org/jira/browse/LIVY-968

## How was this patch tested?

Verified manually by creating interactive session via REST API call in a local Yarn cluster. Also, we have add the new unit tests.

### DIFF
--- a/client-common/src/main/java/org/apache/livy/client/common/ClientConf.java
+++ b/client-common/src/main/java/org/apache/livy/client/common/ClientConf.java
@@ -36,7 +36,7 @@ import org.apache.livy.annotations.Private;
 public abstract class ClientConf<T extends ClientConf>
   implements Iterable<Map.Entry<String, String>> {
 
-  protected Logger LOG = LoggerFactory.getLogger(getClass());
+  protected static Logger LOG = LoggerFactory.getLogger(ClientConf.class);
 
   public static interface ConfEntry {
 
@@ -263,6 +263,29 @@ public abstract class ClientConf<T extends ClientConf>
       }
     }
     return altToNewKeyMap;
+  }
+
+  public static long getTimeAsNanos(String time, int sessionId, long defaultVale) {
+    if (time == null) {
+      return defaultVale;
+    }
+
+    Matcher m = Pattern.compile("(-?[0-9]+)([a-z]+)?").matcher(time.toLowerCase());
+    if (!m.matches()) {
+      LOG.error("Invalid time string: " + time + " for session ID: " + sessionId);
+      return defaultVale;
+    }
+
+    long val = Long.parseLong(m.group(1));
+    String suffix = m.group(2);
+
+    if (suffix != null && !TIME_SUFFIXES.containsKey(suffix)) {
+      LOG.error("Invalid suffix: " + time + " for session ID: " + sessionId);
+      return defaultVale;
+    }
+
+    return TimeUnit.NANOSECONDS.convert(val,
+            suffix != null ? TIME_SUFFIXES.get(suffix) : TimeUnit.MILLISECONDS);
   }
 
   public static interface DeprecatedConf {
--- a/client-common/src/main/java/org/apache/livy/client/common/ClientConf.java
+++ b/client-common/src/main/java/org/apache/livy/client/common/ClientConf.java
@@ -36,7 +36,7 @@ import org.apache.livy.annotations.Private;
 public abstract class ClientConf<T extends ClientConf>
   implements Iterable<Map.Entry<String, String>> {
 
-  protected static Logger LOG = LoggerFactory.getLogger(ClientConf.class);
+  protected static final Logger LOG = LoggerFactory.getLogger(ClientConf.class);
 
   public static interface ConfEntry {
 
@@ -265,12 +265,34 @@ public abstract class ClientConf<T extends ClientConf>
     return altToNewKeyMap;
   }
 
+  public static boolean validateTtl(String ttl) {
+    if (ttl == null || ttl.trim().isEmpty()) {
+      return true;
+    }
+    Matcher m = Pattern.compile("(-?[0-9]+)([a-z]+)?").matcher(ttl.trim().toLowerCase());
+    if (!m.matches()) {
+      LOG.error("Invalid ttl string: " + ttl);
+      return false;
+    }
+    long val = Long.parseLong(m.group(1));
+    if (val <= 0L) {
+      LOG.error("Invalid ttl value: " + val);
+      return false;
+    }
+    String suffix = m.group(2);
+    if (suffix != null && !TIME_SUFFIXES.containsKey(suffix)) {
+      LOG.error("Invalid ttl suffix: " + suffix);
+      return false;
+    }
+    return true;
+  }
+
   public static long getTimeAsNanos(String time, int sessionId, long defaultVale) {
-    if (time == null) {
+    if (time == null || time.trim().isEmpty()) {
       return defaultVale;
     }
 
-    Matcher m = Pattern.compile("(-?[0-9]+)([a-z]+)?").matcher(time.toLowerCase());
+    Matcher m = Pattern.compile("(-?[0-9]+)([a-z]+)?").matcher(time.trim().toLowerCase());
     if (!m.matches()) {
       LOG.error("Invalid time string: " + time + " for session ID: " + sessionId);
       return defaultVale;
@@ -285,7 +307,7 @@ public abstract class ClientConf<T extends ClientConf>
     }
 
     return TimeUnit.NANOSECONDS.convert(val,
-            suffix != null ? TIME_SUFFIXES.get(suffix) : TimeUnit.MILLISECONDS);
+            suffix != null ? TIME_SUFFIXES.get(suffix) : TimeUnit.SECONDS);
   }
 
   public static interface DeprecatedConf {
--- a/client-common/src/main/java/org/apache/livy/client/common/ClientConf.java
+++ b/client-common/src/main/java/org/apache/livy/client/common/ClientConf.java
@@ -152,11 +152,15 @@ public abstract class ClientConf<T extends ClientConf>
     String time = get(e, String.class);
     if (time == null) {
       check(e.dflt() != null,
-        "ConfEntry %s doesn't have a default value, cannot convert to time value.", e.key());
+              "ConfEntry %s doesn't have a default value, cannot convert to time value.", e.key());
       time = (String) e.dflt();
     }
+    return getTimeAsMs(time);
+  }
 
-    Matcher m = Pattern.compile("(-?[0-9]+)([a-z]+)?").matcher(time.toLowerCase());
+  public static long getTimeAsMs(String time) {
+    check(time != null && !time.trim().isEmpty(), "Invalid time string: %s", time);
+    Matcher m = Pattern.compile("(-?[0-9]+)([a-z]+)?").matcher(time.trim().toLowerCase());
     if (!m.matches()) {
       throw new IllegalArgumentException("Invalid time string: " + time);
     }
@@ -168,8 +172,12 @@ public abstract class ClientConf<T extends ClientConf>
       throw new IllegalArgumentException("Invalid suffix: \"" + suffix + "\"");
     }
 
+    if (val < 0L) {
+      throw new IllegalArgumentException("Invalid value: " + val);
+    }
+
     return TimeUnit.MILLISECONDS.convert(val,
-      suffix != null ? TIME_SUFFIXES.get(suffix) : TimeUnit.MILLISECONDS);
+            suffix != null ? TIME_SUFFIXES.get(suffix) : TimeUnit.MILLISECONDS);
   }
 
   @SuppressWarnings("unchecked")
@@ -204,7 +212,7 @@ public abstract class ClientConf<T extends ClientConf>
     return (o != null) ? o.getClass() : String.class;
   }
 
-  private void check(boolean test, String message, Object... args) {
+  private static void check(boolean test, String message, Object... args) {
     if (!test) {
       throw new IllegalArgumentException(String.format(message, args));
     }
@@ -263,34 +271,6 @@ public abstract class ClientConf<T extends ClientConf>
       }
     }
     return altToNewKeyMap;
-  }
-
-  public static long getTimeAsNanos(String time, int sessionId, long defaultVale) {
-    if (time == null || time.trim().isEmpty()) {
-      return defaultVale;
-    }
-
-    Matcher m = Pattern.compile("(-?[0-9]+)([a-z]+)?").matcher(time.trim().toLowerCase());
-    if (!m.matches()) {
-      throw new IllegalArgumentException("Invalid ttl string: " + time +
-              " for session: " + sessionId);
-    }
-
-    long val = Long.parseLong(m.group(1));
-    if (val <= 0L) {
-      throw new IllegalArgumentException("Invalid ttl value: " + val +
-              " for session: " + sessionId);
-    }
-
-    String suffix = m.group(2);
-
-    if (suffix != null && !TIME_SUFFIXES.containsKey(suffix)) {
-      throw new IllegalArgumentException("Invalid ttl suffix: " + time +
-              " for session: " + sessionId);
-    }
-
-    return TimeUnit.NANOSECONDS.convert(val,
-            suffix != null ? TIME_SUFFIXES.get(suffix) : TimeUnit.SECONDS);
   }
 
   public static interface DeprecatedConf {
--- a/client-common/src/main/java/org/apache/livy/client/common/ClientConf.java
+++ b/client-common/src/main/java/org/apache/livy/client/common/ClientConf.java
@@ -265,28 +265,6 @@ public abstract class ClientConf<T extends ClientConf>
     return altToNewKeyMap;
   }
 
-  public static boolean validateTtl(String ttl) {
-    if (ttl == null || ttl.trim().isEmpty()) {
-      return true;
-    }
-    Matcher m = Pattern.compile("(-?[0-9]+)([a-z]+)?").matcher(ttl.trim().toLowerCase());
-    if (!m.matches()) {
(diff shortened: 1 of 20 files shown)


server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
```diff
@@ -168,7 +169,9 @@ class SessionManager[S <: Session, R <: RecoveryMetadata : ClassTag](
             false
           } else {
             val currentTime = System.nanoTime()
-            currentTime - session.lastActivity > sessionTimeout
+            val calculatedTimeout =
+              ClientConf.getTimeAsNanos(session.ttl.orNull, session.id, sessionTimeout)
```
gyogal, line 173: The TTL string is currently only validated here. If the formatting is invalid it seems like this will keep logging error messages every time `expired()` is called. To prevent this, the string could be validated as the session in created.

client-common/src/main/java/org/apache/livy/client/common/ClientConf.java
```diff
@@ -36,7 +36,7 @@
 public abstract class ClientConf<T extends ClientConf>
   implements Iterable<Map.Entry<String, String>> {
 
-  protected Logger LOG = LoggerFactory.getLogger(getClass());
+  protected static Logger LOG = LoggerFactory.getLogger(ClientConf.class);
```
ksumit, line 39: should this be final too?

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
```diff
@@ -168,7 +169,9 @@ class SessionManager[S <: Session, R <: RecoveryMetadata : ClassTag](
             false
           } else {
             val currentTime = System.nanoTime()
-            currentTime - session.lastActivity > sessionTimeout
+            val calculatedTimeout =
```
ksumit, line 172: nanoseconds seems too much to me, from my exposure to customer usecases i think ttl in minutes should be enough. curious to know when nanoseconds level of granularity would be useful.

please consider simplifying this further by making this an `Int` or `Long` and avoiding all the parsing and validation logic. for the variable naming, what do you think of `ttlInMinutes` or `ttlInSeconds`?

docs/rest-api.md
```diff
@@ -151,6 +151,11 @@ Creates a new interactive Scala, Python, or R shell in the cluster.
     <td>Timeout in second to which session be orphaned</td>
     <td>int</td>
   </tr>
+  <tr>
+    <td>ttl</td>
+    <td>The timeout for this inactive session</td>
```
ksumit, line 156: would be useful to provide some examples here.

`timeout for this inactive session in minutes/seconds, example: 10`

server/src/main/scala/org/apache/livy/sessions/Session.scala
```diff
@@ -138,6 +138,7 @@ abstract class Session(
     val id: Int,
     val name: Option[String],
     val owner: String,
+    val ttl: Option[String],
```
ksumit, line 141: should we add an auxiliary constructor as well that matches the existing constructor (4 parameters only) and takes the global session timeout as ttl by default? this will help reduce the amount of change in this PR.

ksumit (changes requested):
Overall looks good, not clear on when a nanosecond level of ttl will be useful for a session, my recommendation would to be to keep it at minutes but i'm ok with seconds too.

client-common/src/main/java/org/apache/livy/client/common/ClientConf.java
askhatri: Sure @ksumit , I will try to make it final

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
askhatri: nanoseconds logic is only for internal calculation and `System.nanoTime()` is an existing code. End user has option to provide the value like "30s" for 30 seconds and "2m" for 2 minutes.

docs/rest-api.md
askhatri: Sure @ksumit , I will try to update this note.

server/src/main/scala/org/apache/livy/sessions/Session.scala
askhatri: Sure @ksumit , I will check the possibility to add a auxiliary constructor.

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
ksumit: hi @askhatri, i should have been more specific, i had following specific concerns:
1. the logic on parsing the string and converting it to seconds, milliseconds etc seems more complex. instead i was proposing that we use `Int` or `Long` and use number to denote the ttl value
2. fix the granularity for ttl either in minutes or seconds to avoid the complexity
3. depending on whether we fix it in minutes or seconds to be very explicit, name the variable as `ttlInMinutes` or `ttlInSeconds`

askhatri: I got it @ksumit . So presently we have three options now:
1) Use "ttl" with "String" datatype and keep parsing logic same as global config.
2) Change filed name to "ttlInMinutes" with "Int" datatype and support only minutes in user inputs.
3) Change filed name to "ttlInSeconds" with "Int" datatype and support only seconds in user inputs.

Since the logic to parse string already and it is consistent with global configuration we are planing to go with option 1. i.e. using "ttl" field as "String".

ksumit: I see now that global configuration is also using this complicated logic. Thanks for pointing that out. I also noticed that there are bunch of other configurations that follow this same logic. So we can leave this as it is. I'm concerned about duplicate code logic in `getTimeAsMs()` and `getTimeAsNanos()`, please see if we can avoid the same or reuse what's already there?

client-common/src/main/java/org/apache/livy/client/common/ClientConf.java
```diff
@@ -265,6 +265,51 @@ private Map<String, ConfPair> allAlternativeKeys() {
...
+    }
+    String suffix = m.group(2);
+    if (suffix != null && !TIME_SUFFIXES.containsKey(suffix)) {
+      LOG.error("Invalid ttl suffix: " + suffix);
+      return false;
+    }
+    return true;
+  }
+
+  public static long getTimeAsNanos(String time, int sessionId, long defaultVale) {
```
ksumit, line 290: What is your opinion in reusing `getTimeAsMs()` instead of creating a new one? Other than the granularity, do you see any other gap between `getTimeAsMs()` and `getTimeAsNanos()` implementations?

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSessionServlet.scala
```diff
@@ -52,15 +53,20 @@ class InteractiveSessionServlet(
...
+        createRequest.name,
+        remoteUser(req),
+        proxyUser(req, createRequest.proxyUser),
+        livyConf,
+        accessManager,
+        createRequest,
+        sessionStore,
+        createRequest.ttl)
+    } else {
+      null
```
ksumit, line 68: should this be considered an error case instead and we throw an appropriate error message here? Or could we log the failure and use the global ttl value in this case as the default value?

client-common/src/main/java/org/apache/livy/client/common/ClientConf.java
```diff
@@ -265,6 +265,51 @@ private Map<String, ConfPair> allAlternativeKeys() {
     return altToNewKeyMap;
   }
 
+  public static boolean validateTtl(String ttl) {
```
ksumit, line 268: Looks like `validateTtl()` and `getTimeAsNanos()` are using a lot of duplicate code, could we avoid this?

client-common/src/main/java/org/apache/livy/client/common/HttpMessages.java
```diff
@@ -61,9 +61,10 @@ public static class SessionInfo implements ClientMessage {
     public final String kind;
     public final Map<String, String> appInfo;
     public final List<String> log;
+    public final String ttl;
 
     public SessionInfo(int id, String name, String appId, String owner, String proxyUser,
```
ksumit, line 66: given that ttl is optional field, should we create an override constructor without this parameter? that way you will be able to avoid other changes in test classes.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -131,13 +131,18 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
...
-        val session = sessionManager.register(createSession(request))
-        // Because it may take some time to establish the session, update the last activity
-        // time before returning the session info to the client.
-        session.recordActivity()
-        Created(clientSessionView(session, request),
-          headers = Map("Location" ->
-            (getRequestPathInfo(request) + url(getSession, "id" -> session.id.toString))))
+        val input = createSession(request)
+        if(input == null) {
+          BadRequest(ResponseMessage("Rejected, invalid value for ttl field!"))
```
ksumit, line 136: should we throw an exception in `createSession()` and use appropriate message from that to return the `ResponseMessage` here? There could potentially be other reasons on why `createSession()` could return null, right?

server/src/main/scala/org/apache/livy/server/interactive/CreateInteractiveRequest.scala
```diff
@@ -50,6 +51,7 @@ class CreateInteractiveRequest {
       (if (queue.isDefined) s"queue: ${queue.get}, " else "") +
       (if (name.isDefined) s"name: ${name.get}, " else "") +
       (if (conf.nonEmpty) s"conf: ${conf.mkString(",")}, " else "") +
-      s"heartbeatTimeoutInSecond: $heartbeatTimeoutInSecond]"
+      s"heartbeatTimeoutInSecond: $heartbeatTimeoutInSecond, " +
+      (if (ttl.isDefined) s"driverMemory: ${ttl.get}]" else "]")
```
ksumit, line 55: typo: we need `ttl` instead of `driverMemory`

ksumit (changes requested):

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSessionServlet.scala
askhatri: I have fixed this by adding logic to throw an error.

client-common/src/main/java/org/apache/livy/client/common/ClientConf.java
askhatri: Added the logic to reuse the same method.

client-common/src/main/java/org/apache/livy/client/common/HttpMessages.java
askhatri: This constructor is not used in test classes and it is only one place in InteractiveSessionServlet.scala so we can skip adding extra constructor I believe.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
askhatri: I have updated the code to avoid returning null.

server/src/main/scala/org/apache/livy/server/interactive/CreateInteractiveRequest.scala
askhatri: Done

client-common/src/main/java/org/apache/livy/client/common/ClientConf.java
askhatri: `getTimeAsMs()` is defined at class level as non-static method whereas `getTimeAsNanos()` is a static method. It not easy to reuse I believe.

askhatri:
Hi @ksumit , most of the changes are completed. Could you please check if we are good to merge it?
CC: @gyogal

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
askhatri: I have completed this code changes.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSessionServlet.scala
```diff
@@ -52,15 +53,18 @@ class InteractiveSessionServlet(
 
   override protected def createSession(req: HttpServletRequest): InteractiveSession = {
     val createRequest = bodyAs[CreateInteractiveRequest](req)
+    val sessionId = sessionManager.nextId();
+    ClientConf.getTimeAsNanos(createRequest.ttl.orNull, sessionId, 0L);
```
ksumit, line 57: this could be place for us to parse and set the ttl value correctly:
1. if it's present and valid, parse it and set to the parsed value
2. if it's invalid, set it to the global session timeout value or reject the request.

This method call is doing the same thing but passing 0L as default value seems hacky/confusing to me. Also calling a get method in isolation without using the return value seems hacky/confusing to me.

ksumit (changes requested):
I'm still concerned about duplicate code in `getTimeAsNanos()`. I've suggested one way to reuse that code. Please consider @gyogal's feedback on parsing and using the correct ttl at the session creation time, there are some minor comments on how we are using a get method to do validation as side effect as well, it's confusing.

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
askhatri: Considering the support for session recovery, we have decided to avoid converting it into long value during session creation.

client-common/src/main/java/org/apache/livy/client/common/ClientConf.java
askhatri: I have updated the code to avoid the code duplication.

askhatri:
Hi @ksumit , today I have updated the code for further review. 😊
CC: @gyogal

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
```diff
@@ -131,13 +131,18 @@ abstract class SessionServlet[S <: Session, R <: RecoveryMetadata](
...
+          val session = sessionManager.register(createSession(request))
+          // Because it may take some time to establish the session, update the last activity
+          // time before returning the session info to the client.
+          session.recordActivity()
+          Created(clientSessionView(session, request),
+            headers = Map("Location" ->
+              (getRequestPathInfo(request) + url(getSession, "id" -> session.id.toString))))
+        } catch {
+          case e: IllegalArgumentException =>
+            BadRequest(ResponseMessage("Rejected, Invalid ttl field: " + e.getMessage))
```
ksumit, line 144: Right now, the message may be related to ttl field validation but this doesn't have to be right? Can we do this instead please?

```
BadRequest(ResponseMessage("Rejected, Reason: " + e.getMessage))
```

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSessionServlet.scala
```diff
@@ -52,15 +53,23 @@ class InteractiveSessionServlet(
 
   override protected def createSession(req: HttpServletRequest): InteractiveSession = {
     val createRequest = bodyAs[CreateInteractiveRequest](req)
+    val sessionId = sessionManager.nextId();
+
+    // Calling getTimeAsMs just to validate the ttl value
+    if (createRequest.ttl.isDefined) {
+      ClientConf.getTimeAsMs(createRequest.ttl.orNull);
```
ksumit, line 60: we don't need to do `orNull` here right?

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
```diff
@@ -168,7 +170,12 @@ class SessionManager[S <: Session, R <: RecoveryMetadata : ClassTag](
             false
           } else {
             val currentTime = System.nanoTime()
-            currentTime - session.lastActivity > sessionTimeout
+            var calculatedTimeout = sessionTimeout;
+            if (session.ttl.isDefined) {
+              calculatedTimeout = ClientConf.getTimeAsMs(session.ttl.orNull)
```
ksumit, line 175: we don't need `orNull` here right?

ksumit (changes requested):
Minor comments, looks good otherwise. Please fix them and i will approve from my side. Thank you for your contribution @askhatri

server/src/main/scala/org/apache/livy/sessions/SessionManager.scala
askhatri: I have removed `orNull`.

server/src/main/scala/org/apache/livy/server/interactive/InteractiveSessionServlet.scala
askhatri: I have removed `orNull`.

server/src/main/scala/org/apache/livy/server/SessionServlet.scala
askhatri: I have updated the code as requested.

askhatri:
> Minor comments, looks good otherwise. Please fix them and i will approve from my side. Thank you for your contribution @askhatri

All comments fixed today as suggested by you.

ksumit (approved):
LGTM, +1

gyogal:
This looks good to me as well. Thinking about it, it may be good to introduce a maximum or a limit for this, so that it cannot be longer than a set amount. But this could be addressed in a separate ticket.

gyogal (approved):

PR #249, opened 2019-10-27
[LIVY-702]: Submit Spark apps to Kubernetes

jahstreet:
## What changes were proposed in this pull request?

[Jira](https://issues.apache.org/jira/browse/LIVY-702)

This PR is one of the PRs in the series related to the splitting of the base PR https://github.com/apache/incubator-livy/pull/167 to multiple PRs to ease and speed up review and merge processes.

This PR proposes a way to submit Spark apps to Kubernetes cluster. Points covered:
- Submit batch sessions
- Submit interactive sessions
- Monitor sessions, collect logs and diagnostics information
- Restore sessions monitoring after restarts
- GC created Kubernetes resources
- Restrict the set of allowed Kubernetes namespaces

## How was this patch tested?

Unit tests.

Manual testing with Kubernetes on Docker Desktop for Mac v2.1.0.1.
Environment - Helm charts:
- [cluster-base](https://github.com/jahstreet/spark-on-kubernetes-helm/tree/master/charts/cluster-base) with custom-values.yaml:
```yaml
nginx-ingress:
  controller:
    service:
      loadBalancerIP: 127.0.0.1 # my-cluster.example.com IP address (from /etc/hosts)
      loadBalancerSourceRanges: []
cluster-autoscaler:
  enabled: false
oauth2-proxy:
  enabled: false
```

  - [spark-cluster](https://github.com/jahstreet/spark-on-kubernetes-helm/tree/master/charts/spark-cluster) with custom-values.yaml:
```yaml
livy:
  image:
    pullPolicy: Never
    tag: 0.7.0-incubating-spark_2.4.3_2.11-hadoop_3.2.0-dev
  ingress:
    enabled: true
    annotations:
      kubernetes.io/ingress.class: nginx
      kubernetes.io/tls-acme: "true"
      nginx.ingress.kubernetes.io/rewrite-target: /$1
    path: /livy/?(.*)
    hosts:
    - my-cluster.example.com
    tls:
    - secretName: spark-cluster-tls
      hosts:
      - my-cluster.example.com
  persistence:
    enabled: true
  env:
    LIVY_LIVY_UI_BASE1PATH: {value: "/livy"}
    LIVY_SPARK_KUBERNETES_CONTAINER_IMAGE_PULL1POLICY: {value: "Never"}
[...]

PR #167, opened 2019-04-09
[LIVY-588]: Full support for Spark on Kubernetes

jahstreet:
**NOTE**: this PR is deprecated and kept for discussions history only. Please refer the #249 to get the latest state of the work.

## What changes were proposed in this pull request?

This PR is a new feature proposal: full support for Spark on Kubernetes (inspired by SparkYarnApp implementation).

Since Spark on Kubernetes has been released relatively long ago this can be a good idea to include Kubernetes support to Livy project as well, as it can solve much problems related to working with Spark on Kubernetes, it can fully replace Yarn in case of working atop Kubernetes cluster:
- Livy UI has cached logs/diagnostics page
- Livy UI shows links to Spark UI and Spark History Server
- With Kubernetes Ingress resource Livy can be configured to serve as an orchestrator of Spark Apps atop Kubernetes (PR includes Nginx Ingress support option to create routes to Spark UI)
- Nginx Ingress solves `basePath` support for Spark UI and History Server as well as has lots of auth integrations available: https://github.com/kubernetes/ingress-nginx
- Livy UI can be integrated with Grafana Loki logs (PR provides solution for that)

Dockerfiles repo: https://github.com/jahstreet/spark-on-kubernetes-docker
Helm charts: https://github.com/jahstreet/spark-on-kubernetes-helm

Associated JIRA: https://issues.apache.org/jira/browse/LIVY-588

Design concept: https://github.com/jahstreet/spark-on-kubernetes-helm/blob/develop/README.md

## How was this patch tested?

Was tested manually on AKS cluster (Azure Kubernetes Services), Kubernetes v1.11.8:
- Image: Spark 2.4.3 with Hadoop 3.2.0 (https://github.com/jahstreet/spark-on-kubernetes-docker)
- History Server: https://github.com/helm/charts/tree/master/stable/spark-history-server
- Jupyter Notebook with Sparkmagic: https://github.com/jahstreet/spark-on-kubernetes-helm/tree/master/charts/jupyter

What do you think on that?

PR #388, opened 2023-01-31
Provide a way to disable the fix for LIVY-697

idzikovsky:
The fix for the LIVY-697 (#246) works fine for the problem it was designed to solve.  
But it causes another problem:  
the host reported to Livy Server from the RSC Driver is never used.

I know that this is kind of unsupported for now, but I'm trying to run Livy in Kubernetes in the namespace with Istio injections enabled.

And after fix for LIVY-697 it stopped working (I use Livy 0.7 with patches from #249), howerver when I used patches from #167 applied on Livy 0.5 everything was fine.

**Root cause**
In the fix for #167 we are explicitly [setting the host](https://github.com/apache/incubator-livy/pull/249/files#diff-43114318c4b009c2404f7eb326a84c184fb1501a3237c49a771df851d0f6f328R175-R177) that will be reported by RSC Driver back to Livy Server. And because of the fix for LIVY-697 a value of this property is ignored.

And even in that case, everything works fine, until we try to enable Istio: as Istio wraps all TCP traffic through a proxy, the following piece of code would set the host property not to the host of the actual RSC Driver, but to the IP address of TCP proxy server which would be 127.0.0.6:
```java
        InetSocketAddress insocket = (InetSocketAddress) ctx.channel().remoteAddress();
        host = insocket.getAddress().getHostAddress();
```

So it would be good to add some option to disable the workaround that was made in the scope of LIVY-697.  
Especially taking into account that initially Livy was designed to connect to the driver using host/port reported by the driver, and not by using hacks that get the host from a remote connection channel.

In my fix, I'm providing an option to enable the fix for LIVY-697 and leave it enabled by default to keep the current behavior as is.

## How was this patch tested?

Manual


idzikovsky:
This is still relevant. Can I do something to help to proceed with this PR?

idzikovsky:
This is still relevant

idzikovsky:
This seems to be implemented in https://github.com/apache/livy/pull/544
[LIVY-1071](https://issues.apache.org/jira/browse/LIVY-1071).

PR #390, opened 2023-02-25
[LIVY-971] Support to get session variables when using JDBC to connect to Livy thrift server.

jianzhenwu:
… to Livy thrift server.

Add `DescLivySessionOperation` to support fetching sessionId, applicationId, sessionState, sessionLogs from Livy thrift service.
Solve the bug that the thrift api gets stuck when creating an interactive session fails.

https://issues.apache.org/jira/browse/LIVY-971

## How was this patch tested?

integration tests

### DIFF
--- a/integration-test/src/thriftserver/test/scala/org/apache/livy/test/JdbcIT.scala
+++ b/integration-test/src/thriftserver/test/scala/org/apache/livy/test/JdbcIT.scala
@@ -19,6 +19,7 @@ package org.apache.livy.test
 
 import java.sql.Date
 
+import org.apache.livy.sessions.SessionState
 import org.apache.livy.test.framework.BaseThriftIntegrationTestSuite
 
 class JdbcIT extends BaseThriftIntegrationTestSuite {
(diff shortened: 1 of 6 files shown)


mgaido91:
I think this information is available on the Livy web application. I do not know any SQL engine that allows for a similar command, and we try to follow SQL standard. I think we should use the web app for monitoring the status, so if there is something missing there, I would recommend improving that rather than adding a new (unusual) SQL command.

jianzhenwu:
> I think this information is available on the Livy web application. I do not know any SQL engine that allows for a similar command, and we try to follow SQL standard. I think we should use the web app for monitoring the status, so if there is something missing there, I would recommend improving that rather than adding a new (unusual) SQL command.

jianzhenwu:
hi @mgaido91 thanks for your response. I think this feature has the following two usage scenarios. 
1. When other platforms use JDBC to integrate with Livy, they need to save the session ID and application ID.
2. When the connection creation fails, it may be that the spark session failed to start. At this time, they need to check the reason why the spark session failed to start.
What do you think?

jianzhenwu:
I agree with your points. Thank you for sharing your insights.

PR #393, opened 2023-03-14
[LIVY-972] Add GitHub Action for CI image

dacort:

- Adds Several GitHub Workflows
  - [x] CI image builder
  - [x] Unit tests for `git push`
  - [x] Integration tests for PRs

## How was this patch tested?

Manual validation in the course of the pull request and new GitHub Actions workflows.

### DIFF
--- a/.github/workflows/build-ci-image.yaml
+++ b/.github/workflows/build-ci-image.yaml
@@ -18,6 +18,9 @@ name: 'Build CI images'
 on: 
   push:
     branches: ["main", "feature/github-actions"]
+    # Commented out until PR is ready
+    # paths:
+    # - 'dev/docker/Dockerfile'
 jobs:
   docker-build:
     runs-on: ubuntu-latest
--- a/.github/workflows/build-ci-image.yaml
+++ b/.github/workflows/build-ci-image.yaml
@@ -38,10 +38,9 @@ jobs:
       - 
         name: Build and push image
         id: docker_build
-        uses: docker/build-push-action@v2
+        uses: docker/build-push-action@v4
         with:
           push: true
           context: ./dev/docker
-          file: ./Dockerfile
           tags: |
             ghcr.io/${{ github.repository_owner }}/livy-ci:latest
--- a/.github/workflows/build-ci-image.yaml
+++ b/.github/workflows/build-ci-image.yaml
@@ -17,10 +17,9 @@
 name: 'Build CI images'
 on: 
   push:
-    branches: ["main", "feature/github-actions"]
-    # Commented out until PR is ready
-    # paths:
-    # - 'dev/docker/Dockerfile'
+    branches: ["main"]
+    paths:
+    - 'dev/docker/Dockerfile'
 jobs:
   docker-build:
     runs-on: ubuntu-latest
--- a/.github/workflows/build-ci-image.yaml
+++ b/.github/workflows/build-ci-image.yaml
@@ -1,0 +1,47 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+name: 'Build CI images'
+on: 
+  push:
+    branches: ["main", "feature/github-actions"]
+jobs:
+  docker-build:
+    runs-on: ubuntu-latest
+    steps:
+      - 
+        name: Checkout
+        uses: actions/checkout@v3
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      -
+        name: Login to the GitHub Container Registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - 
+        name: Build and push image
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          context: ./dev/docker
+          file: ./Dockerfile
+          tags: |
+            ghcr.io/${{ github.repository_owner }}/livy-ci:latest
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -35,7 +35,7 @@ jobs:
       uses: actions/cache@v3
       with:
         path: ~/.m2/repository
-        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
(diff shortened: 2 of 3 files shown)


.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,29 @@
...
+  build:
+    runs-on: ubuntu-20.04
+    container: ghcr.io/${{ github.repository_owner }}/livy-ci:latest
+    steps:
+    - 
+      name: Checkout
+      uses: actions/checkout@v3
+    - 
+      name: Build with Maven
+      run: mvn -Pthriftserver -DskipITs -Dmaven.javadoc.skip=true -B -V -e verify
```
ksumit, line 29: I'm assuming we will add more jobs that cover other scenarios?

ksumit (commented):
Thanks for taking this up.

.github/workflows/build-ci-image.yaml
```diff
@@ -0,0 +1,46 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+name: 'Build CI images'
+on: 
+  push:
+    branches: ["main", "feature/github-actions"]
```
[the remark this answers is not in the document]
dacort: Yes, I'll add a `path` on this that limits the action to if the `./dev/docker/Dockerfile` changes and also remove the feature branch - had that here so I could test the action.

.github/workflows/unit-tests.yaml
dacort: Yes, planning on adding the other pre-existing jobs from `.travis.yml` file, which includes thriftserver unit tests and Spark 3.x tests for each.

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,54 @@
...
+      name: Checkout
+      uses: actions/checkout@v3
+    - 
+      name: Cache local Maven repository
+      uses: actions/cache@v3
+      with:
+        path: |
+          /root/.m2/repository
+          !/root/.m2/repository/org/apache/livy
+        key: ${{ runner.os }}-maven-${{ hashFiles('pom.xml', '*/pom.xml', 'thriftserver/*/pom.xml', 'core/*/pom.xml', 'repl/*/pom.xml', 'scala-api/*/pom.xml') }}
```
ksumit, line 42: does `hashFiles` support `**/pom.xml`?

ksumit (commented):
Would it make sense to merge `integration-tests.yaml` and `unit-tests.yaml` and use some sort of flag to use one vs the other. We could also avoid building the repository two times (though I don't think that's a big problem since the builds are happening in parallel and probably not adding to the runtime of the overall pipelines). Thoughts?

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,54 @@
...
+env:
+  MAVEN_OPTS: -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.http.retryHandler.class=standard -Dmaven.wagon.http.retryHandler.count=3
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    # TODO: Possibly point to the ./build-ci-image.yaml with the "uses" key
+    container: ghcr.io/${{ github.repository_owner }}/livy-ci:latest
+    strategy:
+      matrix:
+        spark_version: ["2.4", "3.0"]
```
ksumit, line 30: I think we can add the unit-test vs integration-test logic here.

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,54 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+name: Integration Tests
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+env:
+  MAVEN_OPTS: -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.http.retryHandler.class=standard -Dmaven.wagon.http.retryHandler.count=3
```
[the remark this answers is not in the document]
dacort: Unfortunately, yes. If the Docker build is busy for a while and doesn't download anything, the http pool times out apparently. There have been other projects that resolved this in a similar way like [pulsar](https://github.com/apache/pulsar/pull/8386).

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,52 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+name: Unit Tests
+on: [push]
+env:
+  MAVEN_OPTS: -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.http.retryHandler.class=standard -Dmaven.wagon.http.retryHandler.count=3
```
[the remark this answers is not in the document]
dacort: Yep.

.github/workflows/build-ci-image.yaml
```diff
@@ -0,0 +1,48 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+name: 'Build CI images'
+on: 
+  push:
+    branches: ["main"]
+    paths:
+    - 'dev/docker/Dockerfile'
```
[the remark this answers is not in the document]
dacort: 1. Yes, good idea, will add that.
2. Hm, good point. I had the `main` branch here so we wouldn't push the image unless it got merged into main, but would be good to be able to validate the image as well before doing so.

dacort:
> Would it make sense to merge `integration-tests.yaml` and `unit-tests.yaml`

I wasn't quite sure if there was an easy way to do that. Can definitely look more into it, but ideas welcome.

ksumit:
> > Would it make sense to merge `integration-tests.yaml` and `unit-tests.yaml`
> 
> I wasn't quite sure if there was an easy way to do that. Can definitely look more into it, but ideas welcome.

Here is an example that i found on https://docs.github.com/en/actions/learn-github-actions/contexts
```
name: CI
on: push
jobs:
  prod-check:
    if: ${{ github.ref == 'refs/heads/main' }}
    runs-on: ubuntu-latest
    steps:
      - run: echo "Deploying to production server on branch $GITHUB_REF"
```

.github/workflows/integration-tests.yaml
ksumit: @dacort I wouldn't want to hold this PR for minor things. If you feel comfortable with the current state, I would push it as it is (something better than nothing) and then iterate on the improvements later on. Let me know what you think.

gyogal:
Thanks a lot for working on this @dacort! I agree with @ksumit's points and if you feel like this is good to go as is, we could merge it to restore CI functionality for the project.
Post-merge would it be possible to run this for PRs received recently where the builds were not kicked off (such as #396)?

dacort:
Thanks @gyogal - Yea, @ksumit I think for now we should merge this. I imagine the PR workflow and unit-test workflow will diverge in the future so would rather get this merged and iterate.

@gyogal re: run the checks retroactively, I don't think it'll happen automatically but if folks push another commit it should.

lmccay (approved):
+1

gyogal (approved):

PR #394, opened 2023-03-14
[LIVY-593] Add flags to support hadoop, spark and scala version specific builds

ksumit:
Added support for specific flags to support hadoop/spark/scala version specific builds.

## How was this patch tested?

Ran `mvn install` with the new flags to confirm they work.


README.md
```diff
@@ -88,3 +88,13 @@ between different Spark versions.
...
+
+### Build Profiles
+
+| Flag        | Purpose                                                            |
+|-------------|--------------------------------------------------------------------|
+| -Phadoop2   | Choose Hadoop2 based build dependencies (default configuration)    |
+| -Pspark2    | Choose Spark 2.x based build dependencies (default configuration)  |
+| -Pscala-2.11 | Choose Scala 2.11 based build dependencies (default configuration) |        
+| -Pscala-2.12 | Choose scala 2.12 based build dependencies                         |
+| -Pspark3 | Choose Spark 3.x based build dependencies |
```
ayushtkn, line 100: This sounds tough to do and very hard coded, Do we have an option like -Dhadoop.version=3.3.5 and have a way to provide them at runtime, if we can do that

ksumit: It is hardcoded intentionally, getting different versions of hadoop, hive and spark to play with a version of livy and making it work across their dependency versions is a hard problem. you would notice that the functionality to pass and override [hadoop](https://github.com/apache/incubator-livy/blob/master/pom.xml#L82), [scala](https://github.com/apache/incubator-livy/blob/master/pom.xml#LL109-L111C57), [spark](https://github.com/apache/incubator-livy/blob/master/pom.xml#LL86-L86C63) and other versions is already available but we still need these profiles to resolve version incompatibilities in dependencies.

PR #398, opened 2023-04-03
LIVY-973 Always close RSCClient's EventLoopGroup

RajshekharMuchandi:

Added try-catch block to jobs' `setFailure()`and  log entries if there is any exception. `eventLoopGroup.shutdownGracefully();` statement will be executed after exception is caught.

https://issues.apache.org/jira/browse/LIVY-973

## How was this patch tested?


gyogal (approved):
LGTM

PR #404, opened 2023-05-03
[LIVY-974][SERVER] Remove verbose output on Livy UI error pages

askhatri:

On error, the Livy UI shows verbose output on error pages including the Jetty version number. This could be considered as a security vulnerability. We can make it configurable and avoid sending server version details. The Jetty version is there in every response header as well.

JIRA: https://issues.apache.org/jira/browse/LIVY-974

## How was this patch tested?

Verified manually by calling get sessions via REST API call in a local Yarn cluster.

### DIFF
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -48,6 +48,9 @@
 # livy.server.request-header.size = 131072
 # livy.server.response-header.size = 131072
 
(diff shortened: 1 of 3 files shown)


RajshekharMuchandi (approved):

askhatri:
> Contributor

Thank you @gyogal !

PR #405, opened 2023-05-14
 [LIVY-593] Add flags to support hadoop, spark and scala version specific builds

ksumit:
Added support for specific flags to support hadoop/spark/scala version specific builds.

## How was this patch tested?

Ran `mvn install` with the new flags to confirm they work.


ksumit:
@lmccay @dacort @askhatri please check when you've a chance.

lmccay (approved):
+1

Issue #415, opened 2023-08-24
Escape backtick from spark-submit arguments


s0nskar:
Currently, livy does not escape backticks from user-provided spark-submit arguments. So if a customer is passing any arguments that contain backticks, it will be considered as command substitution during spark-submit causing that argument to become blank or invalid.

Example:
```
--query 'select * from test_db.`test_table`' 
```
will become
```
--query 'select * from test_db.' 
```

s0nskar:
cc: @pralabhkumar

s0nskar:
I'm working on this.

PR #417, opened 2023-08-26
Support for escaping backtick from spark-submit arguments

s0nskar:

Fixes - #415 

Adding support for escaping backticks from spark-submit arguments.
- This feature will be behind a flag `livy.server.escapeBackTicks` (default: false)
- This also handles the case where the user is already escaping backtick and livy won't perform double escaping.

## How was this patch tested?

- Added unit test for escape backticks function.

### DIFF
--- a/server/src/main/scala/org/apache/livy/utils/SparkProcessBuilder.scala
+++ b/server/src/main/scala/org/apache/livy/utils/SparkProcessBuilder.scala
@@ -154,6 +154,12 @@ class SparkProcessBuilder(livyConf: LivyConf) extends Logging {
     this
   }
 
+  def escapeBackTicks(arg: String): String = {
+    // Unescape the args first then escape it again to avoid double escaping
+    val tmp = arg.replace("\\`", "`")
+    tmp.replace("`", "\\`")
+  }
+
   def start(file: Option[String], args: Traversable[String]): LineBufferedProcess = {
     var arguments = ArrayBuffer(_executable)
 
@@ -194,6 +200,11 @@ class SparkProcessBuilder(livyConf: LivyConf) extends Logging {
 
     arguments += file.getOrElse("spark-internal")
     arguments ++= args
+
+    if (livyConf.getBoolean(LivyConf.SPARK_SUBMIT_ESCAPE_BACKTICKS)) {
+      debug("Escaping backticks if present in spark-submit arguments")
+      arguments = arguments.map(escapeBackTicks(_))
+    }
 
     val argsString = arguments
       .map("'" + _.replace("'", "\\'") + "'")
--- a/server/src/test/scala/org/apache/livy/utils/SparkProcessBuilderSuite.scala
+++ b/server/src/test/scala/org/apache/livy/utils/SparkProcessBuilderSuite.scala
@@ -1,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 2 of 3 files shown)


s0nskar:
cc: @pralabhkumar

pralabhkumar:
Hi @s0nskar  . I am looking into this .

server/src/test/scala/org/apache/livy/utils/SparkProcessBuilderSuite.scala
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.livy.{LivyBaseUnitTestSuite, LivyConf}
+import org.scalatest.{BeforeAndAfter, FunSpec}
+
+class SparkProcessBuildSuite
+  extends FunSpec
+    with BeforeAndAfter
+    with org.scalatest.Matchers
+    with LivyBaseUnitTestSuite {
+
+  it("should only escape unescaped backticks") {
```
pralabhkumar, line 29: Please add the Jira number ([LIVY-988](https://issues.apache.org/jira/browse/LIVY-988))  in the comments .

server/src/main/scala/org/apache/livy/utils/SparkProcessBuilder.scala
```diff
@@ -154,6 +154,12 @@ class SparkProcessBuilder(livyConf: LivyConf) extends Logging {
     this
   }
 
+  def escapeBackTicks(arg: String): String = {
+    // Unescape the args first then escape it again to avoid double escaping
+    val tmp = arg.replace("\\`", "`")
```
pralabhkumar, line 159: Nit : Better naming for the tmp variable

server/src/main/scala/org/apache/livy/utils/SparkProcessBuilder.scala
```diff
@@ -154,6 +154,12 @@ class SparkProcessBuilder(livyConf: LivyConf) extends Logging {
     this
   }
 
+  def escapeBackTicks(arg: String): String = {
```
pralabhkumar, line 157: Please add scala doc for the method

pralabhkumar, line 157: If this method is used only in SparkProcessBuilder , can u please make its access specification as private . Also please add check for arg , (what if its null/ None)

pralabhkumar (changes requested):

pralabhkumar:
@s0nskar  Build is failing because of checkstyle issue , please look into fixing this.

server/src/test/scala/org/apache/livy/utils/SparkProcessBuilderSuite.scala
s0nskar: Done

server/src/main/scala/org/apache/livy/utils/SparkProcessBuilder.scala
s0nskar: done

server/src/main/scala/org/apache/livy/utils/SparkProcessBuilder.scala
s0nskar: done

RonZhang724:
@pralabhkumar My company uses AWS EMR for data processing, and EMR uses livy for job submissions. We ran into issue with backticks recently because the content in backticks gets removed. 

Would you mind following up on this PR and include it in the future releases?

server/src/main/scala/org/apache/livy/utils/SparkProcessBuilder.scala
s0nskar: exposed `escapeBackTicks` for testing purposes.

PR #426, opened 2023-10-06
[LIVY-989][SERVER] Livy core support for interactive session idleTimeout

askhatri:

Currently, a Livy interactive session has a field called ttl, which kills the session if it has been idle for a given amount of time. However, here is the expected behaviour:

**ttl:** kills the session if it has been active for a certain duration, irrespective of idleness.
**idleTimeout:** kills the session if it has been idle for the given duration. (The current TTL behaves in this manner.)

JIRA: https://issues.apache.org/jira/browse/LIVY-989

## How was this patch tested?

Tested manually by creating interactive session with idle time and TTL.


gyogal:
LGTM, thanks for your contribution!

PR #433, opened 2023-11-14
Add IPv6 preferred JVM support

Alexvsalexvsalex:
Now loopback address is 127.0.0.1 always.
But it's not work for JVM that was run with `-Djava.net.preferIPv6Addresses=true`. It's better to pass host (that will be ::1 for IPv6) to py4j gateway creation.

## How was this patch tested?
Livy with this fix is using on IPv6 machines now.


Alexvsalexvsalex:
@askhatri
Can you look at the PR, please?

PR #435, opened 2023-11-20
[LIVY-992][SERVER] In livy logs, user is not able to differentiate between interactive session and batch session

RajshekharMuchandi:

Changes done in log info to differentiate session type (interactive or batch) so that it will be easier to identify while deleting sessions
https://issues.apache.org/jira/browse/LIVY-992 

## How was this patch tested?

Created both interactive and batch sessions. Then deleted sessions manually with DELETE API. Then checked logs to verify session type


askhatri (approved):

gyogal (approved):
LGTM, thanks for your contribution!

PR #440, opened 2024-01-24
[LIVY-977][SERVER][CONF] Livy can not be started if HDFS is still in safe mode

RajshekharMuchandi:
Added safe mode check to implement safe mode

HDFS safe mode is checked when livy session is created. If safe mode is ON, then IllegalStateException is thrown after max retry attempts (configurable) with safe mode interval (configurable) checks are done. If safe mode is OFF, then livy will be able to create session.
 
https://issues.apache.org/jira/browse/LIVY-977

## How was this patch tested?

Added unit test cases to validate code changes. Also, done manual testing in CDP cluster by creating livy sessions with HDFS safe mode check ON/OFF.

### DIFF
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -195,3 +195,8 @@
 # Enable to allow custom classpath by proxy user in cluster mode
 # The below configuration parameter is disabled by default.
 # livy.server.session.allow-custom-classpath = true
+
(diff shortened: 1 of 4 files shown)


askhatri (approved):

gyogal:
LGTM, thank you for your contribution @RajshekharMuchandi !

PR #445, opened 2024-04-22
[LIVY-998][THRIF] Support connecting to an existing sessions using session name

askhatri:
Support connecting to an existing session using the session name via Thrift Server.

JIRA: https://issues.apache.org/jira/browse/LIVY-998

The Thrift Server can set the session name, but it cannot use an existing session name and connect to it. Added the fix to use an existing session name and connect to it via Thrift Server.

## How was this patch tested?

Created a session using session name:

```
curl --no-progress-meter -X POST -H "Content-Type: application/json" --data '{"kind": "spark", "name": "sesssionName"}' http://HOST:PORT/sessions
```

Connected to the existing session via Thrift Server using:

```
$ beeline -u "jdbc:hive2://HOST:PORT/?livy.session.name=sesssionName"
```

### DIFF
--- a/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
+++ b/thriftserver/server/src/main/scala/org/apache/livy/thriftserver/LivyThriftSessionManager.scala
@@ -155,12 +155,13 @@ class LivyThriftSessionManager(val server: LivyThriftServer, val livyConf: LivyC
   }
 
   /**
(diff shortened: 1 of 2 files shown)


gyogal (approved):
LGTM, thanks, Asif! A minor nitpick I have is there is a typo in the commit message, but that can updated fixed when the change is merged.

RajshekharMuchandi (approved):

PR #449, opened 2024-06-21
[LIVY-1003][RSC] Interactive session - Setting large value of rsc.ser…

wangdengshan:
…ver.connect.timeout blocks other tasks

The main adjustment here is the thread pool used when creating and closing sessions asynchronously. Scala's default thread pool size is limited, which will cause the waiting thread to be blocked.

https://issues.apache.org/jira/browse/LIVY-1003

## How was this patch tested?

How to reproduce:
1. Set `livy.rsc.server.connect.timeout` to something high like 24h.
2. Create enough interactive livy sessions in YARN so that they are queued in ACCEPTED state. The number of sessions that are stuck in ACCEPTED state should be equal to global execution context [thread pool size](https://docs.scala-lang.org/overviews/core/futures.html#the-global-execution-context) (Runtime.availableProcessors)
3. Try to delete a session using DELETE /sessions/
{sessionId}
and it should not be hang until one of the sessions is no longer stuck in ACCEPTED state.

### DIFF
--- a/server/src/main/scala/org/apache/livy/LivyConf.scala
+++ b/server/src/main/scala/org/apache/livy/LivyConf.scala
@@ -138,6 +138,10 @@ object LivyConf {
     Entry("livy.server.thrift.async.exec.wait.queue.size", 100)
   val THRIFT_ASYNC_EXEC_KEEPALIVE_TIME =
     Entry("livy.server.thrift.async.exec.keepalive.time", "10s")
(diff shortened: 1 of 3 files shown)


askhatri (approved):
The changes look good to me.

nileshrathi345 (approved):
Have tested changes and looks good to me

gyogal:
Thank you for your contribution @wangdengshan (and thanks also for reviewing @askhatri @nileshrathi345)!

PR #456, opened 2024-09-12
[LIVY-785] Enable adding security related HTTP headers

nileshrathi345:

This change introduces a new configuration option `livy.server.security-headers.enabled`.
When this property is set to true, the following security headers are added to HTTP
responses by default:
* X-XSS-Protection
* X-Frame-Options
* X-Content-Type-Options
* Strict-Transport-Security
* Content-Security-Policy

Also, adds content type information to all responses as required when using content type option nosniff

## How was this patch tested?

Tested manually


gyogal (approved):
LGTM, thank you for your contribution!

PR #464, opened 2025-01-17
[LIVY-1009] Add support for global TTL for sessions

nileshrathi345:

Added two new Livy configuration parameters as below:

```
Enabled to check whether TTL Livy sessions should be stopped.
livy.server.session.ttl-check = false

Time in milliseconds on how long Livy will wait before TTL is an inactive session.
Note that the inactive session could be busy running jobs.
livy.server.session.ttl = 8h
```

## How was this patch tested?

Read https://docs.google.com/document/d/1F2hU1RfShmNCKEWWkz_LpjEOev2BwbtZZBEueHpvgHM/edit for details.

### DIFF
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -67,6 +67,13 @@
 # How long a finished session state should be kept in LivyServer for query.
 # livy.server.session.state-retain.sec = 600s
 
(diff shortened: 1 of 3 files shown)


gyogal (approved):
LGTM

PR #467, opened 2025-02-10
[LIVY-1011] Upgrade CI image, fix Spark 3 tests

gyogal:

The CI environment is broken due to library version conflicts between the image and the checkout action.

The livy-ci Docker image needs to be upgraded to a more recent version of Ubuntu, ensuring that the unit tests and integration tests all pass.

Due to a bug in Livy's GitHub workflows, tests had only been executed with Spark 2. Now that this is fixed, we also need to make updates to some of the Python-based tests to work with Python 3 because Spark 3.2 does not support Python 2.

Two R-based integration tests are ignored in the Spark 2 integration test runs because SparkR 2 does not work with R 4, and the last version of Ubuntu with R 3.6 is Ubuntu 20. If we want to re-enable these tests, we may need to build R 3.6 for Ubuntu 24.

## How was this patch tested?

CI and unit test runs in a private fork of the repo.

### DIFF
--- a/.github/workflows/build-ci-image.yaml
+++ b/.github/workflows/build-ci-image.yaml
@@ -17,7 +17,7 @@
 name: 'Build CI images'
 on: 
   push:
-    branches: ["main"]
+    branches: ["master"]
     paths:
     - 'dev/docker/livy-dev-base/Dockerfile'
 jobs:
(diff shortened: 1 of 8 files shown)


RajshekharMuchandi (approved):
LGTM

PR #471, opened 2025-03-24
[LIVY-1012] Use SslContextFactory.Server() instead of SslContextFactory

surendralilhore:

Use SslContextFactory.Server() instead of SslContextFactory() when constructing the factory.

## How was this patch tested?

With Jetty 9.4.50, we should call SslContextFactory.Server(), instead of SslContextFactory(), to create SslContextFactory. Otherwise we get the following error when using a KeyStore with multiple certificates in it.

```
Exception in thread "main" java.lang.IllegalStateException: KeyStores with multiple certificates are not supported on the base class org.eclipse.jetty.util.ssl.SslContextFactory. (Use org.eclipse.jetty.util.ssl.
SslContextFactory$Server or org.eclipse.jetty.util.ssl.SslContextFactory$Client instead)
at org.eclipse.jetty.util.ssl.SslContextFactory.newSniX509ExtendedKeyManager(SslContextFactory.java:1289)
```

### DIFF
--- a/server/src/main/scala/org/apache/livy/server/WebServer.scala
+++ b/server/src/main/scala/org/apache/livy/server/WebServer.scala
@@ -49,7 +49,7 @@ class WebServer(livyConf: LivyConf, var host: String, var port: Int) extends Log
       https.setSendServerVersion(livyConf.getBoolean(LivyConf.SEND_SERVER_VERSION))
       https.addCustomizer(new SecureRequestCustomizer())
 
-      val sslContextFactory = new SslContextFactory()
+      val sslContextFactory = new SslContextFactory.Server()
       sslContextFactory.setKeyStorePath(keystore)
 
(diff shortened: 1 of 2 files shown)


gyogal (commented):
Thanks for your PR! I think your change looks good, however there is a similar "new SslContextFactory" in `ThriftHttpCLIService.scala` as well, could you please fix that as well if the same issue applies there?

surendralilhore:
Thank you, @gyogal , for the review. I have addressed your comment.

Similar fix is present in [hive](https://github.com/apache/hive/blob/28abf17424e97112d75947f1d7a2f043b3961fa4/service/src/java/org/apache/hive/service/cli/thrift/ThriftHttpCLIService.java#L163), and `ThriftHttpCLIService.scala` copy of hive code.

PR #480, opened 2025-07-07
Upgraded jetty version to 10.0.1

namanagraw:
What changes were proposed in this pull request?

1) Upgraded jetty to 10.0.1
2) Upgraded javax.servlet-api.version to 4.0.1

## How was this patch tested?
Tested in local
image


gyogal:
Thank you for opening this PR @namanagraw ! I think it would be good to save this PR from being auto-closed because upgrading to version 11+ will be necessary. Would it be possible to resolve the conflict in `pom.xml` and maybe use an even newer version of Jetty (the latest from 11.x or maybe even 12.x)?

I have opened [LIVY-1045](https://issues.apache.org/jira/browse/LIVY-1045) for the Jetty library upgrade.

gyogal:
@namanagraw I have resolved the conflict in `pom.xml` and started the check jobs, however it seems like there is a scalastyle error in one of the changed lines (an extra space after the comma), which blocks the PR from getting merged:

```
[INFO] --- scalastyle-maven-plugin:0.8.0:check (default) @ livy-server ---
error file=/__w/livy/livy/server/src/main/scala/org/apache/livy/server/WebServer.scala message=Whitespace at end of line line=97 column=49
```

Could you please fix this minor issue?

gyogal:
The Java 8 test runs are failing because the minimum required Java version for Jetty 10 is 11. With the upcoming Spark 4 upgrade, the Java 8 test runs will probably be removed, so this PR can be revisited.

gyogal:
For future reference, here is the compatibility matrix that shows the minimum required JVM and servlet versions: https://jetty.org/download.html

PR #482, opened 2025-09-26
[LIVY-1018] Extract field comments from Spark schema metadata and set in Datatypeutils

ArnavBalyan:
- During schema fetch, ensure we can extract and persist the comments.
- Column comments extracted from Spark metadata for `complexToDataType` and `schemaFromSparkJson`
- Closes: #483 

## How was this patch tested?
 - CI

### DIFF
--- a/thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
+++ b/thriftserver/server/src/test/scala/org/apache/livy/thriftserver/ThriftServerSuites.scala
@@ -310,6 +310,59 @@ trait CommonThriftTests {
 
   }
 
+  def getColumnsWithCommentsTest(connection: Connection): Unit = {
+    val metadata = connection.getMetaData
+    val statement = connection.createStatement()
+    try {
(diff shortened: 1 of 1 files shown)


gyogal:
Hi @ArnavBalyan , thank you for your contribution, I think it looks good! Could you please open a LIVY ticket for it at https://issues.apache.org/ and add it to the commit message? If possible, could you please include unit tests as well?

gyogal:
@ArnavBalyan Thanks for adding the JIRA ticket to the PR description, however is LIVY-483 related to this issue? (It seems to be about sparkR memory configuration.) If a new issue is needed and you don't have a JIRA user yet, could you please request one?

gyogal:
Hi @ArnavBalyan , I was able to confirm that with your change the comments are populated for struct member fields on the server side, however I was not able to find a way to read this information with a JDBC client. Could you please let me know which client or API would allow me to check this on the client side?

gyogal:
Sounds good! Thank you @ArnavBalyan once again for your contribution!

PR #485, opened 2025-09-28
[LIVY-1024] Upgrade Livy to Python3

ArnavBalyan:
 - Python2 was deprecated a while back, most systems/environments don't support it anymore
 - Bump Livy to Python3, remove Python2 support since Spark 2 has also now been deprecated
 - Closes LIVY-1024

## How was this patch tested?
 - CI / Locally tested the python3 files manully


ArnavBalyan:
> Thank you, I think your change looks great! There is one more reference to Python 2 that could be changed: https://github.com/apache/incubator-livy/blob/master/repl/src/main/scala/org/apache/livy/repl/PythonInterpreter.scala#L52 Spark uses "python3" by default: https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/deploy/PythonRunner.scala#L51

Updated thanks

gyogal (approved):
LGTM, thanks, @ArnavBalyan !

PR #495, opened 2025-11-24
[LIVY-786] Interrupt interpreter threads when cancelling statements

ArnavBalyan:
 - Cancel API doesn't cancel anything running on the Spark driver
 - track the interpreter thread for each running statement and interrupt it during cancel
 -  Clean up the threads and reset interrupt after statement completion

## How was this patch tested?
 - Unit test added

Closes LIVY-786

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -189,8 +189,6 @@
     <maven.compiler.source>${java.version}</maven.compiler.source>
     <maven.compiler.target>${java.version}</maven.compiler.target>
 
-    <!-- Exclude test and provided licenses when generating the THIRD-PARTY license file -->
-    <license.excludedScopes>test,provided</license.excludedScopes>
   </properties>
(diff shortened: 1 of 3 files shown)


ArnavBalyan:
Hi @lmccay just wanted to gently bump if you could pls take a look thanks!

lmccay (changes requested):
Added a couple questions for clarification. Please update the description with the answers to make sure it is clear what the intent is here when done.
I'm also not very familiar with this particular code and would be more comfortable with another reviewer approving as well.

nileshrathi345:
@ArnavBalyan Tested on cluster: cancellation works for spark (Scala) sessions — e.g. Thread.sleep() is interrupted and the session returns to idle.
It does not work for pyspark (Python) sessions.

PR #520, opened 2026-04-25
[LIVY-1043] Track ZK connection transitions in Livy

ArnavBalyan:
 - Currently ZKManager does not observe connection transitions. When underlying connection changes, Livy is not aware of it, only ZK side logs show up.
 - Additionally register to ConnectionStateListener, which adds/listens to Livy owned logs for the connection changes.
 - Same events can be used for recovery in the future.

## How was this patch tested?
 - Unit tests


ArnavBalyan:
cc @gyogal @lmccay thanks!

gyogal (approved):
LGTM, thanks, @ArnavBalyan !

PR #522, opened 2026-05-01
[LIVY-1046] Avoid decoding user info twice in HTTP client

ArnavBalyan:
 - The HTTP client decodes the user info in url twice. 
 - Java `URI` returns percent decoded output which is passed to .decode() again. Any password sent with special characters is wrongly decoded.
 - Remove the second decode call and ensure `:` inside the password is not treated as a delimiter
 - Closes LIVY-1046

## How was this patch tested?
 - UT

### DIFF
--- a/client-http/src/main/java/org/apache/livy/client/http/LivyConnection.java
+++ b/client-http/src/main/java/org/apache/livy/client/http/LivyConnection.java
@@ -20,8 +20,6 @@ package org.apache.livy.client.http;
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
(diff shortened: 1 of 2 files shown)


ArnavBalyan:
cc @lmccay @gyogal thanks!

ArnavBalyan:
cc @gyogal just a gentle reminder when you have time thanks!

gyogal:
LGTM, thank you for fixing this bug @ArnavBalyan ! Could you please create a LIVY JIRA for it?

ArnavBalyan:
Ah added the JIRA thanks :)

PR #529, opened 2026-06-23
[LIVY-1059] Add livy.spark.yarn.queue configuration to set default YARN queue

nileshrathi345:

This PR introduces support for configuring a default YARN queue at the Livy server level via a new configuration property: `livy.spark.yarn.queue`.

Previously, if a client did not explicitly specify a YARN queue parameter (`queue`) in their session creation payload, the session would fall back to the global Hadoop cluster's default YARN queue. This change allows cluster administrators to isolate Livy-generated workloads into a specific default queue without requiring end-users to pass it manually in every API request.

When a client provides `queue` in the session creation request, that value takes priority over the server default.

## How was this patch tested?

- Added unit tests in `BatchSessionSpec` that verify `--queue` is passed to spark-submit when using the LivyConf default or a user-provided queue
- Added unit tests in `InteractiveSessionSpec` that verify the resolved queue is stored on the session
- Manual API validation against a YARN cluster

### DIFF
--- a/server/src/main/scala/org/apache/livy/LivyConf.scala
+++ b/server/src/main/scala/org/apache/livy/LivyConf.scala
@@ -44,6 +44,7 @@ object LivyConf {
   val SPARK_HOME = Entry("livy.server.spark-home", null)
   val LIVY_SPARK_MASTER = Entry("livy.spark.master", "local")
   val LIVY_SPARK_DEPLOY_MODE = Entry("livy.spark.deploy-mode", null)
+  val SPARK_YARN_QUEUE = Entry("livy.spark.yarn.queue", null)
 
   // Two configurations to specify Spark and related Scala version. These are internal
   // configurations will be set by LivyServer and used in session creation. It is not required to
@@ -484,6 +485,9 @@ class LivyConf(loadDefaults: Boolean) extends ClientConf[LivyConf](null) {
 
   /** Return the spark master Livy sessions should use. */
   def sparkMaster(): String = get(LIVY_SPARK_MASTER)
+
+  /** Return the default YARN queue Livy sessions should use. */
+  def sparkYarnQueue(): Option[String] = Option(get(SPARK_YARN_QUEUE)).filterNot(_.isEmpty)
 
   /** Return the path to the spark-submit executable. */
   def sparkSubmit(): String = {
--- a/server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
+++ b/server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
@@ -83,7 +83,7 @@ object BatchSession extends Logging {
       request.executorMemory.foreach(builder.executorMemory)
       request.executorCores.foreach(builder.executorCores)
       request.numExecutors.foreach(builder.numExecutors)
-      request.queue.foreach(builder.queue)
+      request.queue.orElse(livyConf.sparkYarnQueue()).foreach(builder.queue)
       request.name.foreach(builder.name)
 
       sessionStore.save(BatchSession.RECOVERY_SESSION_TYPE, s.recoveryMetadata)
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -47,6 +47,10 @@
 # What spark deploy mode Livy sessions should use.
 # livy.spark.deploy-mode =
 
+# What default YARN queue Livy sessions should use if not specified in the client request.
+# By default, it is null and falls back to the global Hadoop cluster default queue.
+# livy.spark.yarn.queue = default
+
 # Configure Livy server http request and response header size.
 # livy.server.request-header.size = 131072
 # livy.server.response-header.size = 131072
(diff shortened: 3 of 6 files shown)


nileshrathi345:
@ArnavBalyan, @gyogal and @roczei could you please take a look when possible thanks!

server/src/main/scala/org/apache/livy/LivyConf.scala
```diff
@@ -41,6 +41,7 @@ object LivyConf {
 
   val TEST_MODE = ClientConf.TEST_MODE
 
+  val SPARK_YARN_QUEUE = Entry("livy.spark.yarn.queue", null)
```
ArnavBalyan: Can we add documentation for this config to `livy.conf.template` too

ArnavBalyan (commented):
Thanks for the changes @nileshrathi345, could you pls look into the CI failures, overall LGTM

server/src/main/scala/org/apache/livy/LivyConf.scala
nileshrathi345: thanks..!! 
Documented config to `livy.conf.template`

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
```diff
@@ -40,6 +40,7 @@ case class BatchRecoveryMetadata(
     appTag: String,
     owner: String,
     proxyUser: Option[String],
+    queue: Option[String],
```
gyogal: Why does the recovery state store save the queue the batch has been started with? Do we want to be able to show it after a Livy server restart? I think this may be independent new functionality not covered by LIVY-1059 (which is only about allowing the user to set a default YARN queue)

server/src/main/scala/org/apache/livy/server/batch/BatchSessionServlet.scala
```diff
@@ -33,6 +33,7 @@ case class BatchSessionView(
   state: String,
   appId: Option[String],
   appInfo: AppInfo,
+  queue: Option[String],
```
gyogal: I think this change is only needed if we want to be able to return which queue the batch job has been submitted into. If so, the Batch object may need to be updated at https://livy.apache.org/docs/latest/rest-api.html
But this is something LIVY-1059 does not currently mention, which is only about defaulting the value for the queue parameter.

gyogal (commented):
Thanks for submitting a PR @nileshrathi345 ! I left two minor comments, could you please check and let me know your thoughts?

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
nileshrathi345: Thanks for the clarification. Agreed that persisting queue in BatchRecoveryMetadata is out of scope for LIVY-1059.

The reason queue appears in recover() is purely a constructor consistency concern — since BatchSession now accepts queue as a constructor parameter (added in create() to store the resolved value for the lifetime of the session), recover() must also pass a value to satisfy that constructor. In this case, can now pass **None** explicitly, meaning no queue is stored or exposed for recovered sessions, which aligns with point that recovery does not need to know or show the original queue.

server/src/main/scala/org/apache/livy/server/batch/BatchSessionServlet.scala
nileshrathi345: As per previous comment now No queue field has been added to BatchRecoveryMetadata, nothing is persisted to the state store, and the REST API Batch object is unchanged. The only observable behaviour change from LIVY-1059 is that batch.queue returns the resolved value during a live session created with the default conf fallback.

server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
```diff
@@ -137,6 +138,7 @@ object BatchSession extends Logging {
       m.owner,
       m.proxyUser,
       sessionStore,
+      None,
```
gyogal: I think I now understand why the queue was added to the state store in the previous version, it was so that this value could be populated correctly. Is the queue in BatchSessions used by anything else other than the unit test?

nileshrathi345: Currently, queue is primarily used during session creation and tracking. However, passing None during recover means that if the Livy server restarts, any recovered batch session will permanently lose its original queue context.
To maintain that have added to the state in previous version

gyogal: It might be possible to remove this and change the unit test in BatchSessionSpec to verify that the queue is set via the mockApp. I am not entirely sure though, but if that is the case this PR could be further simplified.

nileshrathi345: I have removed the parameter from BatchSession.recover and updated the unit tests in BatchSessionSpec to verify that the queue configuration is correctly resolved via the mockApp lifecycle instead. This keeps the recovery metadata schema unchanged and simplifies the PR as suggested.

gyogal (approved):
LGTM, thank you for updating your PR @nileshrathi345 !

PR #531, opened 2026-07-08
[LIVY-1059] Fix missing validation condition and add unit test

nileshrathi345:

This is a follow-up fix for [LIVY-1059](https://github.com/apache/livy/pull/529).

When a client sends `"queue": ""` in the session create JSON, Livy was treating it as a valid queue value instead of falling back to the server default `livy.spark.yarn.queue`.

This change treats an empty queue string the same as a missing queue, so the default YARN queue from Livy config is used when the client passes "".

Updated in both batch and interactive session creation:
`request.queue.filterNot(_.isEmpty).orElse(livyConf.sparkYarnQueue())`

## How was this patch tested?

- Added unit test in `BatchSessionSpec` for `queue = Some("")` and verified the default queue is passed to spark-submit
- Added unit test in `InteractiveSessionSpec` for `queue = Some("")` and verified the session uses the default queue from `LivyConf`
- Existing tests for `queue = None` and user-provided queue values still pass

### DIFF
--- a/server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
+++ b/server/src/main/scala/org/apache/livy/server/batch/BatchSession.scala
@@ -83,7 +83,7 @@ object BatchSession extends Logging {
       request.executorMemory.foreach(builder.executorMemory)
       request.executorCores.foreach(builder.executorCores)
       request.numExecutors.foreach(builder.numExecutors)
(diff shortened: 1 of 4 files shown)


gyogal (approved):
LGTM, thanks for fixing this, @nileshrathi345 !

PR #532, opened 2026-07-23
LIVY-1060: Livy Spark sessions fail to connect to Hive Metastore over SSL with keystore password error

nileshrathi345:

Resolves Hive Metastore SSL credential handling for Spark sessions/jobs submitted by Livy.

- Added `enrichHiveMetastoreSslConf` in `SparkApp.prepareSparkConf` to read `hive.metastore.keystore.password` and `hive.metastore.truststore.password` from Livy’s local credential provider (`livy.hadoop.security.credential.provider.path`) and inject them as `spark.hadoop.*` properties for the Spark app.
- Ensures YARN/K8s driver and executors get explicit HMS SSL passwords without needing access to the Livy host’s local JCEKS file.
- Preserves the existing credential provider path in Spark conf (fix for interactive session creation regression from blanking the path).

JIRA: https://issues.apache.org/jira/browse/LIVY-1060

## How was this patch tested?

- Added `SparkAppSpec` unit tests for no-op, successful HMS SSL resolution, unrelated credentials, and unreadable provider path cases.
- Verified credential resolution from a local JCEKS test store and no-op paths when provider is missing, unreadable, or has no HMS aliases.

No

### DIFF
--- a/server/src/main/scala/org/apache/livy/utils/SparkApp.scala
+++ b/server/src/main/scala/org/apache/livy/utils/SparkApp.scala
@@ -17,9 +17,13 @@
 
 package org.apache.livy.utils
 
+import java.io.IOException
+
 import scala.collection.JavaConverters._
 
-import org.apache.livy.LivyConf
+import org.apache.hadoop.conf.Configuration
+
+import org.apache.livy.{LivyConf, Logging}
 
 object AppInfo {
   val DRIVER_LOG_URL_NAME = "driverLogUrl"
@@ -54,7 +58,7 @@ trait SparkAppListener {
 /**
  * Provide factory methods for SparkApp.
  */
-object SparkApp {
+object SparkApp extends Logging {
   private val SPARK_YARN_TAG_KEY = "spark.yarn.tags"
 
   object State extends Enumeration {
@@ -73,7 +77,7 @@ object SparkApp {
       uniqueAppTag: String,
       livyConf: LivyConf,
       sparkConf: Map[String, String]): Map[String, String] = {
-    if (livyConf.isRunningOnYarn()) {
+    val baseConf = if (livyConf.isRunningOnYarn()) {
       val userYarnTags = sparkConf.get(SPARK_YARN_TAG_KEY).map("," + _).getOrElse("")
       val mergedYarnTags = uniqueAppTag + userYarnTags
       sparkConf ++ Map(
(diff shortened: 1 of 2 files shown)


nileshrathi345:
@gyogal  and @roczei could you please take a look when possible
thanks..!!!!

server/src/main/scala/org/apache/livy/utils/SparkApp.scala
```diff
@@ -89,6 +93,53 @@ object SparkApp {
...
     }
+    enrichHiveMetastoreSslConf(baseConf, livyConf)
+  }
+
+  /**
+   * Resolves Hive Metastore SSL credentials from the local JCEKS credential provider
+   * (reachable only on the Livy server host) and injects them as explicit plaintext
+   * spark.hadoop.* properties so the driver/executors on YARN don't need access to
+   * the local file. The provider path itself is blanked afterward so it isn't
+   * propagated to nodes where it can't be resolved.
```
roczei: It looks like the comment here is outdated. The code in SparkApp.scala doesn't actually blank the provider path. deleteRecursively seems to only exist in SparkAppSpec.scala.

server/src/main/scala/org/apache/livy/utils/SparkApp.scala
```diff
@@ -89,6 +93,53 @@ object SparkApp {
...
+  private def enrichHiveMetastoreSslConf(
+      conf: Map[String, String],
+      livyConf: LivyConf): Map[String, String] = {
+    val credentialProviderPath = livyConf.get(LivyConf.HADOOP_CREDENTIAL_PROVIDER_PATH)
+    if (credentialProviderPath == null || credentialProviderPath.isEmpty) {
+      return conf
+    }
+    val hadoopConf = new Configuration()
+    hadoopConf.set("hadoop.security.credential.provider.path", credentialProviderPath)
+    val hmsSslKeys = Seq("hive.metastore.keystore.password", "hive.metastore.truststore.password")
```
roczei, line 119: The entire point of using a JCEKS credential provider is to avoid exposing passwords in plaintext configs. By injecting them directly as Spark properties, they will very likely show up on the /environment page in both the Spark UI and History Server UI (unless redacted by spark.redaction.regex, which should be enabled by default). Could you double-check how these appear on those UIs and also ensure that the passwords aren't being printed anywhere in the logs?

roczei (commented):
Hi @nileshrathi345 ,

Thanks for the PR! I’ve left a few comments for you to review.

Also, could you please add the AI tool usage section to the PR description, as outlined in our template?

Question to include:
"Was this patch authored or co-authored using generative AI tooling?"

Reference template: https://github.com/apache/livy/blob/master/.github/PULL_REQUEST_TEMPLATE#L13C4-L13C70

Thanks!

server/src/main/scala/org/apache/livy/utils/SparkApp.scala
nileshrathi345: **Spark UI:** spark.hadoop.hive.metastore.truststore.password shows as *********(redacted) — default spark.redaction.regex is sufficient.
**Livy logs/API:** If the password is injected/passed as session Spark conf,
```
"conf": {
  "spark.app.name": "password-in-conf-test",
  "spark.hadoop.hive.metastore.truststore.password": "test-trust-secret-12345"
}
```
it appears in GET /sessions/{id} and the InteractiveSession$ create log in plaintext. YARN startup logs do not print it. **(Same behaviour without this changes also)**
**spark-defaults.conf:** Password reaches Spark (UI redacted) but is not returned by Livy API and not logged at session creation.

server/src/main/scala/org/apache/livy/utils/SparkApp.scala
nileshrathi345: Updated the comment

server/src/main/scala/org/apache/livy/utils/SparkApp.scala
roczei: @nileshrathi345 Okay, thanks for verifying!

roczei (approved):
LGTM

gyogal (approved):
Thanks for the contribution @nileshrathi345 and for the review @roczei !

PR #533, opened 2026-07-24
LIVY-866: Optimize Yarn GetApplications Query to prevent additional load on Yarn and Livy

nileshrathi345:

**Problem**
Livy was calling YarnClient.getApplications(SPARK), which asks the ResourceManager for all Spark applications and then filters client-side by application tag. On clusters with many Spark apps, this produces large RM responses, adds unnecessary load on YARN, and can contribute to RM memory pressure (especially during session startup polling and leaked-app GC).

**Solution**
Use a tag-filtered GetApplicationsRequest so the RM returns only matching applications:
- **getAppIdFromTag()** — when resolving a session’s YARN app ID by tag, query with the lowercase session tag instead of scanning all Spark apps.
- **leakedAppsGCThread** — when cleaning up leaked apps, batch all pending leaked tags into one filtered request per GC cycle instead of fetching the full Spark app list.

Jira: https://issues.apache.org/jira/browse/LIVY-866

## How was this patch tested?

Updated and ran SparkYarnAppSpec, including: 
- "should get App Id" — verifies tag-based app ID lookup uses the filtered getApplications request.
- "should delete leak app when timeout" — verifies the leaked-app GC path uses the filtered request.

Yes, this was co-authored using Cursor to help understanding of  solution for **leakedAppsGCThread**

### DIFF
--- a/server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
+++ b/server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
@@ -16,6 +16,8 @@
  */
 package org.apache.livy.utils
 
+import java.util
+
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
 import scala.collection.mutable.ArrayBuffer
@@ -25,6 +27,7 @@ import scala.language.postfixOps
 import scala.util.Try
 import scala.util.control.NonFatal
 
+import org.apache.hadoop.yarn.api.protocolrecords.GetApplicationsRequest
 import org.apache.hadoop.yarn.api.records.{ApplicationId, ApplicationReport, FinalApplicationStatus, YarnApplicationState}
 import org.apache.hadoop.yarn.client.api.YarnClient
 import org.apache.hadoop.yarn.conf.YarnConfiguration
@@ -68,13 +71,23 @@ object SparkYarnApp extends Logging {
 
   private var sessionLeakageCheckInterval: Long = _
 
+  /**
+   * Build a GetApplicationsRequest filtered by Spark application type and tags.
+   * Tags are normalized to lowercase to match YARN's tag storage behavior.
+   */
+  private def createGetApplicationsRequest(appTags: util.Set[String]): GetApplicationsRequest = {
+    val normalizedTags = new util.HashSet[String]()
+    appTags.asScala.foreach(tag => normalizedTags.add(tag.toLowerCase))
+    val request = GetApplicationsRequest.newInstance(appType)
+    request.setApplicationTags(normalizedTags)
+    request
+  }
+
   private val leakedAppsGCThread = new Thread() {
     override def run(): Unit = {
-      val client = {
-        mockYarnClient match {
-          case Some(client) => client
-          case None => yarnClient
-        }
+      val client = mockYarnClient match {
+        case Some(client) => client
+        case None => yarnClient
       }
 
       while (true) {
@@ -82,13 +95,16 @@ object SparkYarnApp extends Logging {
           // kill the app if found it and remove it if exceeding a threshold
           val iter = leakedAppTags.entrySet().iterator()
           val now = System.currentTimeMillis()
-          val apps = client.getApplications(appType).asScala
+          val tagSet = new util.HashSet[String](leakedAppTags.keySet())
+          val request = createGetApplicationsRequest(tagSet)
+          val apps = client.getApplications(request).asScala
 
           while(iter.hasNext) {
             var isRemoved = false
             val entry = iter.next()
+            val tagLowerCase = entry.getKey.toLowerCase()
 
-            apps.find(_.getApplicationTags.contains(entry.getKey))
+            apps.find(_.getApplicationTags.contains(tagLowerCase))
               .foreach({ e =>
                 info(s"Kill leaked app ${e.getApplicationId}")
                 client.killApplication(e.getApplicationId)
@@ -196,10 +212,10 @@ class SparkYarnApp private[utils] (
     }
 
     val appTagLowerCase = appTag.toLowerCase()
-
-    // FIXME Should not loop thru all YARN applications but YarnClient doesn't offer an API.
-    // Consider calling rmClient in YarnClient directly.
-    yarnClient.getApplications(appType).asScala.find(_.getApplicationTags.contains(appTagLowerCase))
+    val appTags: util.Set[String] = util.Collections.singleton(appTagLowerCase)
+    val request = createGetApplicationsRequest(appTags)
+    val applicationReports = yarnClient.getApplications(request)
+    applicationReports.asScala.find(_.getApplicationTags.contains(appTagLowerCase))
     match {
       case Some(app) => app.getApplicationId
       case None =>
--- a/server/src/test/scala/org/apache/livy/utils/SparkYarnAppSpec.scala
+++ b/server/src/test/scala/org/apache/livy/utils/SparkYarnAppSpec.scala
@@ -16,7 +16,6 @@
  */
 package org.apache.livy.utils
 
-import java.util.ArrayList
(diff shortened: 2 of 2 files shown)


nileshrathi345:
@gyogal, @ArnavBalyan  and @roczei  could you please take a look when possible and review
thanks..!!!!

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -68,6 +71,15 @@ object SparkYarnApp extends Logging {
 
   private var sessionLeakageCheckInterval: Long = _
 
+  /**
+   * Build a GetApplicationsRequest filtered by Spark application type and tags.
+   */
+  private def createGetApplicationsRequest(appTags: util.Set[String]): GetApplicationsRequest = {
```
roczei, line 78: Pulling createGetApplicationsRequest() into a private helper is a nice idea to avoid duplication.

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -82,7 +94,13 @@ object SparkYarnApp extends Logging {
           // kill the app if found it and remove it if exceeding a threshold
           val iter = leakedAppTags.entrySet().iterator()
           val now = System.currentTimeMillis()
-          val apps = client.getApplications(appType).asScala
+          val tagSet = new util.HashSet[String]()
```
roczei: The second iterator (entries) only exists to copy keys into a HashSet. This can be simplified to:

```suggestion
          val tagSet = new util.HashSet[String]()
          val tagSet = new util.HashSet[String](leakedAppTags.keySet())
```

This is one line instead of four does the same thing and eliminates the redundant iterator.

server/src/test/scala/org/apache/livy/utils/SparkYarnAppSpec.scala
```diff
@@ -49,6 +50,13 @@ class SparkYarnAppSpec extends FunSpec with LivyBaseUnitTestSuite {
     Thread.`yield`()
   }
 
+  private def mockGetApplicationsByTags(
+      client: YarnClient,
+      reports: List[ApplicationReport]): Unit = {
+    when(client.getApplications(any(classOf[GetApplicationsRequest])))
```
roczei, line 57: Using `any(classOf[GetApplicationsRequest])` is a bit broad, it won't catch bugs if wrong tags or app types are passed later. How about using an ArgumentCaptor here to verify the exact fields? For example:

```
diff --git a/server/src/test/scala/org/apache/livy/utils/SparkYarnAppSpec.scala b/server/src/test/scala/org/apache/livy/utils/SparkYarnAppSpec.scala
index cfaca49..55d69c4 100644
--- a/server/src/test/scala/org/apache/livy/utils/SparkYarnAppSpec.scala
+++ b/server/src/test/scala/org/apache/livy/utils/SparkYarnAppSpec.scala
@@ -30,6 +30,7 @@ import org.apache.hadoop.yarn.api.records.YarnApplicationState._
 import org.apache.hadoop.yarn.client.api.YarnClient
 import org.apache.hadoop.yarn.exceptions.ApplicationAttemptNotFoundException
 import org.apache.hadoop.yarn.util.ConverterUtils
+import org.mockito.ArgumentCaptor
 import org.mockito.Matchers.any
 import org.mockito.Mockito._
 import org.mockito.invocation.InvocationOnMock
@@ -411,6 +412,14 @@ class SparkYarnAppSpec extends FunSpec with LivyBaseUnitTestSuite {
 
           verify(mockYarnClient, atLeast(1)).getApplicationReport(appId)
           verify(mockListener).appIdKnown(appId.toString)
+
+          val requestCaptor = ArgumentCaptor.forClass(classOf[GetApplicationsRequest])
+          verify(mockYarnClient, atLeast(1)).getApplications(requestCaptor.capture())
+          val capturedRequest = requestCaptor.getValue
+          assert(capturedRequest.getApplicationTags.contains(appTag.toLowerCase),
+            s"Request must contain the lowercase appTag '${appTag.toLowerCase}'")
+          assert(capturedRequest.getApplicationTypes.contains("SPARK"),
+            "Request must filter by application type 'SPARK'")
         }
       }
     }
@@ -685,11 +694,20 @@ class SparkYarnAppSpec extends FunSpec with LivyBaseUnitTestSuite {
         SparkYarnApp.init(livyConf, Some(client))
 
         SparkYarnApp.leakedAppTags.clear()
-        SparkYarnApp.leakedAppTags.put("leakApp", System.currentTimeMillis())
+        val leakAppTag = "leakApp"
+        SparkYarnApp.leakedAppTags.put(leakAppTag, System.currentTimeMillis())
 
         Eventually.eventually(Eventually.timeout(TEST_TIMEOUT), Eventually.interval(100 millis)) {
           assert(SparkYarnApp.leakedAppTags.size() == 0)
         }
+
+        val requestCaptor = ArgumentCaptor.forClass(classOf[GetApplicationsRequest])
+        verify(client, atLeast(1)).getApplications(requestCaptor.capture())
+        val capturedRequest = requestCaptor.getValue
+        assert(capturedRequest.getApplicationTags.contains(leakAppTag.toLowerCase),
+          "Leaked app GC request must include the leaked app tag")
+        assert(capturedRequest.getApplicationTypes.contains("SPARK"),
+          "Leaked app GC request must filter by application type 'SPARK'")
       }
     }
 
```

nileshrathi345: Stronger tests with ArgumentCaptor — added verifyFilteredGetApplicationsRequest() to assert:
- Request includes expected tags (lowercase)
- Request filters by SPARK application type

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -35,13 +38,22 @@ import org.apache.livy.{LivyConf, Logging, Utils}
...
+
   def init(livyConf: LivyConf, client: Option[YarnClient] = None): Unit = {
     mockYarnClient = client
     sessionLeakageCheckInterval = livyConf.getTimeAsMs(LivyConf.YARN_APP_LEAKAGE_CHECK_INTERVAL)
     sessionLeakageCheckTimeout = livyConf.getTimeAsMs(LivyConf.YARN_APP_LEAKAGE_CHECK_TIMEOUT)
-    leakedAppsGCThread.setDaemon(true)
-    leakedAppsGCThread.setName("LeakedAppsGCThread")
-    leakedAppsGCThread.start()
+    if (!leakedAppsGCThreadStarted) {
+      synchronized {
```
roczei: @nileshrathi345 

Why do we need this synchronized block? Could we avoid it and restore the previous behavior?

nileshrathi345: The synchronized block was added to just to make sure SparkYarnApp.init() is safe to call more than once for the unit tests.
I have removed it and restore the previous behavior

roczei (approved):
LGTM

server/src/main/scala/org/apache/livy/utils/SparkYarnApp.scala
```diff
@@ -68,27 +71,40 @@ object SparkYarnApp extends Logging {
...
+    request.setApplicationTags(normalizedTags)
+    request
+  }
+
   private val leakedAppsGCThread = new Thread() {
     override def run(): Unit = {
-      val client = {
-        mockYarnClient match {
+      while (true) {
+        val client = mockYarnClient match {
```
gyogal: Could this be evaluated once, before the loop?

nileshrathi345: Done

PR #535, opened 2026-07-29
LIVY-1062: Handle HTTP 307 redirects for POST/DELETE in LivyConnection

nileshrathi345:

Handle HTTP **307 Temporary Redirect** in `LivyConnection.sendRequest()` so the Java HTTP client (`client-http`) retries POST, DELETE, and multipart requests to the `Location` URL instead of failing with `IOException: Temporary Redirect`.

Apache HttpClient follows redirects automatically for GET, but not for non-idempotent methods. `LivyConnection.sendRequest()` previously treated any non-200 response as fatal, so POST/DELETE/multipart calls failed whenever the server returned 307.

Also validates that `Location` is present/non-empty and consumes the redirect response entity before retry.

JIRA: [LIVY-1062](https://issues.apache.org/jira/browse/LIVY-1062)

## How was this patch tested?

Unit tests in `LivyConnectionSpec`:

- **Happy path:** POST (JSON), DELETE, and multipart POST via a mock standby server returning 307 to an active server — verifies method, headers (`X-Requested-By`), body, and response type are preserved.
- **Failure paths:** missing/empty `Location`, redirect loop, non-redirect 404.

Yes, this was co-authored using Cursor to write the new test cases.

Generated-by: Cursor Composer, Cursor 3.13.21

### DIFF
--- a/client-http/src/main/java/org/apache/livy/client/http/LivyConnection.java
+++ b/client-http/src/main/java/org/apache/livy/client/http/LivyConnection.java
@@ -24,6 +24,7 @@ import java.security.Principal;
 import java.util.concurrent.TimeUnit;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.http.Header;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpHeaders;
 import org.apache.http.HttpStatus;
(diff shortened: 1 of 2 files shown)


nileshrathi345:
Closing PR as this is more likely related to HA part which is not supported yet.

PR #536, opened 2026-08-04
[LIVY-1063] Fix resource leak in Kubernetes service account token refresh

nileshrathi345:

Reuse the existing `KubernetesClient` during service account token refresh instead of
recreating `DefaultKubernetesClient` every 5 minutes without closing the old client.

JIRA: https://issues.apache.org/jira/browse/LIVY-1063

## How was this patch tested?

- Existing unit tests: `mvn -pl server -am -DskipITs test`
- Manual review: refresh loop only calls `config.setOauthToken()`; no new client creation
- Manual testing: deployed on Kubernetes with in-cluster SA auth;
  verified "Refreshed Kubernetes service account token." logs every ~5 minutes and
  native/HTTP client thread count remains stable over several refresh cycles

Yes, used Cursor AI to write down the test case.

### DIFF
--- a/server/src/main/scala/org/apache/livy/utils/SparkKubernetesApp.scala
+++ b/server/src/main/scala/org/apache/livy/utils/SparkKubernetesApp.scala
@@ -76,27 +76,45 @@ object SparkKubernetesApp extends Logging {
     }
   }
 
+  /**
+   * Reload the OAuth token into the existing client configuration in place. Reuses the
+   * KubernetesClient created at startup; does not create a new one.
+   */
+  private[utils] def refreshServiceAccountToken(
+      client: KubernetesClient,
+      autoConfigure: String => Config = Config.autoConfigure(_)): Unit = {
+    val config = client.getConfiguration
+    val contextName = Option(config.getCurrentContext).map(_.getName).getOrElse("")
+    val newestConfig = autoConfigure(contextName)
+    config.setOauthToken(newestConfig.getOauthToken)
+  }
+
   val RefreshServiceAccountTokenThread = new Thread() {
+    private val tokenRefreshIntervalMs = 300000L
+
     override def run(): Unit = {
-      while (true) {
-        var currentContext = new Context()
-        var currentContextName = new String
-        val config = kubernetesClient.getConfiguration
-        if (config.getCurrentContext != null) {
-          currentContext = config.getCurrentContext.getContext
-          currentContextName = config.getCurrentContext.getName
+      while (!Thread.currentThread().isInterrupted) {
+        try {
+          refreshServiceAccountToken(kubernetesClient)
+          info("Refreshed Kubernetes service account token.")
+        } catch {
+          case e: InterruptedException =>
+            Thread.currentThread().interrupt()
+            warn("RefreshServiceAccountTokenThread was interrupted, exiting.", e)
+            return
+          case NonFatal(e) =>
+            warn("Failed to refresh Kubernetes service account token, will retry.", e)
         }
 
-        var newAccessToken = new String
-        val newestConfig = Config.autoConfigure(currentContextName)
-        newAccessToken = newestConfig.getOauthToken
-        info("Refreshed Kubernetes service account token.")
-
-        config.setOauthToken(newAccessToken)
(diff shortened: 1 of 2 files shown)


nileshrathi345:
@gyogal and @roczei could please take a look when possible and review.
thanks..!!!

gyogal (approved):
LGTM, thanks @nileshrathi345 !

server/src/main/scala/org/apache/livy/utils/SparkKubernetesApp.scala
```diff
@@ -78,25 +78,24 @@ object SparkKubernetesApp extends Logging {
...
+
+          // Token will expire 1 hour default, community recommend to update every 5 minutes
+          Thread.sleep(300000)
+        } catch {
+          case e: InterruptedException =>
+            Thread.currentThread().interrupt()
+            warn("RefreshServiceAccountTokenThread was interrupted, exiting.", e)
+            return
+          case NonFatal(e) =>
+            warn(s"Failed to refresh Kubernetes service account token, will retry.", e)
```
gyogal, line 98: I was wondering in case there is a NonFatal exception, shouldn't there be a delay before retrying? If an exception is thrown during token refresh, it may be spinning in the loop too fast... is this something that can happen?

gyogal (changes requested):

server/src/main/scala/org/apache/livy/utils/SparkKubernetesApp.scala
nileshrathi345: Valid point — without a delay, a persistent failure would retry in a tight loop. I've added a 5-minute sleep in the NonFatal path, same as the normal refresh interval, before retrying.

server/src/main/scala/org/apache/livy/utils/SparkKubernetesApp.scala
```diff
@@ -77,26 +77,27 @@ object SparkKubernetesApp extends Logging {
...
+          info("Refreshed Kubernetes service account token.")
+          Thread.sleep(tokenRefreshIntervalMs)
+        } catch {
+          case e: InterruptedException =>
+            Thread.currentThread().interrupt()
+            warn("RefreshServiceAccountTokenThread was interrupted, exiting.", e)
+            return
+          case NonFatal(e) =>
+            warn(s"Failed to refresh Kubernetes service account token, will retry.", e)
+            Thread.sleep(tokenRefreshIntervalMs)
```
gyogal: @nileshrathi345 , would it be possible to move the `Thread.sleep()` out of the exception handler? What I have in mind is a scenario where `Thread.sleep()` is interrupted and the `InterruptedException` is not caught or handled somewhere else (not sure if this would cause any problems, but it looks a bit unpredictable where execution goes next).

nileshrathi345: Good point — now I got it, having Thread.sleep() inside the NonFatal handler meant an interrupt during backoff wouldn't be caught cleanly. 
Moved the sleep outside the exception handler so both success and failure paths wait the same way, and InterruptedException is handled in one place.

server/src/main/scala/org/apache/livy/utils/SparkKubernetesApp.scala
```diff
@@ -77,26 +77,36 @@ object SparkKubernetesApp extends Logging {
   }
 
   val RefreshServiceAccountTokenThread = new Thread() {
+    // Reuse the KubernetesClient created at startup; refresh its OAuth token in place
+    private val tokenRefreshIntervalMs = 300000L
+
     override def run(): Unit = {
-      while (true) {
-        var currentContext = new Context()
```
roczei, line 82: Hi @nileshrathi345 

It'd be good to validate this with a unit test. Could you maybe add a new test for it?

nileshrathi345: Thanks @roczei  for suggestion — extracted the token refresh logic into refreshServiceAccountToken and added unit tests to verify the OAuth token is updated on the existing client config

roczei (approved):
Thanks for the new unit test! LGTM

PR #545, opened 2026-08-20
[LIVY-1072] Reduce Kubernetes API load in SparkKubernetesApp polling

soumyadeeplogin:
`SparkKubernetesApp.getApplicationReport` previously issued a single LIST returning every pod tagged with the session's app tag. On large jobs this deserializes hundreds of executor pods into memory on every poll cycle, even though application state is derived only from the driver pod.

Two changes:
- Narrow the driver lookup to `spark-role=driver` at the API server so the LIST returns a single pod.
- Gate the separate executor LIST behind a new config flag `livy.server.kubernetes.executor-tracking.enabled` (default `true`, so existing deployments see no behavior change). Operators running large clusters can disable it to skip the executor LIST entirely. When disabled, per-executor Grafana/Loki log URLs and per-executor entries in session diagnostics are omitted; driver state tracking is unaffected.

Adds two tests: default-enabled invariant, and diagnostics graceful degradation when the executor list is empty.

## How was this patch tested?
Existing and newly added unit tests in `SparkKubernetesAppSpec`, plus manual validation against a live Kubernetes cluster running Spark jobs with hundreds of executors.

Yes. Generated-by: Claude Code (Sonnet 5, Anthropic), used to scan/plan the extraction of local UIP patches for upstreaming and to draft/iterate on the code and tests in this PR under human review. Please refer to the [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html) for details.

## Note on JIRA
Filed as [LIVY-1072](https://issues.apache.org/jira/browse/LIVY-1072).

### DIFF
--- a/conf/livy.conf.template
+++ b/conf/livy.conf.template
@@ -275,6 +275,10 @@
 # details, routes, etc...)
 # livy.server.kubernetes.poll-interval = 15s
 
+# Whether Livy fetches executor pods each poll cycle (used for executor log URLs and
+# per-executor diagnostics). Disable on large clusters; driver pod is polled regardless.
+# livy.server.kubernetes.executor-tracking.enabled = true
+
 # Weather to create Kubernetes Nginx Ingress for Spark UI. If set to true, configure the desired
 # options below
 # livy.server.kubernetes.ingress.create = false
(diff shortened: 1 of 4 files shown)


De-ASI-INTERFACE (approved):

roczei (approved):
LGTM

gyogal (approved):

PR #546, opened 2026-08-20
[LIVY-1073] Add SessionManager refresh and /recovery operator endpoint

soumyadeeplogin:
## What changes were proposed in this pull request?
Adds a `SessionManager` refresh capability and a `/recovery` operator endpoint on `LivyServer`, allowing an operator to trigger a manual re-sync of in-memory session state against the recovery/session-state store without restarting the Livy server process.

The `/recovery/*` endpoints are implemented as a standalone `RecoveryServlet`, restricted to `livy.superusers`. If `livy.server.recovery.mode` is `off` (no state store configured), the endpoints return `409` rather than a silently-empty result. The three endpoints (`/recovery/sessions/refresh`, `/recovery/batches/refresh`, `/recovery/refresh`) are documented in `docs/rest-api.md`.

## How was this patch tested?
Unit tests added in `SessionManagerSpec` covering the refresh path, and `RecoveryServletSpec` covering the servlet's authorization (403 vs 200) and recovery-mode-off (409) behavior.

## Was this patch authored or co-authored using generative AI tooling?
Yes. Generated-by: Claude Code (Sonnet 5, Anthropic), used to scan/plan the extraction of local patches for upstreaming and to draft/iterate on the code and tests in this PR under human review. Please refer to the [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html) for details.

## Note on JIRA
Filed as [LIVY-1073](https://issues.apache.org/jira/browse/LIVY-1073).

PR #552, opened 2026-09-09
[LIVY-1078] Add StateStore.tryExclusiveCreate() abstraction across state-store impls

soumyadeeplogin:

Adds `StateStore.tryExclusiveCreate(key, value): Boolean`, an atomic create-if-absent primitive alongside the existing `set()` (which always overwrites). Each backend implements it using its own native atomic guarantee rather than a check-then-act race:

- **ZooKeeper**: relies on `create()` failing with `NodeExistsException` if the znode already exists.
- **FileSystem**: uses `FileContext.create()` with `CREATE` (no `OVERWRITE`), which fails atomically with `FileAlreadyExistsException`.
- **Blackhole**: always returns `true` (recovery disabled, so there's nothing to conflict with).

`SessionStore.trySave()` exposes this at the session level for callers that need "create this session's recovery record only if it doesn't already exist" semantics.

This is plumbing only — the new method is unused by any existing caller in this PR, so there is no behavior change. It's a building block for a follow-up fix to a session-ID collision issue (a stale id counter on a recovered Livy instance can otherwise cause one session to silently overwrite another's state file).

## How was this patch tested?

Added unit tests to `ZooKeeperStateStoreSpec`, `FileSystemStateStoreSpec`, `BlackholeStateStoreSpec`, and `SessionStoreSpec`, covering both the create-succeeds path and the key-already-exists path (asserting no side effects / no overwrite) for every backend.

Ran `mvn -pl server -am -Pscala-2.12 -Pspark3 verify`: 55/55 tests pass (8 new), scalastyle/checkstyle/RAT all clean.

Also addressed review feedback: `FileSystemStateStore.tryExclusiveCreate()` now writes to a uniquely-named temp file and performs the exclusive claim via a non-overwriting atomic rename (crash-safety fix), and wraps that rename in a `try`/`finally` so the temp file is cleaned up on any rename failure, not just a lost claim race.

Yes. Generated-by: Claude Code (Sonnet 5, Anthropic), used to implement `tryExclusiveCreate()` across all `StateStore` backends and their corresponding unit tests, and to draft/iterate on the follow-up crash-safety and temp-file-cleanup fixes from review feedback, under human review throughout. Please refer to the [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html) for details.


gyogal (approved):
LGTM

server/src/main/scala/org/apache/livy/server/recovery/FileSystemStateStore.scala
```diff
@@ -108,6 +108,22 @@ class FileSystemStateStore(
...
+      usingResource(fileContext.create(absPath(key), createFlag, CreateOpts.createParent())) {
+        newFile =>
+          newFile.write(serializeToBytes(value))
+          newFile.close()
+      }
+      true
+    } catch {
+      case _: FileAlreadyExistsException => false
+    }
+  }
```
gyogal, line 145: Actually, just one observation: the .crc file is removed by set, but not by tryExclusiveCreate: https://github.com/apache/livy/commit/805dcf26eb0155a54014829b44aa9bccf3778136

This is not necessarily as issue, but on checksummed filesystems, if the file is changed outside of Hadoop and the .crc file is not updated, a ChecksumException could occur during read.

soumyadeeplogin:
Hi @gyogal — this one is approved and CI is green. Could you merge it when you have a chance? Thanks!

gyogal:
Thanks for your contribution @soumyadeeplogin , it looks good, however I have one additional question before merging: the original `set()` function in `FileSystemStateStore` used the write-then-rename logic to avoid file corruption if livy-server crashes. The new `tryExclusiveCreate()` function does not follow this. Maybe I don't understand the purpose of the PR, but is this intentional, and if so, how do we know the file will not be partially written?

soumyadeeplogin:
Good catch, thanks for looking closely — you're right, and it was not intentional. The direct create() at the final path meant a crash between create and close would leave a truncated file at the destination, and every future tryExclusiveCreate call for that key would see it as already claimed (FileAlreadyExistsException) with no way to recover.

Pushed a fix: tryExclusiveCreate() now writes to a uniquely-named temp file first, then performs the exclusive claim via a non-overwriting atomic rename (fileContext.rename(tmpPath, absPath(key), Rename.NONE)) — same write-then-rename shape set() already uses. The destination is only ever visible once it's complete, and the rename itself still fails atomically with FileAlreadyExistsException if another writer wins the race, so the exclusivity guarantee holds. The temp file (and its .crc sidecar) is cleaned up if the rename loses the race.

Updated the FileSystemStateStoreSpec tests to cover this and re-ran the full suite locally — 14/14 pass. Let me know if you'd like anything else adjusted.

server/src/main/scala/org/apache/livy/server/recovery/FileSystemStateStore.scala
soumyadeeplogin: Good catch here too — this was against the original naive tryExclusiveCreate() (direct create() at the destination path, no temp file). The crash-safety fix pushed just now (write to a temp file, then non-overwriting atomic rename into place) already resolves this as a side effect: it deletes the temp file's .crc sidecar the same way set() does, right after the rename attempt, regardless of whether the rename won or lost the race. So there's no leftover/stale .crc at the destination in either outcome. Let me know if you still see a gap after the latest push.

server/src/main/scala/org/apache/livy/server/recovery/FileSystemStateStore.scala
```diff
@@ -108,6 +109,41 @@ class FileSystemStateStore(
...
   }
 
+  override def tryExclusiveCreate(key: String, value: Object): Boolean = {
+    // Write to a uniquely-named temp file first, then atomically rename it into place
+    // without overwrite -- same write-then-rename shape as set(), so a crash mid-write
+    // never leaves a partially-written file at the destination. Omitting Rename.OVERWRITE
+    // still makes the claim exclusive: the rename fails atomically with
+    // FileAlreadyExistsException if another writer already claimed the key in the
+    // meantime, so no separate (and racy) exists-check is needed.
+    val tmpPath = absPath(s"$key.${UUID.randomUUID()}.tmp")
```
gyogal, line 119: Thanks for updating `tryExclusiveCreate()`, it looks much better now, however I can still imagine a scenario where the `$key.${UUID.randomUUID()}.tmp` files are not cleaned up due to a crash or an exception. In `save()`'s case, the second call of `save()` will clean up the leftover `.tmp` file, but that is not the case here since the file name is partially random. Maybe a `try {} finally` cleanup for the `.tmp` file would be better (although it would not protect against a crash, it would still help for example when the rename operation throws a non-`FileAlreadyExistsException`).

gyogal:
@soumyadeeplogin Thanks for updating your PR. Could you please update the description so that the "Was this patch authored or co-authored using generative AI tooling?" section has the "Generated-by:" header, similarly to the recent Livy or Spark commits?

soumyadeeplogin:
@gyogal Updated the description to include the \`Generated-by:\` header, matching the convention from #546 — thanks for flagging it.

Also, per your earlier comment on the temp file: pushed a \`try\`/\`finally\` around the rename in \`tryExclusiveCreate()\` so the temp file is cleaned up on any rename failure, not just \`FileAlreadyExistsException\`. Added a test for that path too, 15/15 passing.

Let me know if this looks good — appreciate an approve/merge when you have a chance.

server/src/main/scala/org/apache/livy/server/recovery/ZooKeeperManager.scala
```diff
@@ -157,6 +157,19 @@ class ZooKeeperManager(
     }
   }
 
+  // Atomically create the znode only if it doesn't already exist. Relies on ZooKeeper's
+  // create() failing with NodeExistsException rather than a separate exists-check, which
+  // would be racy against concurrent creators.
+  def tryCreate(key: String, value: Object): Boolean = {
+    val data = serializeToBytes(value)
+    try {
+      curatorClient.create().creatingParentsIfNeeded().forPath(key, data)
```
gyogal: I asked Claude to take a closer look and a potential correctness issue was pointed out here:

  **Fix before this gets a caller — `ZooKeeperManager.tryCreate`:**
  Under the retry policy (`RetryNTimes`), a `create()` whose ack is lost can be retried and
  throw a *spurious* `NodeExistsException` for the node this client just created — so
  `tryCreate` returns `false` ("someone else won") when it actually won. `set()` shares the
  underlying `create()` but doesn't treat `NodeExistsException` as a signal, so this is the
  first place it becomes a correctness bug.

  `.withProtection()` is **not** the fix here — it mangles the znode name (`_c_<uuid>-…`),
  breaking this store's `key == path` contract. Disambiguate by reading the node back instead:

  ```diff
       } catch {
  -      case _: NodeExistsException => false
  +      case _: NodeExistsException =>
  +        // Under the retry policy, a create() whose ack was lost can be retried and see the
  +        // node this client itself just created. Read back to disambiguate: if it holds the
  +        // bytes we wrote, we are the creator. (Sound while values are unique per creator —
  +        // true for session RecoveryMetadata, which embeds the id.)
  +        java.util.Arrays.equals(curatorClient.getData().forPath(key), data)
       }
```

This is not necessarily a merge blocker (no callers yet), but it is worth considering before tryCreate() gets used. If this is added, new unit tests could be useful for this scenario.

gyogal: Actually this fix may not be correct if two threads want to create the node with the same content and whether this fix is needed depends on how `tryCreate()` will be used. Your change may be OK in its current form if you feel like this is not an issue that can happen in a real life scenario, so please let me know.

PR #554, opened 2026-09-15
[LIVY-1072][FOLLOWUP] Fix Scala 2.13 compile error in SparkKubernetesApp

gyogal:

`KubernetesExtensions.getApplicationReport` failed to compile under Scala 2.13 (the default build): `client.pods...list.getItems.asScala` returns `scala.collection.mutable.Buffer`, but the enclosing `val executors: Seq[Pod]` requires `scala.collection.immutable.Seq` (the 2.13 default for the unqualified `Seq` alias). Scala 2.12 allowed the implicit widening; 2.13 does not.

Adds `.toSeq`, matching the same conversion already used a few lines above for the driver-pod lookup.

## How was this patch tested?

`mvn install -pl core/scala-2.13,server -am -DskipTests` failed with this exact type-mismatch error before the fix and succeeds after it.

Generated-by: Claude Code (Sonnet 5)


roczei:
@gyogal 

I tested it and it resolved the compilation error. Thanks for the fix!

roczei (approved):

soumyadeeplogin:
LGTM, thanks for catching this @gyogal!

gyogal:
Thank you for reviewing the PR @roczei and @soumyadeeplogin!

PR #557, opened 2026-09-17
[LIVY-1082] Add config to disable Kubernetes driver pod log polling

soumyadeeplogin:

`SparkKubernetesApp` currently fetches the driver pod's Kubernetes log (`pods/<driver>/log`) on every poll cycle to populate live log lines in `/sessions/:id/log` and `/batches/:id/log` responses. On large clusters, or when driver logs are already collected externally (e.g. via a centralized log pipeline / Spark History Server), this is wasted request volume against the Kubernetes API server, and can be a source of transient errors during pod lifecycle transitions.

This PR adds a new config, `livy.server.kubernetes.driver-log-polling.enabled` (default `true`, preserving existing behavior), that when set to `false` skips the `pods/<driver>/log` fetch while Livy continues to poll pod state and diagnostics normally.

## Why are the changes needed?

To let operators opt out of the driver log poll when it's unnecessary, reducing Kubernetes API server load and avoiding transient log-fetch errors during pod lifecycle transitions, without losing pod state/diagnostics monitoring.

## Does this PR introduce _any_ user-facing change?

Yes. A new optional config `livy.server.kubernetes.driver-log-polling.enabled` is added (default `true`). When explicitly set to `false`, `/sessions/:id/log` and `/batches/:id/log` will no longer include live Kubernetes driver log lines; pod state and diagnostics continue to be polled and reported as before.

## How was this patch tested?

Added/updated unit tests in `SparkKubernetesAppSpec`:
- `should enable driver log polling by default`
- `resolveDriverAppLog` - `should skip fetching the log when driver log polling is disabled`
- `resolveDriverAppLog` - `should fetch the log when driver log polling is enabled`

Ran `mvn -pl server -am test -Dtest=SparkKubernetesAppSpec -DwildcardSuites=org.apache.livy.utils.SparkKubernetesAppSpec`; all 14 tests in the suite pass.

Yes. Generated-by: Claude Code (Sonnet 5, Anthropic), used to implement the driver-log-polling
toggle and its unit tests, under human review. Please refer to the
[ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html)
for details.


soumyadeeplogin:
Filed LIVY-1082 for this change.

soumyadeeplogin:
Hi @gyogal / @roczei — could one of you approve the pending CI workflow run for this PR (fork-contributor approval gate)? Happy to address any review feedback as well. Thanks!

conf/livy.conf.template
```diff
@@ -294,6 +294,10 @@
 # per-executor diagnostics). Disable on large clusters; driver pod is polled regardless.
 # livy.server.kubernetes.executor-tracking.enabled = true
 
+# Whether Livy fetches the driver pod's Kubernetes logs each poll cycle. Disable if driver
+# logs are collected externally and freshness in /sessions/:id/log, /batches/:id/log is not needed.
+# livy.server.kubernetes.driver-log-polling.enabled = true
```
gyogal, line 301: Your change looks good, I only have a minor observation here on the documentation: I think "freshness is not needed" and "won't include fresh Kubernetes driver logs." later on implies that there will some kind of logs, just not fresh. However if I understand correctly, based on the changes, there will not be any driver logs recorded in this case. During testing, did "/logs" return anything? If not, I think it would be better to indicate that logs will not be available at all.

soumyadeeplogin: You're right, thanks for catching that. Confirmed: when driver-log-polling is disabled, resolveDriverAppLog always returns an empty log, so kubernetesAppLog is permanently empty and /sessions/:id/log and /batches/:id/log return no driver log lines at all (not stale ones). Updated the doc wording in both livy.conf.template and LivyConf.scala to say so explicitly.

gyogal (approved):
Thanks for the update @soumyadeeplogin, the change LGTM now!

